PR #128, opened 2015-02-08
Plex code cleanup

maxpiva:
* Refactored Plex code.
* Fixed almost all glitches on Plex clients. (May increse the startup time of JMMServer the first time)
* Added experimental UPnP to the FileServer. Experimental UPnP requires changes in the plex plugin, but, may fix (Remote Clients may be unable to play video, if the video do not requires transcoding. PlexPlugin Issue #9) if the router supports Hairpin NAT.
* Fixed subtitle/audiostreams languages name in plex (somehow visual studio dosn't like kana strings).
* Fixed Plex Plugin does not display info text about the episode correctly. PlexPlugin Issue #8
* May fix Plex for XBMC has thumbnails broken. PlexPlugin #2. (Require testing).
* Fixed unicode characters in hasher.dll pinvoke


miracle152004:
I think the new files are missing in the commit. At least all the files in the plex folder and the UPnP.cs in the FileServer folder is nowhere to be found.

maxpiva:
strange, only changes got commited... Fixed in the new pull...

PR #197, opened 2015-11-28
KodiMediaInfo for Kodi - independend for Plex changes

bigretromike:
Depend on the view this is a bit dirty way but this fixed few problems with Plex implementation and left Plex untouched. More and more to be done in future :+1:


bigretromike:
Some bonus fix

bigretromike:
Maybe PLEX didn't need count of seasons, but Kodi dose ^_^

AreFinstad:
if this function for seasons for Kodi could be used for the Plex plugin aswell i would be really happy! :D 
@maxpiva

PR #211, opened 2016-01-16
Adding Lang selection for ALL section

bigretromike:


bigretromike:
Vote per Episode support added for Kodi

bigretromike:
We already had Trakt Support inside JMM which work ok. 
I added Scrobbling ability from inside JMM and gave it to Kodi via Api.
So Nakamori or other plugin can add notification to trakt also.

bigretromike:
Clean up from code, optimalization, typo fixes, code responds in void function so Nakamori know when something happend with success or not.

PR #222, opened 2016-02-18
Added dialog to browse DLNA devices

scorpioncoder:
Started work on UPnP device support. Created new dialog to view servers
and new class with functions to read servers. **Does not yet add to imported folders or import files.**


jmediamanager:
Do you have a plan on how this would fit into the architecture of JMM?
i.e. Hashing of files, import folders etc?

scorpioncoder:
I've only had a brief glance at the code for hashing and importing but once i have a good grasp of how they work, then i intend to start writing the necessary bits to integrate it fully.

PR #227, opened 2016-03-05
Added more support for importing DLNA devices.

scorpioncoder:
Adds device to import list. Import runs without errors however does not hash or add to database.

To-Do's:
- Need to change import query to add value of ImportFolderType into database.
- Need to figure out how the hasher works.

If i had to guess, the hasher needs certain information from the file it cannot get if it tries to read it off the network, however, some information about the file is stored in the XML document generated by my Browser function. Would it be possible to use this information (<a href="https://dl.dropboxusercontent.com/u/65476870/sample_browse.xml">sample</a>) to hash the file?


scorpioncoder:
Hashing now works

maxpiva:
You seems to not to understand what hasher do, please refrain to edit those files without knowing.
Hasher will calculate the ed2k, and optionally SHA1, MD5 and CRC32 of the file. You simple bypassed the hasher function, that will in fact, get no hashes to the remote file. The correct way to handle this, is to use the C# function hashing (not the dll optimized C version), download the remote (upnp) file in chunks and calculate the hashes for the remote file.

JMM is in fact a AniDB Client, and AniDB uses ED2K Hashes to identify files.
Hashes are required by JMM without them it will not work as intended.

The approach you need is: Map DLNA files into Stream object.
Stream should support seeking (mapped to http range) and by fully supported. (not write of course).

Then use that stream object to feed the hasher, and or the remote clients streaming (plex, kodi).

Also there is a big problem, Mediaportal uses mediainfo.dll to get the media information from the file, you should recreate that information from the DLNA Servere, or by downloading the entire file and run mediainfo.dll on the cached file.

ElementalCrisis:
No updates in almost two months so I'm going to close this. Feel free to comment and re-open if there is development on it.

PR #229, opened 2016-03-08
Localzing JMM Server

ElementalCrisis:
Updated all hardcoded text strings to allow JMM to be completely localized.


ElementalCrisis:
Please don't merge until 3.5 beta is over.

PR #243, opened 2016-04-28
Polish translation and new Kodi option

bigretromike:


da3dsoul:
Woo hoo

bigretromike:
I review all strings. A lot of them been very tricky, but its better than the one from translator - but it was a good base to start!

jmediamanager:
What is the new Kodi option.
I also notice that you have commented out quite a lot of TvDB code etc, so it will take me a while to review it before I merge it

JMMServer/Commands/TvDB/CommandRequest_TvDBDownloadImages.cs
```diff
@@ -37,7 +37,7 @@ public string PrettyDescription
 		public CommandRequest_TvDBDownloadImages()
 		{
 		}
-
+        /*
```
jmediamanager: I'm guessing you have commented this out because it is has 0 references?
The reason it has 0 references is because the class isn't actually used anywhere at the moment, but if it was you need this constructor

da3dsoul:
The Kodi option is to report the number of episodes that you have locally of a show, rather than the total that is reported by AniDB

bigretromike:
Yes @da3dsoul is right. and also yes @jmediamanager you are right about commenting this - in feb I started to cleaning not used code because I planed to replace every not needed code with something that would work, aparantly it was not merge then and sticked with my other work. I will hope you will have time to review my changes and leave function you think will be needed (because some of them were already implemented in v2 of same class).

JMMServer/JMMServiceImplementationKodi.cs
```diff
@@ -639,6 +640,9 @@ private System.IO.Stream GetGroupsFromFilter(int userid, string GroupFilterId)
                                         }
                                         v.OriginalTitle = v.OriginalTitle.Substring(0, v.OriginalTitle.Length - 1);
                                         //proper naming end
+                                        List<AnimeSeries> sers = grp.GetAllSeries();
+                                        AnimeSeries ser = sers[0];
+                                        v.totalLocal = ser.GetAnimeEpisodesCountWithVideoLocal();
```
da3dsoul: 644-645 should be
```
foreach (AnimeSeries ser in sers)
{
    v.totalLocal += ser.GetAnimeEpisodesCountWithVideoLocal();
}
```
to get proper totals, should it not?

bigretromike: Are you sure? because your increase this, and mine took the value once. Also I did test my code with this and it look ok, if your sum function also work intention way ?

da3dsoul: I haven't gotten it to work at all. It shows everything negative (-15 etc)

da3dsoul: I fixed it, there was an error in getAnimeEpisodesCountWithVideoLocal. Unfortunately, the system is flawed to begin with, as it returns the entire series total, not just for each level.
```
-Group Filters (no info)
---Group (first underlying series info, not total)
-----Series (total for whole series, this will be the only correct value)
-------Credits (This is counted in the watched state, causing everything else to be wrong)
-------Episodes (This will show an incorrect value, as it takes total for the series including specials and credits)
-------Specials (This will be incorrect, as it takes the total including episodes and credits)
```

bigretromike:
that why my code worked with the code it had ;-) good you fixd the system bug ;-) but then, you commit again without validating it

JMMServer/JMMServiceImplementationKodi.cs
da3dsoul: Yeah, I don't know kover doesn't show the unwatched numbers at every level, so it's kind of hard to test. It wasn't a system error actually. I misread one of my test queries. It was a typo in Nakamori. Made a PR for it.

da3dsoul:
I'm pretty bad about that. I use commits as a point to restore from if things get too messed up, but it will automatically add them to PRs even if there are not tested. Then over the next hour or day(s), I find an issue and recommit over it.

da3dsoul:
Also, do you mind if (for now) we comment out all of the WIP character code? It sends a lot of data, slowing things down for lists.

bigretromike:
characters are not WIP. It just didnt show up on Kodi 15.1 because it wasnt supported in easy way, 16 didnt change anything. Maybe 17 API will change this. I know it is a "hog" but I dont want to later add it and see people complain. For now let it be; I plan to implement a JMM side implementation to turn it off/on but that would need to be also implented on both sites.

bigretromike:
When you refer to WIP i was thinking about options for trakt etc, but I would like to leave characters. There shouldn't be a big boost in term of speed because JMM will send that data anyway, it was scary the first time I implemented it when each episode had it own characters ;-) bleach was 30mb XML ;-) that was scary, now its not that bad

da3dsoul:
Yeah, no kidding. We need an Extended info context menu item then. As far as I'm aware, that should work with cast without scraping. The current gamepad settings don't make it easy to open extended info.

bigretromike:
Could you explain a bit about the need of Extended Info context menu ? We talking about right click menu? and You want to insert what in there?

bigretromike:
Character info extended because that was missing

bigretromike:
This is my finaly fix till 3.5 release or if there is any bug that was introduce with those changes

da3dsoul:
That is assuming 3.5 is ever released :laughing:

JMMServer/Entities/TvDB_ImageFanart.cs
```diff
@@ -90,5 +90,16 @@ public bool Populate(int seriesID, XmlNode node)
 				return false;
 			}
 		}
+
+        public void Valid()
```
da3dsoul: Is this method used? If not, no point in it, right?

bigretromike: yes, but @jmediamanager said few time to not delete not used functions...

da3dsoul: but didn't you add it in this commit? or is it just whitespace and github diff sucks...

PR #256, opened 2016-05-16
Cache and Db rework

### LINKED ISSUE #142: Crossovers might force anime got stored on the wrong directory.
Some ANIDB episodes (Crossovers, Toriko x One Piece, etc) are in both Series, because the way JMM   get the default directory. (If there is an episode already stored, then use the episode path).

Per example:  If Toriko gets populated before one piece, and you got the crossover, One Piece episodes will end in the Toriko directory... and viceversa. Related to #38

### LINKED ISSUE #251: Groupfilters, cachestats and Anime Updates performance is subpar
On big collections, perfomance is subpar. Migrating the database to NOSQL and redo the whole system is a nono right now for me (No time).

So is time for plan B.

* groupfilter precalculations stored in DB.
* Optimize DB indices if possible.
* More intelligent approach to Anime Updates. (Recalculate less information).
* Stats Cache converted to a Group View Cache stored in the db, in a common format (class) that satisfy all clients.

maxpiva:
This is a major change, backup your database before running.  I'll recommend testing from my repo before push this request. So far so good on my machine, but additional testing will be welcomed

* Major rework of the cache system, will fix #251, and also JMMServer will return data faster, for clients consumption, startup will be really fast right now, but Db needs generation, you should expect from 15 min to 1 hour startup (Depending of the size of your collection) the first time (db generation), after that startup will go down to a few seconds.

* Internal changes that may partially fix #142 (more work needed), we have the assumption that one ANIDB_Episode might have only one ANIDB_Series, but that is not the case. One AnIDB_Episode might be in more than one series. (Toriko x One Piece crossover) is episode 1 in Toriko and Episode 452 in One Piece. JMMServer Signatures changed. Changes for JMMClient are already made. Additional work must be made, removing the AnimeID relation in AniDB_Episode. And create an One To Many Relation Table.


maxpiva:
Assumption on #142 were wrong, closing pull request. Till fix

PR #313, opened 2016-07-14
Introduce 'Nancy' as core for APIv2

bigretromike:
Using 'Nancy' for best cross-compatible. Also introduce easier route making for API. Currently port is set to 1234, but eventually will be change. This commit don't break anything, and code that is used as base for current API was untouched. This is push-ed now to avoid future compatibility breaks that could occur if someone would mess with core-api objects.

This is a introduce for APIv2


bigretromike:
Fix'd against conflict

bigretromike:
pushing to separate branch

PR #391, opened 2016-09-08
Performance improvements for AniDB_Anime cache initialisation

pmcleish:
After recently updating JMM Server, I encountered the cache initialization process which took several hours to complete on my media PC. So, I thought I'd see if I could improve the performance a bit.
Below is a quick summary of what I've done:

- AniDB_AnimeRepository.InitCache now uses a stateless session. Using the standard ISession has a large amount of overhead due to change tracking, etc. (the profiler I used indicated that approximately 80% of the time was spent in NHibernate's Commit method).
- The above change required the introduction of an ISessionWrapper to provide the common functionality from both ISession and IStatelessSession for performing SELECT queries in various parts of the system.
- AniDB_Anime processing was switch over to work in batches instead of one by one to reduce the number of round-trips to the database.

Below are the results of the change (Tested on an i7 2600k, 16GB RAM, SSD; with 3661 AniDB_Anime records in DB):

Database | Before | After
--- | --- | ---
**SQL Server (2014)** | 1h 02m 35s | 25s
**SQLite** | 2h 11m 30s | 27s
**MySQL (5.17.5)** | 1h 16m 12s | 2m 45s

Note that this PR only improves the cache initialization of the AniDB_Anime stage. I haven't attempted to tackle some of the other slow stages yet due to the amount of work involved. If I do make an attempt, I'll do it in a separate PR (assuming this one turns out ok).

Also, one bit that hasn't been fully tested is handling the default images in bulk. The database I'm working doesn't have any records for images (I think I must have turned that feature off), so it would be greatly appreciated if someone could give it a quick test (if not, I'll have to find some time to setup some test cases).

### DIFF
--- a/JMMServer/AniDBHelper.cs
+++ b/JMMServer/AniDBHelper.cs
@@ -12,6 +12,7 @@ using JMMServer.Commands;
 using JMMServer.Commands.Azure;
 using JMMServer.Entities;
 using JMMServer.Repositories;
+using JMMServer.Repositories.NHibernate;
 using NHibernate;
 using NLog;
 
@@ -1096,20 +1097,21 @@ namespace JMMServer
 
             AniDB_AnimeRepository repAnime = new AniDB_AnimeRepository();
             AniDB_Anime anime = null;
+            ISessionWrapper sessionWrapper = session.Wrap();
 
             bool skip = true;
             if (forceRefresh)
                 skip = false;
             else
             {
-                anime = repAnime.GetByAnimeID(session, animeID);
+                anime = repAnime.GetByAnimeID(sessionWrapper, animeID);
                 if (anime == null) skip = false;
             }
 
             if (skip)
             {
                 if (anime == null)
-                    anime = repAnime.GetByAnimeID(session, animeID);
+                    anime = repAnime.GetByAnimeID(sessionWrapper, animeID);
 
                 return anime;
             }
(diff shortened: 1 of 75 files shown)


da3dsoul:
In the future, try not to reformat the code, as it makes git throw a hissy calculating diff. I never had poor performance with this, but speedup is always welcome if it doesn't break anything. Do me a favor. I noticed you made a .wrap() to all of the sessions. Put that in a helper function so that we can call it as session and it forwards session.wrap(). @maxpiva what do think of this?

da3dsoul:
Also, I am committing a fix to a previous issue that modifies many of those files, so check the code again after.

pmcleish:
Sorry about the code reformat. It wasn't intentional, I think there was a file or two that Visual Studio complained about line endings and I clicked "Yes" out of habit to fix them.

I should have added more detail into the issue. It normally isn't a major problem when you start JMM Server, except for when the AniDB_Anime.ContractVersion is less than what it should be (e.g. like after a major JMM upgrade, etc). A friend and I both recently updated, and the cache initialization process took many hours for both of us (plus our collections are probably larger than average which wouldn't help).

I'm not quite sure I understand about the .Wrap() helper you want. I added .Wrap() as an extension method to both ISession and IStatelessSession to make it easy to use. Are you requesting something that converts ISessionWrapper to ISession?

da3dsoul:
Nvm I see what you did. And I just now regenerated my AniDB anime contracts. Yes, it takes a long time. It'll be nice if this speeds up contract generation. Does this affect performance of retrieval at all?

pmcleish:
General retrieval of contracts? no.
Every other operation is still using an ISession (But for the things I needed to change, it is simply wrapped by ISessionWrapper). So their performance remains the same.
I was going to attempt to improve performance of other areas as well, but it is a lot of work, so I just wanted to dip my toe in, so to speak, and get some other peoples thoughts.

da3dsoul:
Ok I made my commit. Look over it and see if any changes need to be adapted to your code. I'll try and leave that section alone for a bit.

ElementalCrisis:
Thanks for the PR! @maxpiva is one of our lead programmers so he'll have to review, in the meantime and hopefully going forward come join us on our Discord server! It's always a pleasure to meet our users. :)

https://discordapp.com/channels/96234011612958720/96234011612958720

da3dsoul:
I lied. There is a contract issue with my commit

ElementalCrisis:
Hello @pmcleish, 

Looks like @da3dsoul commit's made the PR unable to be merged in, we talked it over in and would like to include this very nice addition to the last maintenance release for 3.6 which we're labeling 3.6.1.0 before we move on to the next version. So once the conflicts have been fixed, I'll be able to merge it in.

maxpiva:
Merged

PR #415, opened 2016-09-14
Replace ObjectExtensions class with DeepCloner Library

pmcleish:
While investigating a separate issue, I started up JMM Server and had to wait 45 mins for it to create 3500 Group Filter tags (i.e. populating GroupFilter table for the first time). During that time I attached a profiler and found that 95% of the time was spent in the ObjectExtensions.DeepCopy method.
I've done the following:

- Removed ObjectExtensions class.
- Added [DeepCloner](https://github.com/force-net/DeepCloner) package.
- Replaced references to ObjectExtensions.DeepCopy with calls to DeepCloner equivalent.

After the above changes, the Group Filter population step takes about 5 mins (so for me it was roughly 9x faster)


ElementalCrisis:
Awesome work! 

I know @maxpiva is starting work on merging Nancy in so I'll let him review and merge, don't want to delay his process. :)

We actually had some questions for you, when you have a moment hop on Discord.

PR #470, opened 2016-10-10
Improved performance of RecreateAllGroups operation

pmcleish:
- Moved RecreateAllGroups operation into it's own class.
- Auto group creation based on aniDB relations moved into AutoAnimeGroupCalculator class.
- Updated WCF and Nancy endpoints to use new RecreateAllGroups logic.
- Modified PocoCache to allow Clearing.
- BaseCachedRepository.Populate no longer key selector delegate. A SelectKey method is used instead (prevents caller needing to know what key to use).
- Changed some return types/method parameters of BaseXXXRepository from List to IReadOnlyXXX instead (This was intended for a reason that I didn't get around to implementing).

NOTE: This will need testing by someone that has more understanding of what the RecreateAllGroups operation is supposed to do.
I've tried to replicate the logic as best as I could, but some of it I didn't quite understand or wasn't sure if it was necessary. I've also had to bypass all the standard entity logic (e.g. Saving, Deleting, etc), so I'm hoping I've covered the existing logic the best I could.

The things I've tested and seem to work:
- Recreating all groups both with and without auto grouping based on aniDB relations.
- Checked to make sure the relation and anime type exclusions influence auto grouping.
- Changed language settings (e.g. primary language set to Kanji) and made sure the group names reflect the language choice.

I've also test with both SQL Server and SQLite (haven't tested with MySQL yet).

The original code for me took **5h03m** (Not auto grouping using relations) using SQL Server (1130 series, 3713 aniDB anime).
The new code now takes around **6 to 9 seconds**. Note that the new code uses parallelism for a few things, on my machine that made it 2 to 3 times faster than without.


da3dsoul:
Sweetness! I'll test it upon returning home.

da3dsoul:
If you had questions, you could've asked. The code is a mess, we know.

da3dsoul:
@maxpiva, you wrote the cache and know more about db than me at least, so it'd be nice if you would look over those parts and make sure it's all good.

da3dsoul:
But Christ, 6-9 seconds!

pmcleish:
Yeah, just a bit of an improvement :)
Assuming it works as expected.

pmcleish:
Another thing I forgot to mention is to possibly check the auto grouping based on relation logic.
I've added some very simple main title checking so that things like _Fate/Kaleid Liner Prisma Illya_ and _Fate/Stay Night_ go into different groups. That might need an option to disable or something if it isn't wanted.

da3dsoul:
I noticed that. That might cause some issues on shows that have season 1 and season 2 with different names, unless you made it only happen with relation types other than sequel/prequel.

da3dsoul:
An example I can think of would be Selector Infected WIXOSS, Selector Spread WIXOSS, etc

pmcleish:
No, I have separate logic for prequels/sequels/etc.

da3dsoul:
K

pmcleish:
the title checking is only for things like other/alternate setting/etc. relations

da3dsoul:
There'll need to be a setting for it, though. I want my stuff grouped if there is even a slight relation. I'll handle the setting, though, so don't worry about it.

pmcleish:
ok, no problem. You'd probably need to add the option check in AutoAnimeGroupCalculator.ShouldConsiderAnimeRelation (possibly same if statement as line 343).

da3dsoul:
OK thanks. Just to be sure, this does run in the importer for newly added series, right?

pmcleish:
Ummmmm. That is possibly something I've missed?

da3dsoul:
I know it needs to, otherwise you'd need to run it after every new series is added

da3dsoul:
At the very least, it needs to have the same logic, so groups don't get screwy when a new series is added.

pmcleish:
I've found the code that needs updating and have a reasonable idea of what I need to do to update it (I think). I probably won't be able to have a go at it until next weekend though.
In the meantime, you could have a test of what is here and let me know what I got wrong :)

da3dsoul:
OK, if that long, then it'll prolly need to wait until 3.7.0.1 bugfix update, depending on how soon we can get the other major issues cleaned up. Still, though, this'll be nice.

da3dsoul:
I'm merging into a test branch, where we can modify it further, and then merge into master. This purely makes it easier to clone and push to

JMMServer/Tasks/AutoAnimeGroupCalculator.cs
```diff
@@ -0,0 +1,594 @@
...
+            public int Sequels;
+            public int AlternateVersions;
+            public int FullStory;
+            public int ParentStory;
+
+            /// <summary>
+            /// Gets the calculated score for a particular anime.
+            /// The higher the score, the higher the priority this anime should be given.
+            /// </summary>
+            public int Score
```
da3dsoul: This is quite interesting, and I'll need to play with it. I'll need to Make a setting for those who purely want the first aired to be the main, though. I'll also need to look into using said setting and have the desired style for all clients as well.

JMMServer/Tasks/AutoAnimeGroupCalculator.cs
```diff
@@ -0,0 +1,594 @@
...
+
+            return title.Split(Whitespace, StringSplitOptions.RemoveEmptyEntries);
+        }
+
+        public AutoGroupExclude Exclusions
+        {
+            get { return _exclusions; }
+        }
+
+        [DebuggerDisplay("{AnimeNode.AnimeId} (Score: {Score})")]
```
da3dsoul: I didn't know you could do that. Thank you.

JMMServer/Tasks/AutoAnimeGroupCalculator.cs
```diff
@@ -0,0 +1,594 @@
...
+            }
+            else // Specified anime doesn't appear to have any relations
+            {
+                _animeGroupMap[animeId] = mainAnimeId = animeId;
+            }
+
+            return mainAnimeId;
+        }
+
+        private static int FindSuitableAnimeForGroup(Dictionary<int, RelationNode> nodes, HashSet<RelationEdge> edges)
```
da3dsoul: In this, or a different method toggled by the score setting, I'll need to have either score or first air date, determined by setting.

JMMServer/Tasks/RecreateAllGroupsTask.cs
```diff
@@ -0,0 +1,373 @@
...
+
+            return tempGroup;
+        }
+
+        /// <summary>
+        /// Deletes the anime groups and user mappings as well as resetting group filters and moves all anime series into the specified group.
+        /// </summary>
+        /// <param name="session">The NHibernate session.</param>
+        /// <param name="tempGroupId">The ID of the temporary anime group to use for migration.</param>
+        private void ClearGroupsAndDependencies(ISessionWrapper session, int tempGroupId)
```
da3dsoul: This prolly saved a good 5 minutes.

JMMServer/Tasks/RecreateAllGroupsTask.cs
```diff
@@ -0,0 +1,373 @@
...
+            _log.Info("AnimeGroups have been removed and GroupFilters have been reset");
+        }
+
+        private void UpdateAnimeSeriesContractsAndSave(ISessionWrapper session, IReadOnlyCollection<AnimeSeries> series)
+        {
+            _log.Info("Updating contracts for AnimeSeries");
+
+            // Update batches of AnimeSeries contracts in parallel. Each parallel branch requires it's own session since NHibernate sessions aren't thread safe.
+            // The reason we're doing this in parallel is because updating contacts does a reasonable amount of work (including LZ4 compression)
+            Parallel.ForEach(series.Batch(DefaultBatchSize), new ParallelOptions { MaxDegreeOfParallelism = 4 },
```
da3dsoul: Should this be updated to match logical core count on the cpu, or is 4 a number decided via other means?

JMMServer/Tasks/RecreateAllGroupsTask.cs
```diff
@@ -0,0 +1,373 @@
...
+
+        private void UpdateAnimeGroupsAndTheirContracts(ISessionWrapper session, IReadOnlyCollection<AnimeGroup> groups)
+        {
+            _log.Info("Updating statistics and contracts for AnimeGroups");
+
+            var allCreatedGroupUsers = new ConcurrentBag<List<AnimeGroup_User>>();
+
+            // Update batches of AnimeGroup contracts in parallel. Each parallel branch requires it's own session since NHibernate sessions aren't thread safe.
+            // The reason we're doing this in parallel is because updating contacts does a reasonable amount of work (including LZ4 compression)
+            Parallel.ForEach(groups.Batch(DefaultBatchSize), new ParallelOptions { MaxDegreeOfParallelism = 4 },
```
da3dsoul: Noting. Same as above thread question.

JMMServer/Tasks/RecreateAllGroupsTask.cs
```diff
@@ -0,0 +1,373 @@
...
+                    },
+                localFinally: localSession => { localSession.Dispose(); });
+
+            _animeGroupRepo.UpdateBatch(session, groups);
+            _log.Info("AnimeGroup statistics and contracts have been updated");
+
+            _log.Info("Creating AnimeGroup_Users and updating plex/kodi contracts");
+
+            // The reason we're doing this in parallel is because updating contacts does a reasonable amount of work (including LZ4 compression)
+            Parallel.ForEach(allCreatedGroupUsers.SelectMany(groupUsers => groupUsers), groupUser =>
```
da3dsoul: This is missing the Parallel options. An error, or a reason?

JMMServer/Tasks/RecreateAllGroupsTask.cs
```diff
@@ -0,0 +1,373 @@
...
+        {
+            _log.Info("Updating Group Filters");
+
+            IReadOnlyList<GroupFilter> grpFilters = _groupFilterRepo.GetAll(session);
+            ILookup<int, int> groupsForTagGroupFilter = _groupFilterRepo.CalculateAnimeGroupsPerTagGroupFilter(session);
+            IReadOnlyList<JMMUser> users = _userRepo.GetAll();
+
+            // The main reason for doing this in parallel is because UpdateEntityReferenceStrings does JSON encoding
+            // and is enough work that it can benefit from running in parallel
+            Parallel.ForEach(grpFilters.Where(f => ((GroupFilterType)f.FilterType & GroupFilterType.Directory) != GroupFilterType.Directory), filter =>
```
da3dsoul: This is also missing parallel options.

JMMServer/Tasks/RecreateAllGroupsTask.cs
```diff
@@ -0,0 +1,373 @@
...
+                AnimeSeries mainSeries = groupAndSeries.FirstOrDefault(series => series.AniDB_ID == mainAnimeId);
+                AnimeGroup animeGroup = new AnimeGroup();
+                string groupName = null;
+
+                if (mainSeries != null)
+                {
+                    animeGroup.Populate(mainSeries, now);
+                    groupName = mainSeries.GetSeriesName(session);
+                }
+                else // The anime chosen as the group's main anime doesn't actually have a series
```
da3dsoul: Good fallback, but if the series is null, that is an issue. AniDB_Anime.CreateAnimeSeriesAndGroup should called on every AniDB_Anime, so lacking an AnimeSeries is a huge problem that will cause an issue elsewhere.

JMMServer/Tasks/AutoAnimeGroupCalculator.cs
```diff
@@ -0,0 +1,594 @@
...
+            {
+                return false;
+            }
+            if (rel.RelationType != AnimeRelationType.AlternativeSetting && rel.RelationType != AnimeRelationType.Other
+                && rel.RelationType != AnimeRelationType.SameSetting)
+            {
+                return true;
+            }
+
+            // Perform a very poor man's string metric test.
```
da3dsoul: Also good, but will need setting to allow lazy grouping (group everything).

da3dsoul:
To remind myself: The score setting will need to be implemented into GetSeries and GetAllSeries sort, as well as contract updating if there is a reference to ser.AirDate sorting

JMMServer/Repositories/BaseCachedRepository.cs
```diff
@@ -115,7 +126,7 @@ public virtual void Delete(T cr)
                 EndDeleteCallback?.Invoke(cr);
             }
         }
-        public virtual void Delete(List<T> objs)
+        public virtual void Delete(IReadOnlyCollection<T> objs)
```
maxpiva: First of all, great work, So far, so good i'm only worried about the use of IReadOnlyCollection on the Save and Delete function of base Repository, the save collection function is a helper in fact, it should accept IEnumerable instead List, I only used list because of the free count. This force the user to convert any Enumerable into IReadOnlyCollection and i don't see any performance gain of doing so.

da3dsoul:
True IReadOnlyCollection is generally a limiter for the developer to know that the collection should not be modified from here, but in case I agree it probably isn't necessary here

JMMServer/Tasks/RecreateAllGroupsTask.cs
pmcleish: I limited that because each task/thread is opening a new NHibernate session. I didn't want to overburden the database with concurrent queries, etc.
For example, if the CPU had 32 cores, then it could well open 32 connections to the database concurrently, and you'd end up running into IO contention, etc. at that point (for SQLite anyway, SQL Server, MySQL would probably be serving queries from data in memory).

I just chose 4 as an arbitrary number. I did test doubling it and didn't notice any difference on my machine (8 logical cores).

JMMServer/Tasks/RecreateAllGroupsTask.cs
pmcleish: This and the below Parallel didn't use NHibernate sessions (like the above two) as far as I could tell by tracing through the code, so didn't see any reason to impose a limit.

JMMServer/Repositories/BaseCachedRepository.cs
pmcleish: You're right it could have been IEnumerable. I chose IReadOnlyCollection because it was the lowest interface that had a Count property which the current implementation was using (as you mentioned).

JMMServer/Tasks/RecreateAllGroupsTask.cs
pmcleish: Ah, there is actually a reason for that. I should have explained how my auto grouping works. Before I get to that, I'll mention why I modified the algorithm in the first place.
When I ran the existing implementation on my database, I ended up with groups such as:
- Initial D Fifth Stage
- Yuru Yuri 2
- Minami-ke Omatase
- Chihayafuru 2
- Durarara!!x2 Shou
- Gekijouban Toaru Majutsu no Index: Endymion no Kiseki
etc.

Which didn't seem right to me. My version basically builds the groups as needed and tries to pick a "master" anime to use for generating each group name (NOTE: It doesn't look at the series, but looks at the AniDB_Anime and relations). For the same database I ended up with the below anime chosen for the group names (with the series I actually have in JMM. I only have series from the last few years in my JMM database, everything else was in aniDB already and sitting on a file server):
- Initial D
  - Initial D Fifth Stage
  - Initial D Final Stage
- Yuru Yuri
  - Yuru Yuri 2
  - Yuru Yuri Nachuyachumi!
  - Yuru Yuri San Hai!
  - Yuru Yuri Nachuyachumi!+
- Minami-ke
  - Minami-ke Tadaima
  - Minami-ke Omatase
- Chihayafuru
  - Chihayafuru 2
  - Chihayafuru 2: Waga Mi Yo ni Furu Nagame Seshi Ma ni
- Durarara!!
  - Durarara!!x2 Shou
  - Durarara!!x2 Ten
  - Durarara!!x2 Ketsu
- Toaru Majutsu no Index
  - Toaru Majutsu no Index II
  - Toaru Kagaku no Railgun OVA
  - Gekijouban Toaru Majutsu no Index: Endymion no Kiseki
  - Toaru Kagaku no Railgun S
  - Toaru Kagaku no Railgun S: Daiji na Koto wa Zenbu Sentou ni Osowatta

So as you can see, for my database, it has actually chosen to use Anime that I don't actually have a series of (hence the reason for the fallback). I can try and explain how it works further if you want me to.
I've uploaded a simple test project that I used for experimentation purposes here:
[pmcleish/JMMAutoGroupTest](https://github.com/pmcleish/JMMAutoGroupTest)
You might be able to play around with that to see how it works as well.

JMMServer/Tasks/RecreateAllGroupsTask.cs
da3dsoul: Fair enough. I'm on a nice RAID, so db actions are decently fast, but I also use MSSQL, so most are probably pulling from RAM

JMMServer/Tasks/RecreateAllGroupsTask.cs
da3dsoul: Okay, so it defaults to logical core count?

pmcleish: I'm not sure. I think it relies on various factors. The best I can do is point you to [this stack overflow question](http://stackoverflow.com/questions/18943703/how-many-threads-parallel-forforeach-will-create-default-maxdegreeofparalleli).
I know when I was debugging I saw 4 threads working on a couple of those loops (might have been the ones I limited, can't remember now).

PR #472, opened 2016-10-13
Share RecreateAllGroups Logic with CreateAnimeSeriesAndGroup

pmcleish:
- Updated AniDB_Anime.CreateAnimeSeriesAndGroup to use the same
  grouping logic as RecreateAllGroups.
- Updated AutoAnimeGroupCalculator to allow configuring what relation
  types use fuzzy title matching (Defaults to "secondary" relation
  types. e.g. Alternate setting, Other, etc).
- Added minimum air date strategy for selecting anime to use for group
  name and made it the default (Gives very similar results to the
  "weighted" approach I added, so probably no need to support it.
  Have left the strategy in just in case though).

I tested adding 10 new series that weren't in my database yet while auto grouping was enabled and they all went into the groups I was expecting them to. But, some further testing would still be needed of course.

Also, I think a server setting might be required to control the relationsToFuzzyTitleTest parameter of the AutoAnimeGroupCalculator (it could be added to the Create method in a similar way exclusions setting).
Either that or you could set it to whatever default is deemed best.


da3dsoul:
Okay, I've found two issues.
1. JMM Desktop hangs perpetually after clicking the button @maxpiva we'll need input on that
2. Some filters (TvDB Links Missing is what I noticed) are filled with empty groups or groups that don't belong. I suspect it's an issue updating contracts as I have that filter set to the series level and it has empty results.

da3dsoul:
Definitely a contract updating issue. My kodi contracts are fucked. The groups and series are there, but no info on anything. Banners and Fanarts, which are populated on request instead of by contract are the only working things.

da3dsoul:
Contracts regenerated. I would have been upset, and then remember why when I go to watch something later.

pmcleish:
@da3dsoul is there something I need look into? or has your recent commit fixed the contract issue?

da3dsoul:
No I haven't fixed it. I've been fixing other bugs and keeping this branch updated.

pmcleish:
ah ok, How do I go about seeing the problem? I assume I'll need to setup kodi (I've never used it)? since it was its contracts you mentioned? I don't think I touched kodi contract generation, so it they're not right, I must not be setting something up correctly before generating the contracts. I'll take a look once I know how to reproduce the issue.

da3dsoul:
Setting up Kodi and testing would be the right way. The lazy way would be to use the master branch, Regen contracts, and call arbitrary XML and save them for comparison. http://localhost/JMMServerKodi/GetMetadata/1/0/9.XML is an example URL

da3dsoul:
Specifically, look at Series.updatestats and Group.Updatefromtoplevel

da3dsoul:
Those names aren't accurate, just enough to know what method

da3dsoul:
Another way good way to check is to run a code coverage debug and check what is missing

pmcleish:
That's a pretty good idea. I've checked the XML you mentioned and can certainly see there are issues. I'll setup Kodi tomorrow and start investigating.

PR #473, opened 2016-10-13
Update with master

da3dsoul:


da3dsoul:
Man that appveyor thing is cool. Slow, but it doesn't need to be fast.

PR #477, opened 2016-10-15
Fixed kodi/plex contracts for RecreateAllGroups

pmcleish:
Follow up to #472.

Fixed incorrect kodi/plex contract caused by stale/empty caches during RecreateAllGroups process. Basically when creating the kodi/plex contracts, the contract creation code couldn't find the Groups/Group_Users because the cache wasn't up to date.

The only issue I'm seeing now is the Art attribute isn't populated in the XML. However, I also checked the contracts on a database that I had run the original RecreateAllGroups code on (i.e. before my rewrite) and it was missing it as well. So, the issue seems to have possibly existed before my rewrite? thoughts/confirmation?

### DIFF
--- a/JMMServer/PlexAndKodi/CommonImplementation.cs
+++ b/JMMServer/PlexAndKodi/CommonImplementation.cs
@@ -1130,7 +1130,8 @@ namespace JMMServer.PlexAndKodi
                     Dictionary<Contract_AnimeGroup, Video> order = new Dictionary<Contract_AnimeGroup, Video>();
                     if (gf.GroupsIds.ContainsKey(userid))
                     {
-                        foreach (AnimeGroup grp in gf.GroupsIds[userid].Select(a => RepoFactory.AnimeGroup.GetByID(a)).Where(a => a != null))
+                        // NOTE: The ToList() in the below foreach is required to prevent enumerable was modified exception
+                        foreach (AnimeGroup grp in gf.GroupsIds[userid].ToList().Select(a => RepoFactory.AnimeGroup.GetByID(a)).Where(a => a != null))
                         {
                             Video v = grp.GetPlexContract(userid)?.Clone<Directory>();
                             if (v != null)
--- a/JMMServer/Repositories/Cached/AnimeGroup_UserRepository.cs
+++ b/JMMServer/Repositories/Cached/AnimeGroup_UserRepository.cs
@@ -138,6 +138,39 @@ namespace JMMServer.Repositories.Cached
             }
         }
 
+        /// <summary>
+        /// Inserts a batch of <see cref="AnimeGroup_User"/> into the database.
+        /// </summary>
+        /// <remarks>
+        /// <para>It is up to the caller of this method to manage transactions, etc.</para>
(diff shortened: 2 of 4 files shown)


da3dsoul:
As long as art is populated for episodes, it's all good. Like I said, fanart and banners is generated in request

da3dsoul:
Okay, posters aren't populated in the kodi contracts, everywhere except in episodes (different source, and not really posters). Otherwise it looks good. Contracts other than kodi don't seem to populated correctly, though (binary/desktop). After looking into it further, JMM Desktop hangs because it's throwing a nullpointer, catching throwing a different one, and recursively repeating, because things that aren't supposed to be null are.

pmcleish:
I'll look into the JMM Desktop hanging issue. Do you have any details regarding what is null; the contract in question; and roughly where in the desktop code I should be looking?
I should be able to figure it out myself if you're unable to get back to me before I start.

Regarding the posters: That would be caused by the missing Art attribute I mentioned in the PR description, yeah? Do you have a quick summary of where it is getting the Art from?
I found the same behavior before I rewrote the RecreateAllGroups stuff. So, I'm just wondering if it is caused by a bug I might have introduced into the AniDB_Anime contract generation a while back.

da3dsoul:
I don't think so, the poster is retrieved from AniDB_Anime.GetDefaultPosterNoBlanks. The desktop thing looks like a change tracker issue. The contract itself is null. It starts at MainListHelperVM.UpdateAll(), and throws null pointers and key not found exceptions all over the place.

da3dsoul:
The http client thing confused me for awhile. The method would be equivalent in JMMServerImplementation for whatever is called on the http client

da3dsoul:
In the case of RecreateAllGroups, I would accept marking everything as changed in batch, as it's likely that everything has. There's the possibility of only a few changing, though.

pmcleish:
I've modified the server to correctly record the removal of AnimeGroup_User and AnimeGroups. I also had to modify the MainListHelperVM.UpdateAll() in the desktop because when it asked the server for the changes, it was trying to update series information first (which uses AllGroupsDictionary). The problem was that AllGroupsDictionary was modified with the changes after the Series had attempted Populate (which resulted in key reference errors). I modified it so that Group changes are processed first, then Series. This seems to work and the AllGroupsDictionary and AllSeriesDictionary seem to have the correct entries by the end of MainListHelperVM.UpdateAll.

The problem I have now is that after running "Recreate All Groups" from the desktop client, it still sometimes freezes when I switch to the Collections tab. I've attached the exception from my log file (it is repeated a LOT).
[JMMDesktop_log.txt](https://github.com/japanesemediamanager/jmmserver/files/537803/JMMDesktop_log.txt)
I'm not familiar enough with WPF or the JMM desktop codebase to easily track this down. Any ideas?

da3dsoul:
Looks like it wants you to tell it you are modifying it with an event or something. @maxpiva knows more about wpf. I prefer not to touch it as I know next to nothing about it either.

da3dsoul:
Although I will admit it is odd, as it refreshed just fine before, even when removing groups.

PR #484, opened 2016-10-20
Kodi image & AnimeGroup change tracking fixes

pmcleish:
- Kodi images should now be working correctly.
- Deleting all AnimeGroups/AnimeGroup_Users will now be tracked by their change trackers.
- Added dev nancy API for extracting contracts in a readable form (nancy route is only available in DEBUG compiles. Could be put behind a config setting maybe).

(Requires japanesemediamanager/jmmclient#482 to test with JMM Desktop)
(Also: maybe regen of AniDB_Anime and AnimeSeries contracts if it doesn't initially work as expected)

### DIFF
--- a/JMMServer/API/APIv2_core_Module.cs
+++ b/JMMServer/API/APIv2_core_Module.cs
@@ -16,6 +16,8 @@ using Newtonsoft.Json;
 using System.IO;
 using JMMServer.Repositories.Direct;
 using JMMServer.Utilities;
+using JMMServer.Tasks;
+using Nancy.Responses;
 
 namespace JMMServer.API
 {
@@ -181,6 +183,11 @@ namespace JMMServer.API
             Post["/log/rotate"] = x => { return SetRotateLogs(); };
             Get["/log/rotate"] = x => { return GetRotateLogs(); };
             Get["/log/rotate/start"] = x => { return StartRotateLogs(); };
+
+            // 20. Dev
+#if DEBUG
+            Get["/dev/contracts/{entity?}"] = x => { return ExtractContracts((string)x.entity); };
+#endif
         }
 
         JMMServiceImplementationREST _rest = new JMMServiceImplementationREST();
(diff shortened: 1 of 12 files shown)


da3dsoul:
AnimeSeries should have at least part of the contracts regenerated on group recreation. AniDB_Anime however likely does not get written at all.

JMMServer/API/APIv2_core_Module.cs
```diff
@@ -180,7 +182,12 @@ public APIv2_core_Module() : base("/api")
             Get["/log/get/{max}/{position}"] = x => { return GetLog((int)x.max, (int)x.position); };
             Post["/log/rotate"] = x => { return SetRotateLogs(); };
             Get["/log/rotate"] = x => { return GetRotateLogs(); };
-            Get["/log/rotate/start"] = x => { return StartRotateLogs(); };
+            Get["/log/rotate/start"] = x => { return StartRotateLogs(); };
+
+            // 20. Dev
+#if DEBUG
+            Get["/dev/contracts/{entity?}"] = x => { return ExtractContracts((string)x.entity); };
```
da3dsoul: great idea. We could expand this concept a lot

JMMServer/Tasks/ContractExtractor.cs
```diff
@@ -0,0 +1,233 @@
...
+        /// <exception cref="ArgumentNullException"><paramref name="session"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">The specified <paramref name="entityType"/> is not valid.</exception>
+        public System.IO.Stream GetContractsAsZipStream(ISessionWrapper session, string entityType)
+        {
+            if (session == null)
+                throw new ArgumentNullException(nameof(session));
+
+            MemoryStream buffer = new MemoryStream();
+
+            using (ZipArchive zip = new ZipArchive(buffer, ZipArchiveMode.Create, leaveOpen: true))
```
da3dsoul: Is there a reason you didn't use Gzip? I've just notice we use gzip instead everywhere else.

pmcleish: Because I'm dumping a json file per entity. gzip doesn't support file structure. I'd have to tar it first if I was using gzip.

da3dsoul: and you just explained something else I've wondered about for years....why tar.gz exists

PR #491, opened 2016-10-23
Fix for Upload Watched Status to MAL

monsieurtaco:
Temporary / in-progress fix for now. The fix works as it should, but it could be improved. Would love to get some input on this and make sure it works for others as well. Thanks!

@hidden4003

### DIFF
--- a/JMMServer/Providers/MyAnimeList/MALHelper.cs
+++ b/JMMServer/Providers/MyAnimeList/MALHelper.cs
@@ -400,10 +400,10 @@ namespace JMMServer.Providers.MyAnimeList
 
                 // Populate MAL animelist hashtable if isNeverDecreaseWatched set
                 Hashtable animeListHashtable = new Hashtable();
+                myanimelist malAnimeList = GetMALAnimeList();
                 if (ServerSettings.MAL_NeverDecreaseWatchedNums)
                     //if set, check watched number before update: take some time, as user anime list must be loaded
                 {
-                    myanimelist malAnimeList = GetMALAnimeList();
                     if (malAnimeList != null && malAnimeList.anime != null)
                     {
                         for (int i = 0; i < malAnimeList.anime.Length; i++)
--- a/JMMServer/Providers/MyAnimeList/MALHelper.cs
+++ b/JMMServer/Providers/MyAnimeList/MALHelper.cs
@@ -482,8 +482,6 @@ namespace JMMServer.Providers.MyAnimeList
                                 lastWatchedEpNumber = epNum;
                             }
 
-
-                            /* Commented this out for now; errors from this making it unable to upload.
                             List<Contract_VideoDetailed> contracts = ep.GetVideoDetailedContracts(user.JMMUserID);
 
                             // find the latest episode number in the collection
@@ -495,7 +493,7 @@ namespace JMMServer.Providers.MyAnimeList
                                 if (!string.IsNullOrEmpty(contract.AniDB_Anime_GroupNameShort) &&
                                     !fanSubGroups.Contains(contract.AniDB_Anime_GroupNameShort))
                                     fanSubGroups.Add(contract.AniDB_Anime_GroupNameShort);
-                            }*/
+                            }
                         }
                     }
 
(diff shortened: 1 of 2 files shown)


JMMServer/Providers/MyAnimeList/MALHelper.cs
```diff
@@ -530,15 +532,21 @@ public static void UpdateMALSeries(AnimeSeries ser)
                             ser.GetAnime().AnimeID);
                         continue;
                     }
+
+                    string confirmationMessage = "";
+                    if (animeListHashtable.ContainsKey(malID))
```
hidden4003: I am not sure we can rely on this as MAL list can be cached, there is an open issue on their forums about anime list not always returning latest data.

monsieurtaco: @hidden4003 Actually, I think most things are back to normal, but I'm still having issues updating series. As a test, I removed some series from my MAL and adjusted some from Completed with votes to Watching with random numbers or 0 episodes watched with or without ratings. The log stated that they were successfully updated with the correct mal id, eps, and score, but my MAL had changes made to the series that I removed. Any series that I changed and needed to be updated using ModifyAnime() were once again not changed.

hidden4003: I don't think there is an existing condition, I am just pointing out that in case of mass update this API call will most likely give you outdated list thus it can fail in case where you delete series but MAL still says it is there, so it will try to update instead of add and will fail.
We can try your code and see if there will be any issues, I am not against it or anything it is just I don't see it being more reliable than existing one.

monsieurtaco: > I am not sure we can rely on this as MAL list can be cached, there is an open issue on their forums about anime list not always returning latest data.

I see what you mean. With the animeListHashTable, it is only populated if we have Never Decrease Series Status set to True. If it is false, then the Upload function will just default to calling "Add" for every series, which is not ideal as that is the current broken behavior. If it is true, then everything works fine unless mysteriously on the MAL end the anime list is not returning latest data which is out of our hands.

monsieurtaco: > I don't think there is an existing condition, I am just pointing out that in case of mass update this API call will most likely give you outdated list thus it can fail in case where you delete series but MAL still says it is there, so it will try to update instead of add and will fail.

That's definitely a good point. Due to a lack of good ways to control using Add or Modify, we could always fall back to calling both commands one after the other. While I think it's wasteful to do it that way, I still think it's better than having un-updateable series in MAL. Thanks for taking the time to look at this though - criticism is welcome and I just want to see all these series that were stuck without watch progress / votes in MAL get updated again. 

By the way, there actually appears to be some disturbing behavior regarding Update. I was testing the case you're talking about by attempting to use MAL API Update on a series that is not in my MAL. I ran an Update call through curl and updated series 2168 with 1 episode and watched status. MAL API responded that this was SUCCESSFUL, it even added it to my anime history
"Sakura Taisen: New York ep. 1"
but did not add it to my list / change the watched status. 

So in the case where we attempt to update a series that is no longer / never been in our MAL, it will do nothing and create a phantom history for the command. I have not encountered this personally except in this particular test case, but you're right in that it could happen. However, due to MAL's lack of useful responses (201 Created / Updated being the only things they respond with), I am not sure what to do.

monsieurtaco: I think you were making the right choice, but it would seem that some time in the past couple of months the API responses changed :(

monsieurtaco:
I made the requested change. Please let me know if that was what you wanted. 

The condition I was using relies on the "animeListHashtable", so if we wanted to make sure that was always populated the proper change would be to instead just comment out 
`if (ServerSettings.MAL_NeverDecreaseWatchedNums)`, but that would make that server setting useless. Although since the commit I just pushed moved `myanimelist malAnimeList = GetMALAnimeList();` out of the if-block, I think it is still doing the same thing.

hidden4003:
Yes change is good.
It will not make the server setting useless, you probably misunderstood what setting does as it is worded poorly, the setting does not allow series become "Watching" on MAL it it is already "Completed" unless it was in "Plan to watch" state (6)

monsieurtaco:
@hidden4003 thanks for the review and merge. Let me know how it goes for you.

PR #570, opened 2017-02-02
Fix renamer when file don't need a rename

VadAbel:
Prevent empty filename for VideoLocal when the file don't need to be rename

This is link to issues 505 of ShokoClient


da3dsoul:
That looks like it works

PR #601, opened 2017-04-11
hasher_long_file_name_fix

bond4u:
prepend "disable long file name check" prefix to file name.


Cazzar:
In what does this resolve/improve?

da3dsoul:
It fixes the file not found for long paths in windows 10

bond4u:
This fixes hashing of files with long path - over 260 chars.
With "\\?\" prefix wstat64() and OpenFile() don't return error when file name with full path is over 260 chars.

PR #651, opened 2017-08-06
Remove System.Net.Http.dll from the Build Directory

KingJ:
@Cazzar 

On Mono 5.0, there is a bug with System.Net.Http.dll that prevents HTTP requests from working for certain bits of Shoko. For example, AniDB and TheTVDB work fine for me however TheMovieDB throws a `System.MissingMethodException: Method 'System.Net.Logging.get_On' not found.` exception and so parts of the collection can't be linked to TheMovieDB and banners/posters/fanart can't be downloaded.

The issue should be fixed in Mono 5.2, currently in beta, however a workaround for now is to remove `System.Net.Http.dll` from `/usr/src/app/build/`.  I've manually removed this from my container and restarted Shoko and it all appears to be working fine.

This PR adds a new command to the Dockerfile that removes `/usr/src/app/build/System.Net.Http.dll` after the build has completed. I'm thinking this is probably the easiest way to work around this without potentially affecting parts of the master branch?

Further reading: https://github.com/dotnet/corefx/issues/19914#issuecomment-302327100


Cazzar:
I forgot this happens

PR #663, opened 2017-09-01
fixing some null logic.

Jimmyson:
moved a return, and added some comments for empty catches for good measure too.
(Rider wanted those dealt with)


da3dsoul:
Just looks like general clean up. Merged

PR #702, opened 2018-01-29
Fix FileSystemWatcher thread leak and BlockingList synchronization.

bond4u:


da3dsoul:
I'll assume it's tested. Nice one

PR #742, opened 2018-06-22
CommandRequest query optimisations.

bond4u:
Take(1) and RowCount() instead of just List().

### DIFF
--- a/Shoko.Server/Repositories/Direct/CommandRequestRepository.cs
+++ b/Shoko.Server/Repositories/Direct/CommandRequestRepository.cs
@@ -124,44 +124,31 @@ namespace Shoko.Server.Repositories.Direct
             {
                 using (var session = DatabaseFactory.SessionFactory.OpenSession())
                 {
-                    IList<CommandRequest> crs;
-
+                    HashSet<int> types = CommandTypesGeneral;
                     // This is called very often, so speed it up as much as possible
                     // We can spare bytes of RAM to speed up the command queue
                     if (ShokoService.AnidbProcessor.IsHttpBanned && ShokoService.AnidbProcessor.IsUdpBanned)
                     {
-                        crs = session.QueryOver<CommandRequest>()
-                            .WhereRestrictionOn(field => field.CommandType).IsIn(CommandTypesGeneralFullBan.ToArray())
-                            .OrderBy(cr => cr.Priority).Asc
-                            .ThenBy(cr => cr.DateTimeUpdated).Asc
-                            .List<CommandRequest>();
+                        types = CommandTypesGeneralFullBan;
                     }
                     else if (ShokoService.AnidbProcessor.IsUdpBanned)
                     {
-                        crs = session.QueryOver<CommandRequest>()
-                            .WhereRestrictionOn(field => field.CommandType).IsIn(CommandTypesGeneralUDPBan.ToArray())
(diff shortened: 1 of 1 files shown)


da3dsoul:
I appreciate the effort, but I'm going to reject this for one reason: It's going to need to be like that for Queue Factory, as there's going to need to be more complex selection of the next command.
The RowCount ones can stay

da3dsoul:
Actually I'll just tweak it a bit. Remove the obsolete tags as we'll need them later, and it's good

bond4u:
what's wrong with Take(1) ?

da3dsoul:
Because you can use UniqueResult instead

hidden4003:
So that is why large queues are so slow... @da3dsoul what is wrong with Take(1)? should be good we have full refactor planned in the near future anyways should be fine as a hotfix

Cazzar:
Personally speaking, I would refactor this a little more to dynamically generate the commands to not try.
This could be done via the annotation
https://github.com/ShokoAnime/ShokoServer/blob/c34eadfd61b3621258410fe7ea418c810fde77e3/Shoko.Server/Commands/CommandRequestImplementation.cs#L16-L22
Adding an enum/flag to say:
```csharp
[Flags]
enum AniDBBan {
    Http = 1, Udp = 2, None = 0
}
```
and filter it out on the attribute, as we would already scan the available types here:
https://github.com/ShokoAnime/ShokoServer/blob/c34eadfd61b3621258410fe7ea418c810fde77e3/Shoko.Server/Commands/CommandHelper.cs#L110-L127

If we classify them dynamically, I don't see why this wouldn't work.

At a minimum, we should limit the result count, but that's harder to manage.
This would also work for designing around to the Queue Factory setup, as with this, we could designate which queue it's in dynamically.

da3dsoul:
Nothing is wrong with Take(1). What is wrong is returning List instead of UniqueResult

da3dsoul:
It creates an enumerable and counts all 1 item instead of using the one item we asked for

bond4u:
In case of mysql, Take(1) resulted query with "limit 1".

da3dsoul:
Ugh

da3dsoul:
```
crs = session.QueryOver<CommandRequest>()
.WhereRestrictionOn(field => field.CommandType)
.IsIn(types.ToArray())
.OrderBy(cr => cr.Priority).Asc
.ThenBy(cr => cr.DateTimeUpdated).Asc
.Take(1)
.UniqueResult<CommandRequest>();
```

da3dsoul:
That'll error, but that's what I meant

bond4u:
CommandRequest cr = session
                        .CreateCriteria(typeof(CommandRequest))
                        .Add(Restrictions.In("CommandType", cTypes.ToArray()))
                        .AddOrder(Order.Asc("Priority"))
                        .AddOrder(Order.Asc("DateTimeUpdated"))
                        .SetMaxResults(1)
                        .UniqueResult<CommandRequest>();
UniqueResult seems to be doing some post-processing, because it throws when resultset contains more than one row. So You have to add SetMaxResults(1) which does "limit 1" query and You get single row.
But I don't like hardcoded strings.
QueryOver...Take(1).SingleOrDefault() with field names looks better.

da3dsoul:
Fair

PR #751, opened 2018-07-28
Fix for systems without file locking

xmodx10:
I've been running shoko server on docker for some time with a single import folder being watched that is also shared out via samba from my NAS. Every time I copied a file into the folder I would have to pause the hashing queue so it would not half hash the file before it finished the copy.

This code is to fix that issue by checking the last write time on the file. I thought of including the check in the existing while loop with something like:

`while ((filesize = CanAccessFile(FileName)) == 0 || FileModified(FileName,3) && (numAttempts < 60))`

But given that a file may take more than 60 seconds to copy I thought it best to separate it out.

Tested and it seems to work with 2 seconds but to be safe I went with 3. Ran some tests and it does not impact anything when running on Windows as this gets caught in the existing file access check.

This is my first time contributing to a project so any feedback/suggestions are welcome.

### DIFF
--- a/Shoko.Server/Commands/Import/CommandRequest_HashFile.cs
+++ b/Shoko.Server/Commands/Import/CommandRequest_HashFile.cs
@@ -88,6 +88,27 @@ namespace Shoko.Server.Commands
             }
         }
 
+        //Used to check if file has been modified within the last X seconds.
+        private bool FileModified(string FileName, int Seconds)
+        {
+            try
+            {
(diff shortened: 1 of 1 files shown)


da3dsoul:
I feel like we talked prior. I'll look at it

da3dsoul:
Looks good. @cazzar @hidden4003 no need to worry, this won't do anything for most people, as they've either already waited the appropriate amount of time, or they'll wait at most 1s for modified date to change

PR #769, opened 2018-12-29
Datatype fix for episode submission.

Vesnyx:
This fixes the issue I created in ShokoAnime/ShokoDesktop#593


Vesnyx:
I don't really know specifically where you'd like the enums to be used, so I just used them everywhere. :( 

Frankly, haven't used enums in a long time either. If there's a different way you want me to implement them just let me know.

da3dsoul:
I'm aware, but it's one of those things we'll leave alone, as the best way to take care of it is Ctrl+A,Delete. New API for AniDB should have better support, so we can rewrite it then

PR #814, opened 2020-04-02
Fix slow episode imports in large series

notnyt:
This adds an UpdatedAt field to AnimeSeries which is used to prevent excessive iteration on import


da3dsoul:
Cool the only change I'd make (aside from a complete redesign of the system) is something I'll just fix myself. One of the revision numbers should be 1 instead of 2. No biggie.

da3dsoul:
Haha already merged BRM. Thanks for looking, though

PR #824, opened 2020-05-19
Update dockerfile for the dotnet build

harshithmohan:
I'm gonna put this here before I accidentally delete these files.
Everything's working properly for me, just check if it's ok and if it is, time for new daily builds?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,7 @@ COPY --from=0 /usr/src/app/build .
 COPY ./dockerentry.sh /dockerentry.sh
 
 WORKDIR /usr/src/app/build/webui
+#RUN curl -L $(curl https://api.github.com/repos/ShokoAnime/ShokoServer-WebUI/releases | jq -r '. | map(select(.prerelease==false)) | .[0].assets[0].browser_download_url') -o latest.zip
 RUN curl -L $(curl https://api.github.com/repos/ShokoAnime/ShokoServer-WebUI/releases | jq -r '.[0].assets[0].browser_download_url') -o latest.zip
 RUN unzip -o latest.zip
 RUN rm latest.zip
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,54 +1,32 @@
-FROM mono:6.0
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1
 
 #MAINTAINER Cayde Dixon <me@cazzar.net>
 ENV PUID=1000 \
-    PGID=100  \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,54 +1,32 @@
...
 
-RUN apt-get update && apt-get install -y apt-utils libmediainfo0v5 librhash0 sqlite.interop jq unzip libunwind-dev && apt-get install -y -t stretch-backports gosu
+RUN apt-get update && apt-get install -y apt-utils gosu jq unzip libmediainfo0v5 librhash-dev libunwind-dev mono-runtime libmono-system-xml-linq4.0-cil
 
 WORKDIR /usr/src/app/build
 COPY --from=0 /usr/src/app/build .
 COPY ./dockerentry.sh /dockerentry.sh
 
 WORKDIR /usr/src/app/build/webui
-#RUN curl -L $(curl https://api.github.com/repos/ShokoAnime/ShokoServer-WebUI/releases | jq -r '. | map(select(.prerelease==false)) | .[0].assets[0].browser_download_url') -o latest.zip
```
Cazzar, line 51: Don't delete this line.

Dockerfile
```diff
@@ -1,54 +1,32 @@
-FROM mono:6.0
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1
 
 #MAINTAINER Cayde Dixon <me@cazzar.net>
 ENV PUID=1000 \
-    PGID=100  \
-    TargetFrameworkDirectory=/usr/lib/mono/
```
Cazzar, line 6: Why this directory?

Dockerfile
```diff
@@ -1,54 +1,32 @@
-FROM mono:6.0
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1
```
Cazzar, line 1: Any specific reason for mcr.microsoft?

Dockerfile
harshithmohan: Sorry, I just rewrote the whole file and copy-pasted some things so I think I missed it

Dockerfile
harshithmohan: As I said, rewrote.
What should that directory point to now? I will fix it

Dockerfile
harshithmohan: It's the official dotnet core image?
[https://hub.docker.com/_/microsoft-dotnet-core-sdk/](url)

Issue #837, opened 2020-09-02
Folder Scan Exclusion


Urlryn:
Feature request/enhancement:

I am looking to have a shoko/docker config option added that will allow you to specify folder or folders that would be excluded from being scanned.  Or maybe have shoko ignore hidden folders (though this option could be a problem for the anime series .hack) during scanning.

On Unraid there is a plugin that adds a hidden '.Recycle.Bin' folder to hold samba deleted files and will remove the aged files on a schedule. A '.Recycle.Bin' folder is created the first time a file is deleted. 

So in the case of Shoko....if you delete any file via shoko desktop the files would be moved to this .recycle.bin folder.  When the import process runs, while this folder contains these deleted files, it will see this and then re-import those files again.

Urlryn

xmodx10:
+1 for this feature as well. I also use Unraid and the ".Recycle.Bin" always causes me grief with Shoko, having to disable the folder watch to work around it every time is a pain.

Cazzar:
I spoke to @Urlryn about this issue beforehand, the best way we could do this is extend https://github.com/ShokoAnime/ShokoServer/blob/3144926b262ff00db253c21686f636c0f4e27510/Shoko.Server/Import/Importer.cs#L274 to be a array (maybe also regex) check instead of a hard-coded string, defaulting to that.

PR #838, opened 2020-09-04
Import Exclusions

xmodx10:
Thought I would have a go at #837 . Nothing fancy but I think it should get the job done. Open to feedback.

### DIFF
--- a/Shoko.Server/Import/Importer.cs
+++ b/Shoko.Server/Import/Importer.cs
@@ -271,7 +271,12 @@ namespace Shoko.Server
                         if (fldr.IsDropSource == 1)
                             dictFilesExisting[fileName].RenameAndMoveAsRequired();
                     }
-                    if (fileName.Contains("$RECYCLE.BIN")) continue;
+
+                    if (ServerSettings.Instance.Import.Exclude.Any(s => fileName.Contains(s)))
+                    {
+                        logger.Trace("Import exclusion, skipping --- {0}", fileName);
+                        continue;
+                    }
 
                     filesFound++;
                     logger.Trace("Processing File {0}/{1} --- {2}", i, fileList.Count, fileName);
@@ -317,8 +322,12 @@ namespace Shoko.Server
             foreach (string fileName in fileList)
             {
                 i++;
-                // ignore recycling bins
-                if (fileName.Contains("$RECYCLE.BIN") || fileName.StartsWith(".Trash-")) continue;
+
+                if (ServerSettings.Instance.Import.Exclude.Any(s => fileName.Contains(s)))
+                {
+                    logger.Trace("Import exclusion, skipping --- {0}", fileName);
+                    continue;
+                }
                 filesFound++;
                 logger.Trace("Processing File {0}/{1} --- {2}", i, fileList.Count, fileName);
 
@@ -385,7 +394,11 @@ namespace Shoko.Server
             List<string> fileListNew = new List<string>();
             foreach (string fileName in fileList)
             {
-                if (fileName.Contains("$RECYCLE.BIN") || fileName.StartsWith(".Trash-")) continue;
+                if (ServerSettings.Instance.Import.Exclude.Any(s => fileName.Contains(s)))
+                {
+                    logger.Trace("Import exclusion, skipping --- {0}", fileName);
+                    continue;
+                }
                 if (!dictFilesExisting.ContainsKey(fileName))
                     fileListNew.Add(fileName);
             }
@@ -432,7 +445,12 @@ namespace Shoko.Server
             foreach (string fileName in fileList)
             {
                 i++;
-                if (fileName.Contains("$RECYCLE.BIN")) continue;
+                if (ServerSettings.Instance.Import.Exclude.Any(s => fileName.Contains(s)))
+                {
+                    logger.Trace("Import exclusion, skipping --- {0}", fileName);
+                    continue;
+                } 
+
                 filesFound++;
                 logger.Trace("Processing File {0}/{1} --- {2}", i, fileList.Count, fileName);
 
--- a/Shoko.Server/Import/Importer.cs
+++ b/Shoko.Server/Import/Importer.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 using FluentNHibernate.Utils;
 using NLog;
 using NutzCode.CloudFileSystem;
@@ -272,7 +273,7 @@ namespace Shoko.Server
                             dictFilesExisting[fileName].RenameAndMoveAsRequired();
                     }
 
-                    if (ServerSettings.Instance.Import.Exclude.Any(s => fileName.Contains(s)))
+                    if (ServerSettings.Instance.Import.Exclude.Any(s => Regex.IsMatch(fileName,s)))
                     {
                         logger.Trace("Import exclusion, skipping --- {0}", fileName);
                         continue;
@@ -323,7 +324,7 @@ namespace Shoko.Server
             {
                 i++;
 
-                if (ServerSettings.Instance.Import.Exclude.Any(s => fileName.Contains(s)))
+                if (ServerSettings.Instance.Import.Exclude.Any(s => Regex.IsMatch(fileName,s)))
                 {
                     logger.Trace("Import exclusion, skipping --- {0}", fileName);
                     continue;
@@ -394,7 +395,7 @@ namespace Shoko.Server
             List<string> fileListNew = new List<string>();
             foreach (string fileName in fileList)
             {
-                if (ServerSettings.Instance.Import.Exclude.Any(s => fileName.Contains(s)))
+                if (ServerSettings.Instance.Import.Exclude.Any(s => Regex.IsMatch(fileName,s)))
                 {
                     logger.Trace("Import exclusion, skipping --- {0}", fileName);
                     continue;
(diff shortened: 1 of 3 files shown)


Shoko.Server/Server/ShokoServer.cs
```diff
@@ -658,7 +658,11 @@ void WorkerFileEvents_DoWork(object sender, DoWorkEventArgs e)
 
                                 foreach (string file in files)
                                 {
-                                    if (FileHashHelper.IsVideo(file))
+                                    if (ServerSettings.Instance.Import.Exclude.Any(s => file.Contains(s)))
+                                    {
+                                        logger.Info("Import exclusion, skipping file {0}", file);
```
Cazzar, line 663: This won't break out of the loop, you might need to add `continue;` after the log line.

Shoko.Server/Server/ShokoServer.cs
```diff
@@ -671,7 +675,11 @@ void WorkerFileEvents_DoWork(object sender, DoWorkEventArgs e)
                             {
                                 logger.Info("New file detected: {0}: {1}", evt.FullPath, evt.ChangeType);
 
-                                if (FileHashHelper.IsVideo(evt.FullPath))
+                                if (ServerSettings.Instance.Import.Exclude.Any(s => evt.FullPath.Contains(s)))
+                                {
+                                    logger.Info("Import exclusion, skipping file: {0}", evt.FullPath);
```
Cazzar, line 680: As above.

Shoko.Server/Import/Importer.cs
```diff
@@ -271,7 +271,12 @@ public static void RunImport_ScanFolder(int importFolderID, bool skipMyList = fa
                         if (fldr.IsDropSource == 1)
                             dictFilesExisting[fileName].RenameAndMoveAsRequired();
                     }
-                    if (fileName.Contains("$RECYCLE.BIN")) continue;
+
+                    if (ServerSettings.Instance.Import.Exclude.Any(s => fileName.Contains(s)))
```
Cazzar, line 275: Are we able to change this to a [Regex]? It would then allow us to handle a 'Contains' or an exact match.

[Regex]: https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex?view=netstandard-2.1

Cazzar (changes requested):

Shoko.Server/Server/ShokoServer.cs
xmodx10: I left that out because I assume it still needed to flow down to>>
`queueFileEvents.Remove(evt);`

It looks like an event loop/thread from what I could tell

Shoko.Server/Import/Importer.cs
xmodx10: In the settings would it make sense for it to allow a list of Regex strings or only allow one Regex string? , I ask because in a single Regex im fairly sure you can define mulitple conditions so to speak.

Cazzar: For easier user understanding, do multiple

Shoko.Server/Server/ShokoServer.cs
xmodx10: After spending some time looking over WorkerFileEvents_DoWork() and gaining a better understanding of it. It makes sense for it to just flow down/on.

xmodx10:
Switched things over to Regex. Did some testing and everything seems to work as expected.

da3dsoul:
You know...regex is a good idea here

Shoko.Server/Settings/ImportSettings.cs
```diff
@@ -8,6 +8,8 @@ public class ImportSettings
         
         public List<string> VideoExtensions { get; set; } = new List<string> { "MKV", "AVI", "MP4", "MOV", "OGM", "WMV", "MPG", "MPEG", "MK3D", "M4V" };
 
+        public List<string> Exclude { get; set; } = new List<string> { "$RECYCLE.BIN", ".Recycle.Bin", ".Trash-" };
```
Cazzar, line 11: Only comment here is I don't think it's only ".trash"
Do we want to constrain these as `^\$RECYCLE\.BIN^` as `$RECYCLE.BIN` won't actually match, I'll give more examples

Cazzar: https://regex101.com/r/6ynWJN/1 for example on the `$RECYCLE.BIN` example.
and https://regex101.com/r/nBx5Hq/1 for `.Recycle.Bin`

Cazzar: @da3dsoul thoughts?

Cazzar (changes requested):

Shoko.Server/Settings/ImportSettings.cs
xmodx10: Thanks for picking that up. I tried changing it to: 
`public List<string> Exclude { get; set; } = new List<string> { @"\$RECYCLE\.BIN", @"\.Recycle\.Bin", @"\.Trash-" };`

In the config file its getting serialized to:
`"Exclude": [
      "\\$RECYCLE\\.BIN",
      "\\.Recycle\\.Bin",
      "\\.Trash-"
    ],`

and then deserialized to:
image

I tried to see what would happen if I edit the config and remove the extra "\" and it then fails to deserialize:
`Shoko.Server.Settings.ServerSettings.Instance.**get** returned null.`

Any ideas ? I guess one way is to try and unescape the Regex string but having to use non-standard Regex in the config file so that it converts properly is not ideal from a user perspective.

Cazzar: That's fine, it's a side-effect of JSON as `\` is the escape character, The editing of the config is really mostly handled in webui anyway.

The main thing you will want to do is use `^` and `$` to anchor to the regexes.
So you want to do:
```suggestion
        public List<string> Exclude { get; set; } = new List<string> { @"^\$RECYCLE\.BIN$", "^\.Recycle\.Bin$", "^\.Trash-" };
```

xmodx10: In the WorkerFileEvents_DoWork() loop we are actually testing against the full path.
image
So that would not result in a match. Ill need to double check the other places to ensure its consistant. I'm not sure if there is a way to get a relative paths in WorkerFileEvents_DoWork(), quickly looking I could not see the info being passed to it.

Example >> https://regex101.com/r/rvVwdB/2

The problem is that the regex pattern we need to pass into the Regex.IsMatch function is for exmaple `\$RECYCLE\.BIN` but due to the JSON serialization/deserialization its coming through as the below:
image
This wont match and will be read as a backslash instead of an escape for the "$" and "."

xmodx10: Had a think about this and I believe a simple String.Replace should do the trick, basically replace "\\" with "\". Cant think of any edge cases at the momment where this may cause issues.

Cazzar: That shouldn't be a problem, you are looking at the escaped version in the debug tools:

```
$ cat Program.cs
using System;
using Newtonsoft.Json;

namespace aaa
{
    class Program
    {
        static void Main(string[] args)
        {
            var data = JsonConvert.SerializeObject(new { a = @"^\.RECYCLE\.BIN^" });
            Console.WriteLine(data);

            dynamic data1 = JsonConvert.DeserializeObject("{\"a\":\"^\\\\.RECYCLE\\\\.BIN^\"}"); // {"a":"^\\.RECYCLE\\.BIN^"} just escaped.
            Console.WriteLine(data1.a);

            dynamic deser = JsonConvert.DeserializeObject(data);
            Console.WriteLine(deser.a);
        }
    }
}
$ dotnet run
{"a":"^\\.RECYCLE\\.BIN^"}
^\.RECYCLE\.BIN^
^\.RECYCLE\.BIN^
```

xmodx10: ah! I dont even know.. swear I tested it and it wasn't matching but now its working. Thanks for that, will have to keep it in mind when using the debug tools. Its been a while since I touched visual studio.

Cazzar: Hmm, I would still prefer to match by the folder explicitly, just trying to work out a good idea for that.

xmodx10: What about this ? >> `(\\|\/)\$RECYCLE\.BIN(\\|\/)`
https://regex101.com/r/rvVwdB/3

This will keep it to folder only and match linux,windows,cifs/smb paths

Cazzar: After a bit of thought, leaving as full path allows better extensibility, so we can leave as is.

Shoko.Server/Settings/ImportSettings.cs
```diff
@@ -8,7 +8,7 @@ public class ImportSettings
         
         public List<string> VideoExtensions { get; set; } = new List<string> { "MKV", "AVI", "MP4", "MOV", "OGM", "WMV", "MPG", "MPEG", "MK3D", "M4V" };
 
-        public List<string> Exclude { get; set; } = new List<string> { "$RECYCLE.BIN", ".Recycle.Bin", ".Trash-" };
+        public List<string> Exclude { get; set; } = new List<string> { @"(\\|\/)\$RECYCLE\.BIN(\\|\/)", @"(\\|\/)\.Recycle\.Bin(\\|\/)" };
```
Cazzar, line 11: You'll want to leave `.Trash-` in there, as it's a trash folder that ubuntu uses, alongside other OS's

https://askubuntu.com/questions/282332/what-is-the-trash-1000-folder-and-how-to-delete-it

You'd want the regex: `[\\\/]\.Trash-\d+[\\\/]` (I am using [] instead of a capture group by habit)

xmodx10: Added it back in :)

PR #847, opened 2020-09-22
Set 'Name' on 'Shoko.Server.API.v3.Models.Shoko.Episode' in constructor

revam:
Set the `Name` property of `Shoko.Server.API.v3.Models.Shoko.Episode` mimicing the magic formula used in ShokoDesktop, but added two extra fallback languages.

[Shoko.Desktop/ViewModel/Server/VM_AniDB_Episode.cs#L12-L20/](https://github.com/ShokoAnime/ShokoDesktop/blob/master/Shoko.Desktop/ViewModel/Server/VM_AniDB_Episode.cs#L12-L20)

### DIFF
--- a/Shoko.Server/API/v3/Models/Shoko/Episode.cs
+++ b/Shoko.Server/API/v3/Models/Shoko/Episode.cs
@@ -51,24 +51,12 @@ namespace Shoko.Server.API.v3.Models.Shoko
 
             var uid = ctx.GetUser()?.JMMUserID ?? 0;
             Watched = ep.GetUserRecord(uid)?.WatchedDate;
-            Name = GetEpisodeName(anidb);
+            Name = ep.Title;
 
             Size = ep.GetVideoLocals().Count;
         }
 
         internal static string[] TitleLanguages = new string[] { "EN", "X-JAT", "X-ZHT", "X-OTHER" };
-
-        public static string GetEpisodeName(AniDB_Episode a)
-        {
-            if (a == null) return "Episode 0";
-            var titles = RepoFactory.AniDB_Episode_Title.GetByEpisodeID(a.EpisodeID);
-            if (titles != null) foreach (string lang in TitleLanguages)
-            {
-                string title = titles.FirstOrDefault(s => s.Language == lang)?.Title;
-                if (title != null) return title;
-            }
-            return $"Episode {a.EpisodeNumber}";
-        }
 
         public static AniDB GetAniDBInfo(AniDB_Episode ep)
         {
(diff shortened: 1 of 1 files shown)


Shoko.Server/API/v3/Models/Shoko/Episode.cs
```diff
@@ -51,10 +51,25 @@ public Episode(HttpContext ctx, SVR_AnimeEpisode ep)
...
+            Name = GetEpisodeName(anidb);
 
             Size = ep.GetVideoLocals().Count;
         }
 
+        internal static string[] TitleLanguages = new string[] { "EN", "X-JAT", "X-ZHT", "X-OTHER" };
+
+        public static string GetEpisodeName(AniDB_Episode a)
+        {
+            if (a == null) return "Episode 0";
```
Cazzar, line 63: We shouldn't be assuming a name if we can't get it.

Shoko.Server/API/v3/Models/Shoko/Episode.cs
```diff
@@ -51,10 +51,25 @@ public Episode(HttpContext ctx, SVR_AnimeEpisode ep)
...
+            var titles = RepoFactory.AniDB_Episode_Title.GetByEpisodeID(a.EpisodeID);
+            if (titles != null) foreach (string lang in TitleLanguages)
```
Cazzar, line 65: What about user preferences?

Shoko.Server/API/v3/Models/Shoko/Episode.cs
```diff
@@ -51,10 +51,25 @@ public Episode(HttpContext ctx, SVR_AnimeEpisode ep)
...
+            {
+                string title = titles.FirstOrDefault(s => s.Language == lang)?.Title;
+                if (title != null) return title;
+            }
+            return $"Episode {a.EpisodeNumber}";
```
Cazzar, line 70: Again, we shouldn't be forcing a title if we don't know it, this should be up to the client to fall back

da3dsoul:
Ok, I have multiple problems with this. I didn't comment on the file like cazzar, as these are at the design phase.

1. It does not match the series. Use `PreferredTitle` to store this.
2. It does not follow user preferences.
3. Episode Number fallback should be on the client, as cazzar said. It's not hard to do `"Episode " + PadNumber(ep.Number, totalEpisodes)`

In general, don't use Desktop as an example of proper handling of anything. Desktop is such a mess that we decided on a complete rewrite, and we don't make that decision lightly.

revam:
Did some more digging, and found out that `SVR_AnimeEpisode` already have a getter (`Title`) to get a title based on user preference, with a fallback to the English title if the preferred language is not available. I changed the implementation to get the name from `SVR_AnimeEpisode.Title`.

I need to do some more research before deciding to reinvent the wheel. 😑

PR #850, opened 2020-10-06
Add ResumePosition and Watched to File in v3 api

revam:
Also added a new endpoint "Scrobble" to update the either of the two values.


da3dsoul:
Looks good.

Issue #840, opened 2020-09-09
Collection import error - UNIQUE constraint failed: AniDB_File


jmediamanager:
﻿## VERSION INFORMATION

Shoko Server Version: v4.0.1.0
Operating System: Windows 8 / 2012 Server 64-bit
MediaInfo 20.3.0.0 (C:\Program Files (x86)\Shoko\Shoko Server\MediaInfo\MediaInfo.exe)
RHash 0.0.0.0 (C:\Program Files (x86)\Shoko\Shoko Server\x64\librhash.dll)

## LOG FILE

[2020-09-09.log](https://github.com/ShokoAnime/ShokoServer/files/5191930/2020-09-09.log)
[2020-09-09.01.log](https://github.com/ShokoAnime/ShokoServer/files/5191948/2020-09-09.01.log)

## DESCRIPTION

Error 1

```
[2020-09-09 08:24:14:589] Error|BackgroundWorker.WorkerThreadStart => CommandProcessorGeneral.WorkerCommands_DoWork => CommandRequest_ProcessFile.ProcessCommand Error processing CommandRequest_ProcessFile: 5695 - NHibernate.Exceptions.GenericADOException: could not insert: [Shoko.Server.Models.SVR_AniDB_File][SQL: INSERT INTO AniDB_File (Anime_GroupName, Anime_GroupNameShort, AnimeID, CRC, DateTimeUpdated, Episode_Rating, Episode_Votes, File_AudioCodec, File_Description, File_FileExtension, File_LengthSeconds, File_ReleaseDate, File_Source, File_VideoCodec, File_VideoResolution, FileID, FileName, FileSize, FileVersion, IsCensored, IsDeprecated, IsChaptered, InternalVersion, GroupID, Hash, IsWatched, MD5, SHA1, WatchedDate) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?); select last_insert_rowid()]
 ---> code = Constraint (19), message = System.Data.SQLite.SQLiteException (0x800027AF): constraint failed
UNIQUE constraint failed: AniDB_File.FileID
```

Error 2
```
[...]

PR #851, opened 2020-10-07
#840 - check if aniFile is null first

jmediamanager:
#840 Handle the situation where the file exists in store, but the method is forced to get from AniDB


da3dsoul:
Tiny thing, put a space before the ??=

da3dsoul:
Otherwise, I'm not sure how that happened. I probably missed it when cleaning up somewhere. Thanks

da3dsoul:
Okay yeah. I see when it happened. That is the correct fix. Good catch

jmediamanager:
> Tiny thing, put a space before the ??=

done

PR #857, opened 2020-11-20
Net5

swtrse:
A first overhaul of the Shoko.Cli including some cleanup of old entries in the project file that are not used anymore.


swtrse:
> 
> 
> Why are the imports inside the namespace, this is not the current code pattern we use.

It is common that imports are defined inside the deepest block they are used.
Originally imports where not designed to be used globally. Since most people used them globally this is shifted a bit.
However this is in most cases only a preference. There are only very view cases where this makes an actual difference.
If the current code pattern is otherwise I can change that back.

Regarding the code pattern (there is no guideline at the moment). We should create one so that newbies like me can look it up.

swtrse:
> 
> 
> What does this give us?
> We use resharper/rider currently, you don't explicitly have to "Enable support" for this.

I meant that I added the ReSharper Annotations for better ReSharper support. This Attributes are only there in debug and are removed in Release builds.

swtrse:
> 
> 
> A lot of this comment is redundant.

That is true. But that's exactly the thing. I did look at that code the first time. And I made this comments to clarify things. This makes it easier for others to step in.
Also the /// comments are of course redundant. The real benefit will be during design time when VS will show up exactly that description if you hover over the method or class or during the autocomplete, that's why I'm using this, and when you create an MSDN like documentation with Sandcastle.

swtrse:
> Multi-language exceptions aren't going to be of any value to us
> 99.9% of the time, the person is going to come to discord asking us what to do.
> And in these cases most of the dev team either speak English primarily, or as their only language, so whilst i18n can use useful, it slows us down massively.
> For cases when I am having to deal with default multiple translations, I have to deal with, poor machine translations to guess what the relevant exception is.

Putting output strings into resource have nothing to do with Multi-language exceptions. It can be introduced later, yes, but it has other benefits to.

Cazzar:
> Putting output strings into resource have nothing to do with Multi-language exceptions. It can be introduced later, yes, but it has other benefits to.
You're going to have to be a bit more explicit, as managing ResX files are honestly just a pain, so there is a cost-benefit that would need to be done.

> I meant that I added the ReSharper Annotations for better ReSharper support. This Attributes are only there in debug and are removed in Release builds.
Again, what does this give us? I've previously looked into these annotations, and never seen a huge value to them, the lining and static analysis done in resharper and vs these days manages to handle that, especially with Roslyn now existing.

maxpiva:
First, thanks :). 

Second I almost agree to everything @Cazzar said which is strange, because almost all the time we're not in sync ;)

I do see, that you want to create order in the project which is great for us since every 6 months we talk about doing a clean slate ;). But we always come short.

Anyway, welcome aboard. Please be sure to jump into the dev-chat where you have the privilege to hear our rants. 

Since you did an initialization in there, we should discuss the possibility of instantiating Shoko as a service on windows or daemon on Linux/Mac. We have managed old code for years, and we welcome any refactor that introduces DI to more parts of Shoko. We have also a long laundry list. So jump in, and make Shoko yours too.

Cazzar:
>Since you did an initialization in there, we should discuss the possibility of instantiating Shoko as a service on windows or daemon on Linux/Mac. We have managed old code for years, and we welcome any refactor that introduces DI to more parts of Shoko. We have also a long laundry list. So jump in, and make Shoko yours too.

you can't do this @maxpiva with Docker in play.
Without manually exposing the Docker daemon socket to the client... which is inherently insecure, and dangerous.

maxpiva:
> 
> you can't do this @maxpiva with Docker in play.
> Without manually exposing the Docker daemon socket to the client... which is inherently insecure, and dangerous.

Docker should stay docker. no changes in there :)

The approach should be instancing Shoko as a service for desktops. Because it should run with or without actually a user logged in.

Cazzar:
> Docker should stay docker. no changes in there :)
> 
> The approach should be instancing Shoko as a service for desktops. Because it should run with or without actually a user logged in.

This is outside of the PR, but Windows already has the service and trayicon, and officially we only support docker on Linux, as that's been a decision mostly for dependency reasons

maxpiva:
> This is outside of the PR, but Windows already has the service and trayicon, and officially we only support docker on Linux, as that's been a decision mostly for dependency reasons

Of course, is outside the PR, Just throwing bait in there. :). 

Windows Service. The issue is running shoko as a windows service, without an user logged.

hidden4003:
> Windows Service. The issue is running shoko as a windows service, without an user logged.

Windows service without user logged will have major impact on people using networked drives in windows as such service has no access to user credentials or user mapped drives.

PR #858, opened 2020-11-23
Net5

swtrse:
I did rewind Cli back to CLI and usings are outside of namespaces.
Also did start to setup workflows again but not finished yet.


.github/workflows/Shoko-CLI.osx64.yml
```diff
@@ -0,0 +1,41 @@
...
+      - run: echo Build Shoko.CLI v$SHOKO_MAJOR.$SHOKO_MINOR.$SHOKO_PATCH.$SHOKO_BUILD
+      - name: Unlock secrets
+        uses: sliteteam/github-action-git-crypt-unlock@1.1.0
+        env:
+          GIT_CRYPT_KEY: ${{ secrets.GIT_CRYPT_KEY }}
+      - name: Setup dotnet
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: ${{ matrix.dotnet }}
+      - run: dotnet publish Shoko.CLI -c Release -r osx-x64 -p:PublishTrimmed=true -p:TrimMode=Link --self-contained true
```
Cazzar, line 37: Why not use a matrix instead of creating 3 different files https://github.com/ShokoAnime/ShokoServer/blob/master/.github/workflows/cli.yml#L13-L16

Cazzar:
Might be worth re-evaluating this, as of current we have moved to .net 5.

While the logger changes could be done, I think that might be better to do progressively.

.github/workflows/Shoko-CLI.osx64.yml
revam: I assume because of the badges displayed in the readme (one for each platform)

da3dsoul:
I'm going to close this. We are now on net5. We can evaluate the other changes if there are any separately. This is just stale

PR #864, opened 2021-01-07
Add provider-agnostic series and episode types in the v3 api

revam:


Shoko.Server/API/v3/Models/Shoko/Episode.cs
```diff
@@ -251,4 +304,117 @@ public class EpisodeIDs : IDs
             #endregion
         }
     }
+
+    public interface IEpisodeMetadata
```
Cazzar, line 308: Personally, I think this should be split out into its own file.
Multiple classes/interfaces in one class are messy to me and can get confusing.

Shoko.Server/API/v3/Models/Shoko/Series.cs
```diff
@@ -451,14 +495,19 @@ public class AniDB
...
             /// <summary>
             /// TvDB ID
             /// </summary>
             [Required]
             public int ID { get; set; }
 
+            [JsonConverter(typeof(StringEnumConverter))]
+            public SeriesType Type { get { return SeasonNumber == 0 ? SeriesType.TVSpecial : SeriesType.TVSeries; } }
+
+            public bool Restricted { get { return false; } }
```
Cazzar, line 509: Probably best to add docs to the added params

Shoko.Server/API/v3/Models/Shoko/Series.cs
```diff
@@ -593,4 +636,97 @@ public SeriesSearchResult(HttpContext ctx, SVR_AnimeSeries ser, string match, do
             Distance = dist;
         }
     }
+
+    public interface ISeriesMetadata {
```
Cazzar, line 640: Again with the multiple interfaces/classes per file.

Shoko.Server/API/Startup.cs
```diff
@@ -91,6 +91,8 @@ public void ConfigureServices(IServiceCollection services)
                         var xml = Path.Combine(Path.GetDirectoryName(location), $"{Path.GetFileNameWithoutExtension(location)}.xml");
                         if (File.Exists(xml)) options.IncludeXmlComments(xml); //Include the XML comments if it exists.
                     }
+                    options.MapType<v3.Models.Shoko.SeriesType>(() => new Schema { Type = "string" });
```
da3dsoul, line 94: https://www.newtonsoft.com/json/help/html/JsonConverterAttributeProperty.htm

This mapping stuff is way unnecessary, unless I'm wrong, but I don't think I am

Shoko.Server/API/v3/Models/Shoko/Episode.cs
```diff
@@ -70,9 +93,9 @@ public static AniDB GetAniDBInfo(AniDB_Episode ep)
             var titles = RepoFactory.AniDB_Episode_Title.GetByEpisodeID(ep.EpisodeID);
             return new AniDB
             {
-                ID = ep.EpisodeID,
-                Type = (EpisodeType) ep.EpisodeType,
-                EpisodeNumber = ep.EpisodeNumber,
+                ID = ep.EpisodeID.ToString(),
```
da3dsoul, line 96: Why are you toStringing a number ID?

Shoko.Server/API/v3/Models/Shoko/Episode.cs
```diff
@@ -192,7 +240,12 @@ public class TvDB
             /// <summary>
             /// Episode Title, in the language selected for TvDB. TvDB doesn't allow pulling more than one language at a time, so this isn't a list.
             /// </summary>
-            public string Title { get; set; }
+            public string PreferredTitle { get; set; }
```
da3dsoul, line 243: My comment in the doc xml says it all. PreferredTitle and Titles aren't relevant to TvDB. We kept the model for providers separate because the data that a user of the API should expect should be clear by the schema. We don't want to always give a list and a PreferredTitle if there's only ever going to be one title.

Shoko.Server/API/v3/Models/Shoko/Episode.cs
da3dsoul, line 308: If you are splitting out an interface, put it in its own file. Probably, API/v3/Interfaces/Shoko/IEpisode.cs

But also, what's the point here? There won't ever be extended models based on these interfaces aside from the one that already exists.

Shoko.Server/API/v3/Models/Shoko/Series.cs
da3dsoul: TvDB doesn't allow Restricted content. That's why it's not listed. Also why I argue specific models per provider

Shoko.Server/API/v3/Models/Shoko/Episode.cs
revam: Removed the interfaces

Shoko.Server/API/v3/Models/Shoko/Series.cs
revam: Removed the interfaces, and reverted most of the cosmetic changes.

Shoko.Server/API/v3/Models/Shoko/Series.cs
revam: Removed the interfaces.

Shoko.Server/API/Startup.cs
revam: If you have a fix for the types showing up wrong in swagger, then please do tell me how to fix it, since that mapping was their recommended approach (from their project read-me).

Shoko.Server/API/v3/Models/Shoko/Episode.cs
revam: I'm `toString`ing a number, since not all (potential) providers _may_ be using integers as ids, and in Shokofin it's converted to a string and only used as a string value anyway.

Shoko.Server/API/v3/Models/Shoko/Episode.cs
revam: Removed the interfaces, and reverted most of the cosmetic changes.

Shoko.Server/API/v3/Models/Shoko/Episode.cs
revam: Removed the interfaces, and reverted most of the cosmetic changes.

revam:
Changed the PR to only do the minimum changes for adding episode/series types in APIv3.

da3dsoul:
I didn't see anything I'd complain about. We can discuss some of those other things further in a separate conversation

PR #867, opened 2021-01-14
Fix dependency injection cycle when loading plugins (Fix crash on startup)

Mik1ll:
Don't use settings from Service Provider since it is not built yet

### DIFF
--- a/Shoko.Server/Plugin/Loader.cs
+++ b/Shoko.Server/Plugin/Loader.cs
@@ -3,11 +3,13 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using NLog;
 using Shoko.Commons.Extensions;
 using Shoko.Plugin.Abstractions;
 using Shoko.Server.Settings;
+using Shoko.Server.Settings.Configuration;
 
 namespace Shoko.Server.Plugin
 {
@@ -18,7 +20,7 @@ namespace Shoko.Server.Plugin
         private readonly IList<Type> _pluginTypes = new List<Type>();
         private static readonly ILogger Logger = LogManager.GetCurrentClassLogger();
         
-        internal void Load(IServiceCollection serviceCollection)
+        internal void Load(IServiceCollection serviceCollection, IConfiguration config)
         {
             var assemblies = new List<Assembly>();
             var assembly = Assembly.GetExecutingAssembly();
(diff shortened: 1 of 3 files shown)


Cazzar:
I have a few other fixes in mind, one quicker solution would have been call the DI setup for Settings in `Startup.cs`

Shoko.Server/Plugin/Loader.cs
```diff
@@ -18,36 +20,42 @@ public class Loader : ISettingsProvider
...
-                    ServerSettings.Instance.Plugins.EnabledPlugins.TryAdd(name, true);
-                    if (!ServerSettings.Instance.Plugins.Priority.Contains(name))
-                        ServerSettings.Instance.Plugins.Priority.Add(name);
+                    if (!pluginSettings.EnabledPlugins.ContainsKey(name))
+                        pluginConfig.GetSection("EnabledPlugins")[name] = "true";
+                    if (!pluginSettings.Priority.Contains(name))
+                    {
+                        pluginSettings.Priority.Add(name);
+                        for (int i = 0; i < pluginSettings.Priority.Count; i++)
+                            pluginConfig.GetSection("Priority")[i.ToString()] = pluginSettings.Priority[i];
```
Cazzar, line 57: This won't work for IConfiguration, as it is readonly

Shoko.Server/Plugin/Loader.cs
```diff
@@ -18,36 +20,42 @@ public class Loader : ISettingsProvider
...
+            var userPlugins = Directory.Exists(userPluginDir) ? Directory.GetFiles(userPluginDir, "*.dll") : new string[0];
             
             foreach (var dll in Directory.GetFiles(dirname, "plugins/*.dll", SearchOption.AllDirectories).Concat(userPlugins))
             {
                 try
                 {
                     string name = Path.GetFileNameWithoutExtension(dll);
-                    if (ServerSettings.Instance.Plugins.EnabledPlugins.ContainsKey(name) &&
-                        !ServerSettings.Instance.Plugins.EnabledPlugins[name])
+                    if (pluginSettings.EnabledPlugins.ContainsKey(name) && pluginSettings.EnabledPlugins[name] == false)
```
Cazzar, line 43: Ultimately, the discovery will have to be moved outside of the enabled state so you follow the current flow:

Discover DLLs.
Do the container injection for plugins.
Finish DI container
Instantiate based on enabled settings.
Inject in the settings

Shoko.Server/Plugin/Loader.cs
Mik1ll: I've confirmed it's working on my machine. 🤷

Shoko.Server/Plugin/Loader.cs
Mik1ll: That would probably be the best flow, since editing settings before the service provider is built is probably unconventional.

Shoko.Server/Plugin/Loader.cs
Cazzar: https://github.com/aspnet/Configuration/issues/385

Cazzar: It may work on your machine, but it won't persist

PR #869, opened 2021-01-20
Get all plugin assembly types possible

Mik1ll:
Needed for my plugin to use embedded assembly types

### DIFF
--- a/Shoko.Server/Renamer/RenameFileHelper.cs
+++ b/Shoko.Server/Renamer/RenameFileHelper.cs
@@ -101,9 +101,9 @@ namespace Shoko.Server
                     {
                         return a.GetTypes();
                     } 
-                    catch
+                    catch (ReflectionTypeLoadException e)
                     {
-                        return new Type[0];
+                        return e.Types.Where(t => t != null);
                     }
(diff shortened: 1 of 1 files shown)


Cazzar:
My first look at this and I question what would specifically be why.
Are you able to give information about what this gives a bit more than what you have in the description?
And even possibly an example, since you have made the change simply to renamer loading, this is something that could affect plugin loading too.

Mik1ll:
Sure, within my renamer I reference an external assembly from a Nuget package. When I build my dll normally, the external assembly and my assembly are not packed together into one dll. So if Shoko Server calls getTypes on my assembly, there will be a bunch of types from my external assembly references that Shoko has no way of knowing about. Causing this exception. In order to avoid needing the external assembly to be loaded before my plugin is loaded, I decided to add the external assembly as an embedded resource and load it in the static constructor for my class. https://github.com/Mik1ll/ScriptRenamer/blob/414936a863d6e774a14b3f194f0a07a660d4ce96/ScriptRenamer/ScriptRenamer.cs#L58-L72
This I what I could think up as a solution, I don't know if there is any better way or I'm missing something.

Mik1ll:
Including the 3rd party dll seems to work alright anyway. Probably no real point in changing it after all.

PR #873, opened 2021-02-04
Fix external subtitle renaming

Mik1ll:
Would not match extensions because of dot and MediaInfo does not recognize Text streams in subtitle files with utf8bom or utf16be encodings.


Mik1ll:
Removed all MediaInfo meddling, just fix renaming the external subs.

PR #875, opened 2021-02-07
Fix script names not being url decoded.

Mik1ll:


Cazzar:
Given that Desktop, and APIv1 is effectively in maintenance and in the processes of being depreciated, I don't see this a hugely pressing issue.

PR #876, opened 2021-02-09
Update ShokoServer.cs

Baine:
```
0:27] Cazzar: The error itself is, this line of code is sync: https://github.com/ShokoAnime/ShokoServer/blob/master/Shoko.Server/API/v2/Models/core/APIMessage.cs#L66

We need to just allow sync io
[00:28] Cazzar: https://github.com/dotnet/aspnetcore/issues/8302#issuecomment-482187477 is the solution
```
hope i've plugged it into the correct place
at least build/throws no execption
@Cazzar please review :)

### DIFF
--- a/Shoko.Server/Server/ShokoServer.cs
+++ b/Shoko.Server/Server/ShokoServer.cs
@@ -52,7 +52,6 @@ using Trinet.Core.IO.Ntfs;
 using Action = System.Action;
 using LogLevel = NLog.LogLevel;
 using Timer = System.Timers.Timer;
-using Microsoft.AspNetCore.Server.Kestrel.Core;
 
 namespace Shoko.Server.Server
 {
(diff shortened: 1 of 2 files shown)


Shoko.Server/Server/ShokoServer.cs
```diff
@@ -125,6 +126,10 @@ internal static void ConfigureServices(IServiceCollection services)
                 loggingBuilder.AddNLog(new ConfigurationBuilder()
                     .AddJsonFile("appsettings.json", optional: true, reloadOnChange: true).Build());
             });
+	    services.Configure<KestrelServerOptions>(options =>
```
Cazzar, line 129: A: Indentation
B: this should be in Startup.cs

Baine: fixed in 2nd commit

PR #881, opened 2021-05-19
Paginate GetAllSeries endpoint

harshithmohan:


Shoko.Server/API/v3/Controllers/SeriesController.cs
```diff
@@ -25,13 +25,17 @@ public class SeriesController : BaseController
     {
         /// <summary>
         /// Get a list of all series available to the current user
+        /// Use pageSize and page (index 0) in the query to enable pagination.
         /// </summary>
         /// <returns></returns>
         [HttpGet]
-        public ActionResult<List<Series>> GetAllSeries()
+        public ActionResult<List<Series>> GetAllSeries(int pageSize = 100, int page = 0)
```
Cazzar, line 32: Any reason (except for easier math below) to start the page at 0? Naturally, pagination is normally /?page=1....

harshithmohan: Nope. As I said, stolen code from the unrecognized files endpoint. Should I change it?

da3dsoul: It seems like limit and offset are the standard for that. I don't really care, but we should pick one and change all of them to it

Cazzar: My personal preference to be 1 onwards, but I'll leave the choice to you @da3dsoul

da3dsoul: I was referring to api/v3/series?limit=20&offset=40
It doesn't do the page math for you.
Like I said, I don't care what we do there, as long as we decide and stick to it

PR #893, opened 2021-09-13
Fix "original work" tag for tag filer

revam:
and modernise _some_ of the code.

It works as expected on my-machine™ from the v3 api.

Haven't tested the the changes for the series search and the v2 filters, but the code looks fine imo.

### DIFF
--- a/Shoko.Server/API/v2/Models/common/Filters.cs
+++ b/Shoko.Server/API/v2/Models/common/Filters.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.Serialization;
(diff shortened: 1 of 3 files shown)


Shoko.Server/API/v2/Models/common/Filters.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
```
revam, line 1: Unintended.

Shoko.Server/Utilities/TagFilter.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
```
revam, line 1: Unintended.

revam:
Or so i thought.

Cazzar:
This would need to be heavily tested in v2

I would also ask why are you changing the name and signature of functions?

revam:
The signatures changed because the parameters was not needed in the updated function. The names i can revert if you don't like it.

revam:
Reason; probably needs more testing.

revam:
I modified it to queue the "original work" tag for removal, because we keep it and discard the "new" tag if the "new" tag is not queued for removal.

This time I tested with multiple allow/deny filters and with multiple series (anidb ids: 9556 (anime original), 14477 (novel) and 6246 (manga) ), though i couldn't find a series that's missing a "source"-type tag in my collection to test the edge case where no source tags are present.

revam:
> This would need to be heavily tested in v2
> 
> I would also ask why are you changing the name and signature of functions?

I changed the signatures since i noticed that the two last arguments were never used in the current code-base. The `bool addTags` was always set to true (the default) so it always added the new/renamed tags and the `bool invert` (which was set to false by default) was only used if the tag-list contained one element and we already had an invert flag in the filter so i didn't see the need for it.

da3dsoul:
I think I want to try something. This is also a good class to write unit tests for, since it's pretty well generified already. I'll pull this down after work and mess with it

da3dsoul:
Put into the tagfilter branch. I'll open another PR for that one

PR #895, opened 2021-09-14
Fix: Take `number` items instead of a literal `10`

revam:


Shoko.Server/API/v3/Controllers/DashboardController.cs
```diff
@@ -105,16 +105,17 @@ private static bool MissingBothTvDBAndMovieDBLink(SVR_AnimeSeries ser)
         [HttpGet("TopTags")]
         public List<Tag> GetTopTags()
         {
-            return GetTopTags(10);
+            return GetTopTags();
```
da3dsoul, line 108: This will cause infinite recursion. The 10 is passing a default value to the other endpoint. I didn't know how to do optional routes, so this is how I did it at the time.

Shoko.Server/API/v3/Controllers/DashboardController.cs
```diff
@@ -105,16 +105,17 @@ private static bool MissingBothTvDBAndMovieDBLink(SVR_AnimeSeries ser)
...
 
         /// <summary>
-        /// Gets the top <para>number</para> most common tags visible to the current user 
+        /// Gets the top <para>number</para> of the most common tags visible to the current user.
         /// </summary>
-        /// <param name="number"></param>
+        /// <param name="number">The max number of results to return</param>
+        /// <param name="filter">The <see cref="TagFilter.Filter" /> to use. (Defaults to <see cref="TagFilter.Filter.AnidbInternal" /> | <see cref="TagFilter.Filter.Misc" /> | <see cref="TagFilter.Filter.Source" />)</param>
         /// <returns></returns>
         [HttpGet("TopTags/{number}")]
```
da3dsoul, line 117: I would need to look it up, but I believe you can make it {number?} to make it optional, which can let you remove the above method entirely.

da3dsoul (changes requested):

Shoko.Server/API/v3/Controllers/DashboardController.cs
revam: Yeah, apparently that works.

PR #896, opened 2021-09-14
Fix "original work" for tag filer

da3dsoul:
Continuation of https://github.com/ShokoAnime/ShokoServer/pull/893

### DIFF
--- a/Shoko.Server/Utilities/TagFilter.cs
+++ b/Shoko.Server/Utilities/TagFilter.cs
@@ -1,6 +1,8 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Shoko.Commons.Extensions;
+
 // ReSharper disable StringLiteralTypo
 // ReSharper disable StaticMemberInGenericType
 // ReSharper disable IdentifierTypo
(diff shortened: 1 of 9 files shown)


Shoko.Server/Utilities/TagFilter.cs
```diff
@@ -1,597 +1,607 @@
...
+                        toRemove.Add(sourceTag);
+                    }
+
+                    lock (toAdd)
+                    {
+                        toAdd.Add(newTag);
+                    }
+                }
+
+                if (_tagReplacementDictionary.TryGetValue(tag, out newTag))
```
revam, line 597: Why is the dictionary named "\_tag**Replacement**Dictionary" if it is a dictionary of tags to remove if some other tags are present?

revam:
Besides the remark above then it LGTM.

Shoko.Server/Utilities/TagFilter.cs
da3dsoul: Because I couldn't think of a good name. Meh

revam:
Added a few more specific tests, otherwise it LGTM.

PR #898, opened 2021-09-18
Add staff data (Continued from #897)

revam:


revam:
> It's not important, but just for clarity, I'd rename NullHashSet to InverseNullValueSet or something similar

Done.

Shoko.Server/Databases/SQLite.cs
```diff
@@ -548,7 +548,9 @@ public void CreateDatabase()
             new DatabaseCommand(78, 1, DropVideoLocal_Media),
             new DatabaseCommand(79, 1, "DROP INDEX IF EXISTS UIX_CrossRef_AniDB_MAL_MALID;"),
             new DatabaseCommand(79, 1, "DROP INDEX IF EXISTS UIX_CrossRef_AniDB_MAL_MALID;"),
-            new (80, 1, "DROP INDEX IF EXISTS UIX_AniDB_File_FileID;"),
+            new DatabaseCommand(80, 1, "DROP INDEX IF EXISTS UIX_AniDB_File_FileID;"),
+            new DatabaseCommand(81, 1, "CREATE TABLE AniDB_Anime_Staff ( AniDB_Anime_StaffID INTEGER PRIMARY KEY AUTOINCREMENT, AnimeID INTEGER NOT NULL, CreatorID INTEGER NOT NULL, CreatorType TEXT NOT NULL );"),
+            new DatabaseCommand(81, 2, DatabaseFixes.PopulateAniDBEpisodeDescriptions),
         };
```
da3dsoul, line 554: I would make the TEXT an `nvarchar(200)` the 200 is arbitrary and big enough.
Is the database fix just to trigger regen from XML?

revam: 1. SQLite doesn't actually support char and vchar, it maps the data to a TEXT field internally.
2. Not that i found.

revam: 1. Though I can still change it.

da3dsoul: Oh I was looking at the wrong one. I need to pull out my laptop for these from now on. The mobile view is causing too many issues

revam: Should the Database fix be renamed?

da3dsoul: Yeah, it's likely to be reused in the future, so renaming it would be a good idea. And don't change the TEXT. That was my mistake

revam: Also my mistake, the fix only runs on the cached data.

revam: It doesn't try to fetch new data, and will skip the series/entry if no cached data is found.

PR #901, opened 2021-09-22
Minor tweaks and fixes for the controllers in the v3 API

revam:
Removed all the white spaces and some unneeded new lines, updated the action controller to use an `ILogger<T>`, added the AniDB refresh endpoint to the series controller, renamed the series refresh endpoint for known series (it's still unused in clients, and only available in the latest dailies, so no clients were affected by this change), removed any direct dependency on the v2 API from within the v3 API, added missing (empty) parameter descriptions for some parameters, and probably some more that I'm forgetting.


revam:
NOTE 2: I need to figure out how to _not_ change the new lines.

harshithmohan:
> NOTE: this PR _should_ be squashed before merging.

You think?

revam:
Will redo it later.

PR #902, opened 2021-09-25
Add voting to api v3

revam:


revam:
It works on-my-machine™ :+1:

PR #905, opened 2021-09-27
Handle invalid regex pattern in `File/PathRegex` API

harshithmohan:
Not sure if it's the right way to do it. So made a PR for someone to check it.


da3dsoul:
Looks fine to me

PR #913, opened 2021-11-13
Allow Shoko to access and hash seeding files

Hitsounds:
Libtorrent uses FileShare.ReadWrite when opening files to seed (https://github.com/arvidn/libtorrent/blob/2ceadfb9b7378c0d8ddd9e501d7baa5f6324425a/src/file.cpp#L210). 

This can lead to issues on popular files where the file immediately begins seeding and Shoko fails to access the file to hash it. By testing access to the file using the ReadWrite flag and also opening it when hashing with the ReadWrite flag Shoko can access the file while libtorrent is also accessing it.

### DIFF
--- a/Shoko.Server/Commands/Import/CommandRequest_HashFile.cs
+++ b/Shoko.Server/Commands/Import/CommandRequest_HashFile.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


da3dsoul:
You sound like you know what you're talking about here. I'm fine with merging this, but also keep in mind we have several preliminary checks in addition to this to ensure that Shoko doesn't mess with files that are still downloading, copying, etc.

Hitsounds:
The checks in FileModified() should ensure that Shoko still doesn't mess with files being actively modified. It does mean that Shoko no longer acquires an exclusive lock on the file when opening it however I don't believe this will cause problems with the other checks in place.

https://docs.microsoft.com/en-us/dotnet/api/system.io.fileshare?view=net-5.0
image

da3dsoul:
Yep. I didn't think it would. I was just saying if you weren't aware of them, you might encounter behavior you didn't expect

Hitsounds:
Ah, thanks for the heads up. I've been running it on my machine and it has been behaving how I expected it to but I do keep downloading files separate from my drop source folder. Do you have any concerns?

da3dsoul:
I don't think so. I'll go ahead and merge it

PR #915, opened 2022-01-31
Update the file controller in the v3 api

revam:


Shoko.Server/Repositories/Cached/VideoLocalRepository.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
```
revam, line 1: I have no idea what I changed here, nor how to revert it.

da3dsoul (approved):

PR #920, opened 2022-02-13
Restore renamer custom errors

Mik1ll:
Restores ability of renamers to throw exceptions with custom error messages.
Maintains ability to ignore exception and defer to the next renamer with Plugins.DeferrOnError set to true.
If DeferOnError is set and a renamer sets the event args to cancel, then it will throw the exception anyway.
Additionally:
* Prevents errors when moving to same location
* Restores error on moving to location of different file

### DIFF
--- a/Shoko.Server/Renamer/RenameFileHelper.cs
+++ b/Shoko.Server/Renamer/RenameFileHelper.cs
@@ -78,11 +78,8 @@ namespace Shoko.Server
                 }
                 catch (Exception e)
                 {
-                    Logger.Error($"Renamer threw an error. The offending plugin was: {renamer.GetType().GetAssemblyName()} with renamer: {renamer.GetType().Name}. The error was: {e}");
-
-                    if (ServerSettings.Instance.Plugins.DeferOnError) continue;
-
-                    return null;
+                    if (ServerSettings.Instance.Plugins.DeferOnError && !args.Cancel) continue;
+                    throw;
                 }
             }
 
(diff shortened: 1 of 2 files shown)


revam:
Do we want to remove the log points? They tell us when it threw even if it was deferred.

Shoko.Server/Renamer/RenameFileHelper.cs
```diff
@@ -78,11 +78,8 @@ public static string GetFilename(SVR_VideoLocal_Place place, string scriptName)
                 }
                 catch (Exception e)
                 {
-                    Logger.Error($"Renamer threw an error. The offending plugin was: {renamer.GetType().GetAssemblyName()} with renamer: {renamer.GetType().Name}. The error was: {e}");
```
Cazzar, line 81: We shouldn't be losing the logging detail, so I would like something back, even if it was in the `continue` case

Shoko.Server/Renamer/RenameFileHelper.cs
```diff
@@ -131,11 +128,8 @@ public static (ImportFolder, string) GetDestination(SVR_VideoLocal_Place place,
...
```
Cazzar, line 134: We shouldn't be losing the logging detail, so I would like something back, even if it was in the `continue` case

Cazzar (changes requested):

Shoko.Server/Renamer/RenameFileHelper.cs
Mik1ll: Added logging back in [f8c288b](https://github.com/ShokoAnime/ShokoServer/pull/920/commits/f8c288be84a0928494e53235d9d3cde32c7f09b2)

Mik1ll (commented):
Simply adds double quotes to renaming/moving logs for consistency and convenience. Was mixed quotes/non-quotes before.

PR #921, opened 2022-02-16
Fix logging config for ILogger DI

Mik1ll:
I noticed while implementing logging for my renamer that DI'd renamers had their log level restricted to  >error. The minimum log level set by the preprocessor only restricted loggers which were DI'd. It was probably set to >error because of tons of log spam by Microsoft.* and System.Net.Http.*.

* Enabled Info logging for DI'd loggers.
* Debug build logging preprocessor directive moved to InitLoggers and uses Nlog.Configuration.
* Blackholed logging from Microsoft.*, System.Net.Http.*, and Shoko.Server.API.Authentication.CustomAuthHandler so they don't pollute logs.
* Removed the redundant Nlog service adding in ConfigureServices.


Mik1ll:
As a side note I've noticed that enabling trace logging for release build does not include debug logging. This means in order for a user to have both trace and debug, they need to go into nlog.config and change the file logger to debug or below. Also, setting trace logging in nlog.config will not enable it because this function will disable it when the trace log server setting is set to false. https://github.com/ShokoAnime/ShokoServer/blob/ff093c3c05f4d3be9b95fc05c1b5605017e1778c/Shoko.Server/Server/ShokoServer.cs#L165-L174

Shoko.Server/nlog.config
```diff
@@ -7,6 +8,13 @@
         <target name="console" xsi:type="ColoredConsole" layout="${date:format=HH\:mm\:ss}| --- ${message}" />
     </targets>
     <rules>
+        <!--Skip non-critical Microsoft logs and so log only own logs (Blackhole) -->
+        <logger name="Microsoft.*" maxlevel="Info" final="true" />
+        <logger name="System.Net.Http.*" maxlevel="Info" final="true" />
```
revam, line 13: Wouldn't this exclude the requests/responses from the log? (They're very useful for debugging)

revam: The "`System.Net.Http.*`" rule i mean

Mik1ll: Yes they are excluded, but they are already excluded in current build. The log level was set to error and above in the code, so none of the info logging got through. There shouldn't be any change and now you can enable that logging by changing the config line without rebuilding.

Mik1ll: I've re-enabled the http log spam in debug build as the objective of this pull request is not to change logging behavior. Also, turns out I didn't see http logs coming from System.Net.Http, it was from Microsoft.AspNetCore.

Cazzar:
> As a side note I've noticed that enabling trace logging for release build does not include debug logging

This code's intention was down to the fact that we wanted to have an easily configured option to enable trace logging, which AFAIK this should implicitly enable debug, given how most logging tools do a "is level above or below this threshold"

PR #922, opened 2022-02-17
Update APIv3 controllers and related.

revam:
Add the ability to automatically create the corresponding `AnimeSeries`
entry if it doesn't already exists for related series (behind a setting
that's off by default) under import of a file, or manually in the
v3 api.

### DIFF
--- a/Shoko.Server/API/v3/Controllers/ActionController.cs
+++ b/Shoko.Server/API/v3/Controllers/ActionController.cs
@@ -7,6 +7,7 @@ using Microsoft.AspNetCore.Mvc;
 using NLog;
 using Shoko.Server.API.Annotations;
 using Shoko.Server.API.v2.Models.core;
+using Shoko.Server.API.v3.Models.Shoko;
 using Shoko.Server.Commands;
 using Shoko.Server.Providers.MovieDB;
 using Shoko.Server.Providers.TraktTV;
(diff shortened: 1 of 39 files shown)


revam:
I'll fix the line endings tomorrow...
:sigh:

revam:
Finally fixed the line endings.

Shoko.Server/Models/SVR_AnimeSeries.cs
```diff
@@ -395,17 +395,17 @@ public void Populate(SVR_AniDB_Anime anime)
             SeriesNameOverride = string.Empty;
         }
 
-        public void CreateAnimeEpisodes()
+        public void CreateAnimeEpisodes(SVR_AniDB_Anime anime = null)
```
revam, line 398: Should I revert these changes? I didn't see why we needed to consult the database/cache to get the anime if we already had it present in the scope when the function was called.

Shoko.Server/API/v3/Controllers/ActionController.cs
```diff
@@ -6,7 +6,7 @@
 using Microsoft.AspNetCore.Mvc;
 using NLog;
 using Shoko.Server.API.Annotations;
-using Shoko.Server.API.v2.Models.core;
+using Shoko.Server.API.v3.Models.Shoko;
```
revam, line 9: I also removed the reference to the v2 models (used for the action results)

revam:
Time to fix the line endings... again

revam:
fixed again.

revam:
Again, will fix the line endings... tomorrow™

revam:
This weekend™

revam:
...soon:tm:

revam:
It is now tomorrow. Finally.

PR #923, opened 2022-02-21
Ensure consistant directory seperators from renamer plugins

Mik1ll:
Prevents mixed forward and backslash directory separators from renamer plugins on Windows.

### DIFF
--- a/Shoko.Server/Renamer/RenameFileHelper.cs
+++ b/Shoko.Server/Renamer/RenameFileHelper.cs
@@ -116,6 +116,7 @@ namespace Shoko.Server
                     if (args.Cancel) return (null, null);
                     // if no path was specified, then defer
                     if (string.IsNullOrEmpty(destPath) || destFolder == null) continue;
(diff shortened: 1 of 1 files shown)


Shoko.Server/Renamer/RenameFileHelper.cs
```diff
@@ -116,6 +116,7 @@ public static (ImportFolder, string) GetDestination(SVR_VideoLocal_Place place,
                     if (args.Cancel) return (null, null);
                     // if no path was specified, then defer
                     if (string.IsNullOrEmpty(destPath) || destFolder == null) continue;
+                    destPath = destPath.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar);
```
Cazzar, line 119: Problem with this solution, Linux lists `/` as both values, which makes it a pointless replace.;

```suggestion
                    if (Path.AltDirectorySeparatorChar != Path.DirectorySeparatorChar) {
                        destPath = destPath.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar);
                    }
```

PR #935, opened 2022-06-29
Add a SettingsSaved event to the ShokoEventHandler

revam:
for use in plugins (and in the api when the SignalR emitter/(hub is
added)

### DIFF
--- a/Shoko.Server/Server/ShokoEventHandler.cs
+++ b/Shoko.Server/Server/ShokoEventHandler.cs
@@ -99,12 +99,11 @@ namespace Shoko.Server
             });
         }
 
-        internal void OnSettingsSaved(string before, string after)
+        internal void OnSettingsSaved()
         {
             SettingsSaved?.Invoke(null, new SettingsSavedEventArgs
             {
-                Before = before,
-                After = after,
+                
             });
         }
     }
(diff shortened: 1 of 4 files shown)


Cazzar:
Personally not the biggest fan of exposing the settings in the Event context, as it does contain user sensitive details

revam:
So should we just remove the `Before` and `After` properties then?

Shoko.Server/Server/ShokoEventHandler.cs
```diff
@@ -97,5 +98,13 @@ internal void OnEpisodeUpdated(DataSourceEnum source, SVR_AniDB_Anime anime, SVR
                 EpisodeInfo = episode,
             });
         }
+
+        internal void OnSettingsSaved()
+        {
+            SettingsSaved?.Invoke(null, new SettingsSavedEventArgs
+            {
+                
+            });
```
Cazzar, line 107:
```suggestion
            SettingsSaved?.Invoke(null, new SettingsSavedEventArgs());
```

no point having the extra lines/object constructor if nothing is used, we can always add it in if needed.

Cazzar (changes requested):

PR #937, opened 2022-07-01
Add continue watching endpoint

revam:
Blocked by #936 and missing user data update timestamp.

### DIFF
--- a/Shoko.Server/API/v3/Controllers/DashboardController.cs
+++ b/Shoko.Server/API/v3/Controllers/DashboardController.cs
@@ -250,8 +250,8 @@ namespace Shoko.Server.API.v3.Controllers
         public List<Dashboard.EpisodeDetails> GetContinueWatchingEpisodes([FromQuery] [Range(0, 100)] int pageSize = 20, [FromQuery] [Range(0, Int16.MaxValue)] int page = 0)
         {
             var user = HttpContext.GetUser();
-            var episodeList = RepoFactory.VideoLocalUser.Cache.Values
-                .Where(userData => userData.JMMUserID == user.JMMUserID && userData.WatchedDate == null && userData.ResumePosition != 0)
+            var episodeList = RepoFactory.VideoLocalUser.GetByUserID(user.JMMUserID)
+                .Where(userData => userData.WatchedDate == null && userData.ResumePosition != 0)
                 .OrderByDescending(userData => userData.LastUpdated)
                 .Select(userData => RepoFactory.VideoLocal.GetByID(userData.VideoLocalID))
                 .Where(file => file != null)
--- a/Shoko.Server/API/v3/Controllers/DashboardController.cs
+++ b/Shoko.Server/API/v3/Controllers/DashboardController.cs
@@ -252,11 +252,11 @@ namespace Shoko.Server.API.v3.Controllers
             var user = HttpContext.GetUser();
             var episodeList = RepoFactory.VideoLocalUser.Cache.Values
                 .Where(userData => userData.JMMUserID == user.JMMUserID && userData.WatchedDate == null && userData.ResumePosition != 0)
-                // TODO: order by last updated when a last updated timestamp have been added to the user data.
-                .OrderByDescending(userData => userData.VideoLocalID)
+                .OrderByDescending(userData => userData.DateTimeUpdated)
                 .Select(userData => RepoFactory.VideoLocal.GetByID(userData.VideoLocalID))
                 .Where(file => file != null)
                 .Select(file => file.EpisodeCrossRefs.OrderBy(xref => xref.EpisodeOrder).ThenBy(xref => xref.Percentage).FirstOrDefault())
+                .Where(xref => xref != null)
                 .Select(xref => RepoFactory.AnimeEpisode.GetByAniDBEpisodeID(xref.EpisodeID))
                 .Where(episode => episode != null)
                 .DistinctBy(episode => episode.AnimeSeriesID)
--- a/Shoko.Server/API/v3/Controllers/DashboardController.cs
+++ b/Shoko.Server/API/v3/Controllers/DashboardController.cs
@@ -252,7 +252,7 @@ namespace Shoko.Server.API.v3.Controllers
             var user = HttpContext.GetUser();
             var episodeList = RepoFactory.VideoLocalUser.Cache.Values
                 .Where(userData => userData.JMMUserID == user.JMMUserID && userData.WatchedDate == null && userData.ResumePosition != 0)
-                .OrderByDescending(userData => userData.DateTimeUpdated)
+                .OrderByDescending(userData => userData.LastUpdated)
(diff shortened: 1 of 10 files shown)


bigretromike:
So this is not filter anymore? or its just doing the same as filter (continue watching) ?

bigretromike:
> We have a filter for that? Than we can probably use the filter. Will look into if they're the same later today or tomorrow.

yes, there is system one with all series you started to watched based on shoko watch flag.

bigretromike:
good find, definitely something for multi user

revam:
> > We have a filter for that? Than we can probably use the filter. Will look into if they're the same later today or tomorrow.
> 
> yes, there is system one with all series you started to watched based on shoko watch flag.

It doesn't list the most recently watched episode for each series (with some additional series info) in the most-recently-watched order afaik

revam:
https://github.com/ShokoAnime/Shoko.Models/pull/5 and https://github.com/ShokoAnime/Shoko.Commons/pull/6 will have to be merged first.

Also, I still haven't tested the MySQL or MS SQL Server database upgrades. The SQLite upgrade worked fine.

Shoko.Server/API/v3/Controllers/DashboardController.cs
```diff
@@ -240,6 +240,46 @@ public List<Series> GetRecentlyAddedSeries([FromQuery] int pageSize = 20, [FromQ
...
+                .Select(episode => episode.GetAnimeSeries())
+                .Where(series => series != null)
+                .ToDictionary(series => series.AnimeSeriesID);
+
+            if (pageSize <= 0)
+                return episodeList
+                    .Where(episode => user.AllowedSeries(seriesDict[episode.AnimeSeriesID]))
+                    .Select(a => new Dashboard.EpisodeDetails(a.AniDB_Episode, seriesDict[a.AnimeSeriesID].GetAnime(), seriesDict[a.AnimeSeriesID]))
+                    .ToList();
+            if (page <= 0) page = 0;
```
Cazzar, line 274: This should be part of a validation rule

Shoko.Server/Databases/MySQL.cs
```diff
@@ -633,7 +633,11 @@ public class MySQL : BaseDatabase<MySqlConnection>, IDatabase
             new DatabaseCommand(91, 1, DropMALIndex),
             new DatabaseCommand(92, 1, DropAniDBUniqueIndex),
             new DatabaseCommand(93, 1, "CREATE TABLE `AniDB_Anime_Staff` ( `AniDB_Anime_StaffID` INT NOT NULL AUTO_INCREMENT, `AnimeID` int NOT NULL, `CreatorID` int NOT NULL, `CreatorType` varchar(50) NOT NULL, PRIMARY KEY (`AniDB_Anime_StaffID`) );"),
-            new DatabaseCommand(93, 2, DatabaseFixes.RefreshAniDBInfoFromXML),
+            new DatabaseCommand(93, 2, DatabaseFixes.RefreshAniDBInfoFromXML),
+            new DatabaseCommand(94, 1, "UPDATE VideoLocal_User SET WatchedDate = NULL WHERE WatchedDate = '1970-01-01 00:00:00';"),
+            new DatabaseCommand(94, 2, "ALTER TABLE VideoLocal_User ADD WatchedCount INT NOT NULL DEFAULT 0;"),
+            new DatabaseCommand(94, 3, "ALTER TABLE VideoLocal_User ADD DateTimeUpdated datetime NOT NULL DEFAULT '2000-01-01 00:00:00';"),
```
Cazzar, line 639:
```suggestion
            new DatabaseCommand(94, 3, "ALTER TABLE VideoLocal_User ADD DateTimeUpdated datetime NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP"),
```

Reason for this, it will automatically keep the updated timestamp regardless of user set.

Shoko.Server/Databases/SQLServer.cs
```diff
@@ -611,7 +611,11 @@ public void CreateDatabase()
             new DatabaseCommand(84, 1, "DROP INDEX IF EXISTS UIX_CrossRef_AniDB_MAL_MALID ON CrossRef_AniDB_MAL;"),
             new DatabaseCommand(85, 1, "DROP INDEX IF EXISTS UIX_AniDB_File_FileID ON AniDB_File;"),
             new DatabaseCommand(86, 1, "CREATE TABLE AniDB_Anime_Staff ( AniDB_Anime_StaffID int IDENTITY(1,1) NOT NULL, AnimeID int NOT NULL, CreatorID int NOT NULL, CreatorType varchar(50) NOT NULL );"),
-            new DatabaseCommand(86, 2, DatabaseFixes.RefreshAniDBInfoFromXML),
+            new DatabaseCommand(86, 2, DatabaseFixes.RefreshAniDBInfoFromXML),
+            new DatabaseCommand(87, 1, "UPDATE VideoLocal_User SET WatchedDate = NULL WHERE WatchedDate = '1970-01-01 00:00:00';"),
+            new DatabaseCommand(87, 2, "ALTER TABLE VideoLocal_User ADD WatchedCount INT NOT NULL DEFAULT 0;"),
+            new DatabaseCommand(87, 3, "ALTER TABLE VideoLocal_User ADD DateTimeUpdated datetime NOT NULL DEFAULT '2000-01-01 00:00:00';"),
```
Cazzar, line 617:
```suggestion
            new DatabaseCommand(87, 3, "ALTER TABLE VideoLocal_User ADD DateTimeUpdated datetime NOT NULL DEFAULT CURRENT_TIMESTAMP;"),
```

Shoko.Server/Mappings/VideoLocal_UserMap.cs
```diff
@@ -13,7 +13,9 @@ public VideoLocal_UserMap()
             Map(x => x.JMMUserID).Not.Nullable();
             Map(x => x.VideoLocalID).Not.Nullable();
             Map(x => x.WatchedDate);
+            Map(x => x.WatchedCount).Not.Nullable();
             Map(x => x.ResumePosition).Not.Nullable();
+            Map(x => x.DateTimeUpdated).Not.Nullable();
```
Cazzar, line 18: Suggestion: rename this to `LastUpdated` as that is typically the naming format for these columns

Cazzar (changes requested):

Shoko.Server/API/v3/Controllers/DashboardController.cs
revam: I didn't know the range validation rule existed until earlier today. Will convert this now, then the other endpoints using the same pattern later today (in another PR).

Shoko.Server/Mappings/VideoLocal_UserMap.cs
revam: I just stole the name from what was already used in the other models/tables. It can still be renamed.

PR #940, opened 2022-07-03
Add the "Recommended For You" anime recommendation algorithm

revam:
to be used on the dashboard.

### DIFF
--- a/Shoko.Server/API/v3/Controllers/SeriesController.cs
+++ b/Shoko.Server/API/v3/Controllers/SeriesController.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
 using System.IO;
 using System.Linq;
 using System.Web;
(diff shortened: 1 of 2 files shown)


revam:
Forgot to say it depends on https://github.com/ShokoAnime/ShokoServer/pull/937 getting merged first.

revam:
Need to change approval to be percent based instead of absolute values

Shoko.Server/API/v3/Controllers/SeriesController.cs
```diff
@@ -182,6 +182,136 @@ public ActionResult<List<SeriesRelation>> GetAnidbRelationsBySeriesID([FromRoute
...
+                endDate = DateTime.Now;
+            if (endDate.HasValue && !startDate.HasValue)
+                return BadRequest("Missing start date.");
+            if (startDate.HasValue && endDate.HasValue)
+            {
+                if (endDate.Value > DateTime.Now)
+                    return BadRequest("End date cannot be set into the future.");
+                if (startDate.Value > endDate.Value)
+                    return BadRequest("Start date cannot be newer than the end date.");
+            }
```
Cazzar, line 221: Are we able to do this validation via attributes, to avoid boilerplate code?

Cazzar (changes requested):

Shoko.Server/API/v3/Controllers/SeriesController.cs
revam: Will check.

revam: It seems like the `Range` attribute only supports static values, and there is no specific attribute for validating `DateTime` ranges, so I will prefer this boilerplate over the other solution of writing a custom attribute that would only be used once (resulting in even more boilerplate). And I still don't know if it loads the custom attribute into memory once the app starts (resulting in semi-static values) or if it checks the attribute every time the method is accessed.

Unless you (or any other members/users) know a better way then I will leave it there.

PR #947, opened 2022-07-22
feature: add `ListResult` for list endpoints in v3 api

revam:
I didn't touch the paginated dashboard endpoints.

### DIFF
--- a/Shoko.Server/API/v3/Controllers/EpisodeController.cs
+++ b/Shoko.Server/API/v3/Controllers/EpisodeController.cs
@@ -1,10 +1,12 @@
 using System;
 using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Shoko.Commons.Extensions;
 using Shoko.Server.API.Annotations;
+using Shoko.Server.API.v3.Helpers;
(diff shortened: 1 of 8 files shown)


Shoko.Server/API/v3/Models/Common/TotalList.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.ComponentModel.DataAnnotations;
+
+#nullable enable
+namespace Shoko.Server.API.v3.Models.Common
+{
+    /// <summary>
+    /// A list with a total count and a sliced page of the total or the total
+    /// list.
+    /// </summary>
+    public class TotalList<T>
```
da3dsoul, line 13: It's a nitpick, but I would name it something like ListResult, to match more with the ActionResult naming scheme.

Shoko.Server/API/v3/Models/Common/TotalList.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        /// <summary>
+        /// Total number of entries that matched the search.
+        /// </summary>
+        public int Total { get; set; }
+
+        /// <summary>
+        /// A sliced page or the whole list of <typeparamref name="T"/> entries.
+        /// <value></value>
+        public List<T> List { get; set; }
```
da3dsoul, line 43: I would make this an IReadOnlyList<T>, just to make it clear

PR #949, opened 2022-07-27
feature: add web ui extra

revam:


revam:
The property name and query parameter name is subject to change.

Shoko.Server/API/Startup.cs
```diff
@@ -169,7 +169,7 @@ public void ConfigureServices(IServiceCollection services)
                 .AddNewtonsoftJson(json =>
                 {
                     json.SerializerSettings.MaxDepth = 10;
-                    json.SerializerSettings.ContractResolver = new OmitEmptyEnumerableResolver
+                    json.SerializerSettings.ContractResolver = new DefaultContractResolver
```
revam, line 172: This is for "fixing" the omitted lists that break the web ui.

Cazzar:
This whole PR is kinda screaming "hacky solution to a problem" what are we adding here with this "web ui extra" looking at both the code and the commit information really doesn't explain anything.

hidden4003:
For group list we need air date of the main series, but only way to get it takes extra 2 API queries (/v3/Group/{id}/main Series} then v3/Series/AniDB) and those queries are in sequence, so even slower. There is video of lag it causes in the UI in discord. 

So in my opinion solutions are either to prefetch tons of stuff or all of it like desktop does (which I don't want) or make extended result for webui only, as to not unnecessarily slowdown the endpoint for other clients.

Cazzar:
@hidden4003 I understand that side of things, though generally speaking client-specific code is a really bad sign of things.
 if anything we should implement those extra rules in it's own set of controllers, or, work out how they can be included in the base API requests in a performant manner.

hidden4003:
@Cazzar My suggestion was also to make separate endpoint for this, separate controller might be even better. 

Regarding performance getting AniDB data is fast but getting MainSeries can be slow when default id it is not set, then it looks through all the series in a group.

revam:
Time to close this.  Also good morning.

revam:
> Regarding performance getting AniDB data is fast but getting MainSeries can be slow when default id it is not set, then it looks through all the series in a group.

It is already fetching all the series and the series directly within the group when creating the group dto to send, so we could've just adapted it a bit and conditionally include the "main" series.

PR #951, opened 2022-07-28
feature: add group main series id

revam:
and tie it into the auto grouper feature.

### DIFF
--- a/Shoko.Server/API/v3/Controllers/TreeController.cs
+++ b/Shoko.Server/API/v3/Controllers/TreeController.cs
@@ -9,6 +9,7 @@ using Shoko.Server.API.Annotations;
 using Shoko.Server.API.v3.Helpers;
 using Shoko.Server.API.v3.Models.Common;
 using Shoko.Server.API.v3.Models.Shoko;
+using Shoko.Server.Extensions;
 using Shoko.Server.Models;
 using Shoko.Server.Repositories;
 
@@ -48,7 +49,7 @@ namespace Shoko.Server.API.v3.Controllers
 
             return RepoFactory.GroupFilter.GetByParentID(filterID)
                 .Where(filter => showHidden || filter.InvisibleInClients != 1)
-                .OrderBy(OrderByName)
+                .OrderByName()
                 .ToListResult(filter => new Filter(HttpContext, filter), page, pageSize);
         }
 
@@ -68,7 +69,7 @@ namespace Shoko.Server.API.v3.Controllers
             {
                 groups = RepoFactory.AnimeGroup.GetAll()
                     .Where(group => group.AnimeGroupParentID.HasValue && User.AllowedGroup(group))
-                    .OrderBy(OrderByName);
+                    .OrderByName();
             }
             else
             {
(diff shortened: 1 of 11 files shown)


revam:
Will do the TreeController tomorrow.

revam:
Also blocked by https://github.com/ShokoAnime/Shoko.Commons/pull/8 and https://github.com/ShokoAnime/Shoko.Models/pull/7

revam:
I didn't write any migration code because I'm unsure how to handle it for existing series when a user have set up _a lot_ of custom groups, and just running the "Recreate all groups" action will add the auto main id to the groups for those using automatically generated groups.

Shoko.Server/API/v3/Controllers/TreeController.cs
```diff
@@ -88,27 +83,34 @@ public ActionResult<ListResult<Group>> GetFilteredGroups([FromRoute] int filterI
...
         /// Get a list of all the sub-<see cref="Group"/>s belonging to the <see cref="Group"/> with the given <paramref name="groupID"/> and which are present within the <see cref="Filter"/> with the given <paramref name="filterID"/>.
         /// </summary>
         /// <param name="filterID"><see cref="Filter"/> ID</param>
         /// <param name="groupID"><see cref="Group"/> ID</param>
+        /// <param name="includeMissing">Include <see cref="Series"/> with missing <see cref="Episode"/>s in the search.</param>
         /// <param name="randomImages">Randomise images shown for the <see cref="Group"/>.</param>
         /// <returns></returns>
         [HttpGet("Filter/{filterID}/Group/{groupID}/Group")]
-        public ActionResult<List<Group>> GetFilteredSubGroups([FromRoute] int filterID, [FromRoute] int groupID, [FromQuery] bool randomImages = false)
+        public ActionResult<List<Group>> GetFilteredSubGroups([FromRoute] int filterID, [FromRoute] int groupID, [FromQuery] bool includeMissing = false, [FromQuery] bool randomImages = false)
```
Cazzar, line 109: If we are adding new, optional fields, as a general rule of thumb, add them at the end of the parameter list.

This is namely because else it can break other code when you insert a new field in the list of stuff, as using named arguments isn't always the default action - yes this is a HTTP action, but better to avoid in general.

Shoko.Server/API/v3/Controllers/TreeController.cs
```diff
@@ -198,47 +211,50 @@ public ActionResult<List<Series>> GetSeriesInFilteredGroup([FromRoute] int filte
...
                 return null;
 
-            // Check if a default series is set.
-            defaultSeries = group.GetDefaultSeries();
-            if (defaultSeries != null && seriesIDs.Contains(defaultSeries.AnimeSeriesID))
-                return new Series(HttpContext, defaultSeries, randomImages);
+            // Return the auto or user selected main series if set.
+            var mainSeries = group.GetMainSeries();
+            if (mainSeries != null)
+                return new Series(HttpContext, mainSeries, randomImages);
```
Cazzar, line 247: Why are we doing this variable rename pointlessly?

Cazzar (changes requested):
I wasn't able to get through the whole lot, there seems to be a whole bunch of other changes in there at the same time, which if you are going through and making a PR, I would avoid renaming things for the sake of renaming to make stuff easier to review.

Shoko.Server/API/v3/Controllers/TreeController.cs
```diff
@@ -36,7 +36,7 @@ public class TreeController : BaseController
         /// <param name="page">The page index.</param>
         /// <returns></returns>
         [HttpGet("Filter/{filterID}/Filter")]
-        public ActionResult<List<Filter>> GetSubFilters([FromRoute] int filterID, [FromQuery] [Range(0, 100)] int pageSize = 50, [FromQuery] [Range(1, int.MaxValue)] int page = 1)
+        public ActionResult<ListResult<Filter>> GetSubFilters([FromRoute] int filterID, [FromQuery] [Range(0, 100)] int pageSize = 50, [FromQuery] [Range(1, int.MaxValue)] int page = 1)
```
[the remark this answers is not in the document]
revam: Removing the commit from the PR. I forgot to convert it to use the `ListResult` type when i converted the paginated endpoints.

Shoko.Server/API/v3/Controllers/TreeController.cs
revam: Because the value it would retrieve was different.

Shoko.Server/API/v3/Controllers/TreeController.cs
```diff
@@ -180,67 +181,12 @@ public ActionResult<List<Series>> GetSeriesInFilteredGroup([FromRoute] int filte
...
             return (recursive ? group.GetAllSeries() : group.GetSeries())
                 .Where(series => seriesIDs.Contains(series.AnimeSeriesID))
-                .OrderBy(OrderByAirDate)
+                .OrderByAirDate()
                 .Select(series => new Series(HttpContext, series))
                 .Where(series => series.Size > 0 || includeMissing)
                 .ToList();
         }
 
-        /// <summary>
```
revam, line 189: There is no need for a filter specific endpoint to get the main series for a group, if the group is in the filter then just use the normal endpoint to get the main series.

Shoko.Server/Extensions/IEnumerableExtenions.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            return group.SortName;
+        }
+
+        private static string OrderByName(SVR_GroupFilter group)
+        {
+            return group.GroupFilterName;
+        }
+    }
+}
```
Cazzar, line 46: A lot of these seem to be really pointless, abstracting away 1 line of logic, to be another line somewhere

Shoko.Server/Extensions/IEnumerableExtenions.cs
```diff
@@ -0,0 +1,46 @@
...
+    public static class IEnumerableExtensions
+    {
+        public static IOrderedEnumerable<SVR_AnimeSeries> OrderByAirDate(this IEnumerable<SVR_AnimeSeries> enumerable)
+        {
+            return enumerable.OrderBy(OrderByAirDate);
+        }
+
+        private static DateTime? OrderByAirDate(SVR_AnimeSeries series)
+        {
+            var airdate = series.GetAnime()?.AirDate;
```
da3dsoul, line 19: I would remove this and simply use null propagation.
Pseudo code:
```
Order by(a => a.GetAnime()?.AirDate ?? DateTime.MaxValue)
```

Shoko.Server/Extensions/IEnumerableExtenions.cs
revam: I'm too used to JS when it is an optimization for the runtime to declare a static method/function to use instead of in-lining it.

PR #954, opened 2022-08-04
feat(series): Make search not coerce strings to ascii when using non-latin languages in settings

akiduki18:
With these changes, when at least one non-latin language is selected in the list, search features won't coerce the characters to ASCII. 

Please comment on any further change or improvement that should be made. It's my first time with C# so there are probably a few

### DIFF
--- a/Shoko.Server/Utilities/NamingLanguage.cs
+++ b/Shoko.Server/Utilities/NamingLanguage.cs
@@ -1,4 +1,4 @@
-﻿namespace Shoko.Server.Utilities
+namespace Shoko.Server.Utilities
 {
     public class NamingLanguage
     {
@@ -7,6 +7,11 @@
         public string LanguageDescription
         {
             get { return Languages.GetLanguageDescription(Language.Trim().ToUpper()); }
+        }
+
+        public bool IsLatin
+        {
+            get { return Languages.IsLatinLanguage(Language.Trim().ToUpper()); }
         }
 
         public NamingLanguage()
@@ -22,5 +27,7 @@
         {
             return string.Format("{0} - ({1})", Language, LanguageDescription);
         }
+
+
     }
 }
--- a/Shoko.Server/Utilities/SeriesSearch.cs
+++ b/Shoko.Server/Utilities/SeriesSearch.cs
@@ -47,6 +47,8 @@ namespace Shoko.Server.Utilities
             if (!(grouping?.SelectMany(a => a.GetAllTitles()).Any() ?? false)) return null;
             SearchResult<List<SVR_AnimeSeries>> dist = null;
 
+            bool forceAscii = Languages.PreferredNamingLanguages.ContainsOnlyLatin();
+
             foreach (SVR_AnimeSeries item in grouping)
             {
                 foreach (string title in item.GetAllTitles())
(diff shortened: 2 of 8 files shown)


da3dsoul:
Remove the submodule update on Commons

da3dsoul:
Also not sure what happened to SeriesSearch.cs

Shoko.Server/Utilities/SeriesSearch.cs
```diff
@@ -1,372 +1,378 @@
...
+
+            ParallelQuery<SVR_AnimeSeries> allSeries = 
+                RepoFactory.AnimeSeries.GetAll().AsParallel().Where(a =>
+                a?.GetAnime() != null && (a.GetAnime().GetAllTags().Count==0 || !a.GetAnime().GetAllTags().FindInEnumerable(user.GetHideCategories())));
+            
+            ParallelQuery<AniDB_Tag> allTags = RepoFactory.AniDB_Tag.GetAll().AsParallel()
+                .Where(a =>
+                {
+                    return !user.GetHideCategories().Contains(a.TagName) &&
                            !TagFilter.IsTagBlackListed(a.TagName, tagFilter);
```
revam, line 168: I know I'm at fault on this line, but can you **please** change the line endings for the file back to CRLF?

revam: It makes it easier to see what have changed and also won't override the git (blame) history for unrelated lines

revam (changes requested):
Overall it looks ok, but i will still request changing the line endings.

Shoko.Server/Utilities/SeriesSearch.cs
akiduki18: Oh, so that's why the diff is so crazy. Should I put the line endings change in a separate commit and add a . git-blame-ignore-revs or I just leave them as they are?

akiduki18:
Should be ok now. It's my first time working with windows too (and it's line endings hell). 

I know I'm new here, but it probably would help everybody to standardize the line endings to something. I can do it myself if you want.

I should I put a separate PR to the commons project right?

revam:
> Should be ok now. It's my first time working with windows too (and it's line endings hell).
> 
> I know I'm new here, but it probably would help everybody to standardize the line endings to something. I can do it myself if you want.
> 
> I should I put a separate PR to the commons project right?

The problem is that the project uses both LF and CRLF because it's just that _old_, and changing the line ending on existing files that work just fine will just override the git (blame) history and not provide any functional changes so it's adviced against ding. As the saying goes "don't fix it if it ain't broken." Fixing mixed line endings on the other hand — like in SeriesSearch.cs — is ok.

Also, new files should be commited using LF (if possible) and only the older files shold still be using CRLF at this point.

Shoko.Server/Utilities/NamingLanguage.cs
```diff
@@ -22,5 +27,7 @@ public override string ToString()
         {
             return string.Format("{0} - ({1})", Language, LanguageDescription);
         }
+
```
revam, line 30: Remove the added white spaces please

revam (changes requested):
Also found some new empty lines, and can also open a PR for the changes you did to [Shoko.Commons](https://github.com/ShokoAnime/Shoko.Commons)? Those changes needs to be merged before these so this PR can add the updated reference for the submodule.

revam:
> Also found some new empty lines, and can also open a PR for the changes you did to [Shoko.Commons](https://github.com/ShokoAnime/Shoko.Commons)? Those changes needs to be merged before these so this PR can add the updated reference for the submodule.

The consequences for not adding the changes to the Shoko.Commons sub-module first is that it will fail to build in the CI pipeline because either a) it cannot checkout the referenced commit or b) because the reference was not updated and it was using the existing reference to Shoko.Commons without the changes then it fails on an invalid method signature in in the Misc class.

Issue #989, opened 2022-08-13
Docker image not using packaged webui


Queuecumbr:
﻿## PLEASE READ BEFORE POSTING YOUR ISSUE

Please create GitHub issues only for bugs and feature requests. If you need help with using Shoko or have a question, join us on Discord.

https://discord.gg/vpeHDsg

===== Guidelines ====

- Make sure you are using the latest **Stable** or **Daily** version of Shoko Server otherwise your issue will be closed.
- Search this repository (top of the page) for the issue or feature you are reporting, make sure it has not been fixed or reported before.
- If reporting an issue, then you must attach the relevant Shoko Server and Shoko Desktop **LOG FILE**. 
- If requesting a new feature, please explain in detail the feature itself and the benefits of it being added.
- If requesting cosmetic changes, please explain in detail why they should be made. 
- The template below **MUST BE USED** when submitting an issue, no exceptions.

## VERSION INFORMATION

Server Version: Docker Daily

Desktop Version: N/A

If you've compiled your own version, please use the last commit you compiled.

## LOG FILE

N/A

## DESCRIPTION

I have been using the Docker images that you guys provide for quite some time. It works great for my needs and I love it. I recently noticed that the WebUI was having a lot of breakage for me, and since I regularly update my docker image I was curious if something upstream broke something. After some time investigating, I noticed that the WebUI was not using the latest version that was packaged which is located at `/usr/src/app/build/webui`. Instead, the server will point to '/home/shoko/.shoko/Shoko.CLI/webui' or download the latest version there if not present. Now if you follow the Docker installation for Shoko, a volume would be mounted from the host into '/home/shoko/.shoko/Shoko.CLI/' so that server databases and configurations will be persistent. Now since I had been using this setup for a few versions, the WebUI was never being updated ever since I first set it up which left an aging version of the WebUI in this mounted volume and would be used instead of the updated version that came with the docker image.

While you can work around this by toggling a WebUI update manually, this would be undesirable since the philosophy behind docker is that it has everything it already needs up front, and if anything needs to be updated, a new docker image should be built and used instead rather than modifying the current image.

## STEPS TO REPRODUCE

* Follow the [docker installation steps](https://docs.shokoanime.com/server/install/#docker-compose-linux), and run a container using the latest Shoko image. Make sure that /home/shoko/.shoko/Shoko.CLI/ is mounting for persistent storage. Docker-Compose may be the easiest to get up and running.
[...]

PR #990, opened 2022-08-13
Force docker image to use packaged WebUI

Queuecumbr:
#989

This will have a side effect where if an existing docker container is updated, any old WebUI will be forcibly removed and replaced with a symlink.

### DIFF
--- a/dockerentry.sh
+++ b/dockerentry.sh
@@ -64,5 +64,9 @@ UMASK set:    $(umask)
 -------------------------------------
 "
 
+# Make sure we use the packaged WebUI
+rm -rf /home/shoko/.shoko/Shoko.CLI/webui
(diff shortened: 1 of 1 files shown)


Cazzar:
FYI - there is further discussion going on about this internally about some earlier decisions, etc

Queuecumbr:
No worries. This is just what I had in mind as a quick fix if anything. I know there's a lot that has to be factored in before making a decision where to go. This is by no means meant to be a final solution, so I won't mind if it's turned down :)

da3dsoul:
@hidden4003 @harshithmohan this is done, right?

hidden4003:
Not really, it will force all users mount additional docker volume for webui files, I don't see the point in this.

PR #991, opened 2022-08-13
feature: expose shoko events through SignalR + more

revam:
expose the Shoko events through SignalR and refactor the plugin file events.

### DIFF
--- a/Shoko.Server/API/SignalR/Models/EpisodeInfoUpdatedEventSignalRModel.cs
+++ b/Shoko.Server/API/SignalR/Models/EpisodeInfoUpdatedEventSignalRModel.cs
@@ -1,0 +1,32 @@
+using Shoko.Plugin.Abstractions;
+using Shoko.Server.Models;
+using Shoko.Server.Repositories;
+
+namespace Shoko.Server.API.SignalR.Models
+{
+    public class EpisodeInfoUpdatedEventSignalRModel
+    {
+        public EpisodeInfoUpdatedEventSignalRModel(EpisodeInfoUpdatedEventArgs eventArgs)
+        {
+            var series = RepoFactory.AnimeSeries.GetByAnimeID(eventArgs.AnimeInfo.AnimeID);
+            EpisodeID = ((SVR_AnimeEpisode)eventArgs.EpisodeInfo).AnimeEpisodeID;
+            SeriesID = series.AnimeSeriesID;
+            GroupID = series.AnimeGroupID;
+        }
+
+        /// <summary>
+        /// Shoko episode id.
+        /// </summary>
+        public int EpisodeID { get; set; }
+
+        /// <summary>
+        /// Shoko series id.
+        /// </summary>
+        public int SeriesID { get; set; }
+
+        /// <summary>
+        /// Shoko group id.
+        /// </summary>
+        public int GroupID { get; set; }
+    }
+}
--- a/Shoko.Server/API/SignalR/Models/FileHashedEventSignalRModel.cs
+++ b/Shoko.Server/API/SignalR/Models/FileHashedEventSignalRModel.cs
@@ -1,0 +1,11 @@
+using Shoko.Plugin.Abstractions;
+
+namespace Shoko.Server.API.SignalR.Models
+{
+    public class FileHashedEventSignalRModel : FileEventSignalRModel
+    {
+        public FileHashedEventSignalRModel(FileHashedEventArgs eventArgs) : base(eventArgs)
+        {
+        }
+    }
+}
--- a/Shoko.Server/API/SignalR/ShokoEventEmitter.cs
+++ b/Shoko.Server/API/SignalR/ShokoEventEmitter.cs
@@ -1,0 +1,60 @@
+using System;
+using Microsoft.AspNetCore.SignalR;
+using Shoko.Plugin.Abstractions;
+using Shoko.Server.API.SignalR.Models;
+using Shoko.Server.Providers.AniDB;
+using Shoko.Server.Providers.AniDB.Interfaces;
+
+namespace Shoko.Server.API.SignalR
+{
+    public class ShokoEventEmitter : IDisposable
+    {
+        private IHubContext<ShokoEventHub> Hub { get; set; }
+        private IShokoEventHandler EventHandler { get; set; }
+
+        public ShokoEventEmitter(IHubContext<ShokoEventHub> hub, IShokoEventHandler events)
+        {
+            Hub = hub;
+            EventHandler = events;
+            EventHandler.FileDetected += OnFileDetected;
+            EventHandler.FileHashed += OnFileHashed;
+            EventHandler.FileMatched += OnFileMatched;
+            EventHandler.SeriesUpdated += OnSeriesUpdated;
+            EventHandler.EpisodeUpdated += OnEpisodeUpdated;
+        }
+
+        public void Dispose()
+        {
+            EventHandler.FileDetected -= OnFileDetected;
+            EventHandler.FileHashed -= OnFileHashed;
+            EventHandler.FileMatched -= OnFileMatched;
+            EventHandler.SeriesUpdated -= OnSeriesUpdated;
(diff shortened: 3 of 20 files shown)


revam:
The file events are tested on my machine:tm:
image
image

Cazzar:
Just a noticed pattern - it seems very often you seem to be adding extra changes unrelated to what you are doing.
For example
1. adding the docblocks to random objects, while it's not horrible of an idea - it makes it more complex to review.
2. Why are you adding the `: EventArgs` to some of the already existing event args.
If these are done for good reason, it's usually worth noting into the PR description

revam:
1. I didn't add any doc-blocks to any "random" objects in this PR — every added doc-block is followed by an added or updated field. I will reverse back the order of the fields in `FileEventArgs` though, so it's easier to see what actually changed.
2. Because to me is an `*EventArgs` not extending `System.EventArgs` (or a derived class) the same as an `*Exception` not extending `System.Exception`. Both patterns work in practice, but personally I don't like it.

Will try to write more details in the description next time.

revam:
Fixed up the order of the existing fields, removed some empty lines at the beginning of the new files, and removed unused dependencies/imports in the new files.

Cazzar:
> 1. I didn't add any doc-blocks to any "random" objects in this PR — every added doc-block is followed by an added or updated field. I will reverse back the order of the fields in `FileEventArgs` though, so it's easier to see what actually changed.
> 2. Because to me is an `*EventArgs` not extending `System.EventArgs` (or a derived class) the same as an `*Exception` not extending `System.Exception`. Both patterns work in practice, but personally I don't like it.
> 
> Will try to write more details in the description next time.

For future reference, I would suggest discussing changes you want to make "because you don't like it" because there may be unknown side-effects, or reasons it has been done explicitly.

The reason I said random, is when looking at the issue itself, and documentation you had included, a lot of it seemed unrelated, to the scope of "Expose x events"

Shoko.Server/API/SignalR/Models/EpisodeInfoUpdatedEventSignalRModel.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    public class EpisodeInfoUpdatedEventSignalRModel
+    {
+        public EpisodeInfoUpdatedEventSignalRModel(EpisodeInfoUpdatedEventArgs eventArgs)
+        {
+            var series = RepoFactory.AnimeSeries.GetByAnimeID(eventArgs.AnimeInfo.AnimeID);
+            EpisodeID = ((SVR_AnimeEpisode)eventArgs.EpisodeInfo).AnimeEpisodeID;
+            SeriesID = series.AnimeSeriesID;
+            GroupID = series.AnimeGroupID;
+        }
```
Cazzar, line 15: This as a whole seems like a N+1 issue, why do we need to have this lookup - is it better to have it in the EventArgs model itself?

Shoko.Server/API/SignalR/Models/FileHashedEventSignalRModel.cs
```diff
@@ -0,0 +1,11 @@
...
+
+namespace Shoko.Server.API.SignalR.Models
+{
+    public class FileHashedEventSignalRModel : FileEventSignalRModel
+    {
+        public FileHashedEventSignalRModel(FileHashedEventArgs eventArgs) : base(eventArgs)
+        {
+        }
+    }
+}
```
Cazzar, line 11: Unless this is adding some sort of value, it is probably worth just passing in the base class.

Shoko.Server/API/SignalR/ShokoEventEmitter.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        private async void OnSeriesUpdated(object sender, SeriesInfoUpdatedEventArgs e)
+        {
+            await Hub.Clients.All.SendAsync("SeriesUpdated", new SeriesInfoUpdatedEventSignalRModel(e));
+        }
+
+        private async void OnEpisodeUpdated(object sender, EpisodeInfoUpdatedEventArgs e)
+        {
+            await Hub.Clients.All.SendAsync("EpisodeUpdated", new EpisodeInfoUpdatedEventSignalRModel(e));
+        }
```
Cazzar, line 58: For this, why aren't we just sending in the EventArgs directly for most of these, it seems like pointless encapsulation?

Cazzar (commented):
Further to this, when making any changes to Shoko.Plugin.Abstractions, it will force any person who uses daily and a custom plugin to rebuild it, because the dependency changes, so I would suggest only making necessary changes, or batching them together.

revam:
> Further to this, when making any changes to Shoko.Plugin.Abstractions, it will force any person who uses daily and a custom plugin to rebuild it, because the dependency changes, so I would suggest only making necessary changes, or batching them together.

1. I just batched them togther.
2. I'm using my plugin built for 4.1.1 and it's working fine — even on the daily built earlier today — because I didn't touch any of the renamer events, only the new (and hopefully still unused) events introduced in (I think it was) 4.1.2.

Shoko.Server/API/SignalR/Models/EpisodeInfoUpdatedEventSignalRModel.cs
revam: The plugin abstraction uses the SVR_AniDB_Anime for it's IAnime implementation, The SVR_AniDB_Anime model does not contain the shoko ids. In my opinion if we want to fix having to look up the series from the anime id then we need to better align the plugin abstractions with how shoko internally models the data.

Shoko.Server/API/SignalR/Models/FileHashedEventSignalRModel.cs
revam: We can add fields to it in the future if needed. Yes we can change it.

Shoko.Server/API/SignalR/ShokoEventEmitter.cs
revam: Because in both Web UI and Shokofin we don't need the extra data exposed to the plugins sent to the clients. And if we do need the data in the future then we can fetch it using the IDs sent instead.

Shoko.Server/API/SignalR/Models/FileHashedEventSignalRModel.cs
Cazzar: Yeah, I would just add it when it's needed, not trying to plan too far in the future

Shoko.Server/API/SignalR/ShokoEventEmitter.cs
Cazzar: Having extra data isn't a bad thing, you can just use what is needed

Cazzar:
> > Further to this, when making any changes to Shoko.Plugin.Abstractions, it will force any person who uses daily and a custom plugin to rebuild it, because the dependency changes, so I would suggest only making necessary changes, or batching them together.
> 
> 1. I just batched them togther.
> 2. I'm using my plugin built for 4.1.1 and it's working fine — even on the daily built earlier today — because I didn't touch any of the renamer events, only the new (and hopefully still unused) events introduced in (I think it was) 4.1.2.

2. I have seen this happen too many times before, I was just mentioning in terms of previous experience

Shoko.Server/API/SignalR/Models/EpisodeInfoUpdatedEventSignalRModel.cs
Cazzar: That would ultimately solve the N+1 issue that has been created here, because what is now happening is, we look up the details, create the interface details, pass it down to the event and look it back up.

Ultimately this becomes a potential problem for performance issues, and we should be grabbing it only once.

If you need more detail on the concept of N+1 issues: https://medium.com/doctolib/understanding-and-fixing-n-1-query-30623109fe89 while in rails, the concepts are entirely the same, and will become more evident when we move to EF, and don't really have the caches in the long run.

PR #992, opened 2022-08-17
Add series without files endpoint

harshithmohan:


Shoko.Server/API/v3/Controllers/SeriesController.cs
```diff
@@ -134,6 +134,26 @@ public ActionResult<List<SeriesRelation>> GetShokoRelationsBySeriesID([FromRoute
                 .ToList();
         }
         
+        /// <summary>
+        /// Get a paginated list of <see cref="Series"/> without local files, available to the current <see cref="User"/>.
+        /// </summary>
+        /// <param name="pageSize">The page size.</param>
+        /// <param name="page">The page index.</param>
+        /// <returns></returns>
+        [HttpGet("WithoutFiles")]
```
harshithmohan, line 143: Should this be `/Empty` instead? 🤔

revam: Nah. `WithoutFiles` signifies better what it returns imo.

Shoko.Server/API/v3/Controllers/SeriesController.cs
```diff
@@ -134,6 +134,26 @@ public ActionResult<List<SeriesRelation>> GetShokoRelationsBySeriesID([FromRoute
...
+        public ActionResult<ListResult<Series>> GetSeriesWithoutFiles([FromQuery] [Range(0, 100)] int pageSize = 50, [FromQuery] [Range(1, int.MaxValue)] int page = 1)
+        {
+            var user = User;
+            return RepoFactory.AnimeSeries.GetAll()
+                .Select(series => (series, seriesName: series.GetSeriesName().ToLowerInvariant()))
+                .Where(tuple => {
+                    var (series, seriesName) = tuple;
+                    return user.AllowedSeries(series) && series.GetAnimeEpisodesCountWithVideoLocal() == 0;
+                })
+                .OrderBy(a => a.seriesName)
```
revam, line 153: We don't need to get the series name early if we're only going to use it for ordering the result.

revam: I'm too sick to figure out how to do multi-line suggestions, so here is what I would've done since the series name is not needed before the `.OrderBy`;
```cs
            return RepoFactory.AnimeSeries.GetAll()
                .Where(series => user.AllowedSeries(series) && series.GetAnimeEpisodesCountWithVideoLocal() == 0)
                .OrderBy(series => series.GetSeriesName().ToLowerInvariant())
                .ToListResult(tuple => new Series(HttpContext, tuple.series), page, pageSize);
```

revam (changes requested):
It almost LGTM. Though you can merge it as-is I would prefer if you took this sick man's feedback under consideration. :face_with_thermometer:

Shoko.Server/API/v3/Controllers/SeriesController.cs
harshithmohan: Done

revam (approved):
LGTM

Shoko.Server/API/v3/Controllers/SeriesController.cs
```diff
@@ -134,6 +134,22 @@ public ActionResult<List<SeriesRelation>> GetShokoRelationsBySeriesID([FromRoute
...
         }
         
+        /// <summary>
+        /// Get a paginated list of <see cref="Series"/> without local files, available to the current <see cref="User"/>.
+        /// </summary>
+        /// <param name="pageSize">The page size.</param>
+        /// <param name="page">The page index.</param>
+        /// <returns></returns>
+        [HttpGet("WithoutFiles")]
+        public ActionResult<ListResult<Series>> GetSeriesWithoutFiles([FromQuery] [Range(0, 100)] int pageSize = 50, [FromQuery] [Range(1, int.MaxValue)] int page = 1)
```
Cazzar, line 144: FWIW - better choice for `page` is to use `uint` because that itself inherently doesn't support negative values (and gives another bits worth of value for it)

Shoko.Server/API/v3/Controllers/SeriesController.cs
```diff
@@ -134,6 +134,22 @@ public ActionResult<List<SeriesRelation>> GetShokoRelationsBySeriesID([FromRoute
...
+        {
+            var user = User;
```
Cazzar, line 146: What is the use of this temporary variable? Shouldn't this have been in-lined for the one usage?

PR #994, opened 2022-08-19
Add ARM64 support

harshithmohan:


Cazzar:
Is it better to add an arm specific docker file?

harshithmohan:
> Is it better to add an arm specific docker file?

It works on x86 also, just debian got changed to ubuntu. It has to be the same dockerfile if we have to push both archs to the same tag (like in the latest commit).
Or I can do it separate it that's better

harshithmohan:
image
Looks like this on docker hub (that's my repo)

Cazzar:
> image Looks like this on docker hub (that's my repo)

TBH that's just the main part of tagging  - having the seperated docker files means that if we need to make differences to the builds, it is easier.

Cazzar:
For naming conventions, can we look at changing to `Dockerfile.{arch}` as a standard naming - as this is used a lot in other places, ex: https://github.com/microsoft/docker

.github/workflows/docker-daily.yml
```diff
@@ -6,17 +6,49 @@ on:
...
+    name: Build docker image for amd64
     runs-on: ubuntu-latest
 
     steps:
     - uses: actions/checkout@master
       with:
         submodules: recursive
     - name: Build the Docker image
-      run: docker build . --file Dockerfile --tag ${{ secrets.DOCKER_REPO }}:daily
+      run: docker build . --file Dockerfile --tag ${{ secrets.DOCKER_REPO }}:daily-amd64
```
Cazzar, line 18: can we push to the same tag, just the different arch.

.github/workflows/docker-daily.yml
```diff
@@ -6,17 +6,49 @@ on:
...
     - name: Log into docker hub
       run: docker login --username ${{ secrets.DOCKER_USERNAME }} --password ${{ secrets.DOCKER_PASSWORD }}
     - name: Push daily to Docker Hub
-      run: docker push ${{ secrets.DOCKER_REPO }}:daily
+      run: docker push ${{ secrets.DOCKER_REPO }}:daily-amd64
```
Cazzar, line 22: can we push to the same tag, just the different arch.

harshithmohan:
huh, I didn't know. I will change it

.github/workflows/docker-daily.yml
```diff
@@ -6,17 +6,49 @@ on:
...
+    - uses: docker/setup-qemu-action@v1
+      name: Set up QEMU
+      with:
+        platforms: arm64
+    - uses: docker/setup-buildx-action@v1
+      name: Set up Docker Buildx
+    - name: Log into docker hub
+      run: docker login --username ${{ secrets.DOCKER_USERNAME }} --password ${{ secrets.DOCKER_PASSWORD }}
+    - name: Build and Push the Docker image
+      run: docker buildx build . --file Dockerfile_arm --tag ${{ secrets.DOCKER_REPO }}:daily-arm64 --platform linux/arm64 --push
```
Cazzar, line 41: can we push to the same tag, just the different arch.

.github/workflows/docker-manual.yml
```diff
@@ -13,16 +13,48 @@ on:
...
         ref: "${{ github.event.inputs.ref }}"
     - name: Build the Docker image
-      run: docker build . --file Dockerfile --tag ${{ secrets.DOCKER_REPO }}:${{ github.event.inputs.tag }}
+      run: docker build . --file Dockerfile --tag ${{ secrets.DOCKER_REPO }}:${{ github.event.inputs.tag }}-amd64
     - name: Log into docker hub
       run: docker login --username ${{ secrets.DOCKER_USERNAME }} --password ${{ secrets.DOCKER_PASSWORD }}
-    - name: Push daily to Docker Hub
-      run: docker push ${{ secrets.DOCKER_REPO }}:${{ github.event.inputs.tag }}
+    - name: Push to Docker Hub
+      run: docker push ${{ secrets.DOCKER_REPO }}:${{ github.event.inputs.tag }}-amd64
```
Cazzar, line 29: can we push to the same tag, just the different arch.

Cazzar (changes requested):

.github/workflows/docker-manual.yml
harshithmohan: We cannot, we have to push the tags separately and then combine them with manifest
We can delete them after manifest creation

harshithmohan: Let me try something

harshithmohan:
@Cazzar Looks like there's no way. We have to push the separate tags and then link them with the manifest file to the `daily` tag. If we push to the same tag, it just overwrites the existing image.

harshithmohan:
I could only think of some hacky ways to do it. Also, even the dotnet sdk images are in this way. Separate tags for each arch and then combined under a single tag for convenience.

harshithmohan:
image
this is what it looks like currently, btw

harshithmohan:
Well, this PR is messed up.. so I'll make another

PR #997, opened 2022-08-20
Add arm64 support

harshithmohan:
For previous comments, look at #994


.github/workflows/docker-daily.yml
```diff
@@ -6,17 +6,49 @@ on:
...
+    - uses: docker/setup-qemu-action@v1
+      name: Set up QEMU
+      with:
+        platforms: arm64
+    - uses: docker/setup-buildx-action@v1
+      name: Set up Docker Buildx
+    - name: Log into docker hub
+      run: docker login --username ${{ secrets.DOCKER_USERNAME }} --password ${{ secrets.DOCKER_PASSWORD }}
+    - name: Build and Push the Docker image
+      run: docker buildx build . --file Dockerfile.aarch64 --tag ${{ secrets.DOCKER_REPO }}:daily-arm64 --platform linux/arm64 --push
```
Cazzar, line 41: Is there a way we can build arm64 with buildx? This means we could simplify these two with a matrix.

PR #998, opened 2022-08-22
update git workflows and dockerfile

revam:
### Changes

- Added a release channel, commit digest and tag for all automated builds (no tag for the manual builds though)

  The release channel and commit digest are to-be used in Shoko Web UI to determine when/if it should show the
  update modal, but it can also be used for other things, such as, checking which commit the server was built against, etc.

- Added an automated pre-release for each push to `master`, complete with an automated changelog for all the commits since the last push.

- Added build actions for every release and pre-release we do. This change combined all the daily actions/jobs (except the docker build) into a single workflow, and every release actions/jobs (again, except the docker build) into a single workflow.

- Added artifacts to _all_ future releases on Github, not just to the action runs. This will make it easier for _all users_ (not just signed in users!) to download release zips from GitHub.

- Updated the `/api/v3/Init/Version` endpoint to conditionally include any _known_ key-value pairs stored in the assembly version extra field. Known keys include `commit`, `tag` and `channel`. Also replaced the `List<ComponentVersion>` with a `ComponentVersionSet`, making it easier to access the different versions (e.g. in JS `cVS.find(c => c.name === "Server").Version` →  `cVS.Server.Version`, etc.).

- Fixed the broken Web UI version returned in the v2 and v3 api.

  ### Example

  Example responses from `/api/v3/Init/Version`;

  #### Before
  image

  #### After (docker build)
  ```sh
  > docker build . --build-arg release=local --build-arg commit=$(git rev-parse HEAD) --build-arg version=5.0.0.1 --build-arg tag=v5.0.0-dev.1
  ```
   image

  #### After (debug in vscode)
  image


revam:
Do feel free to suggest changes and/or fixes. I haven't tested the actions yet since i don't know how to test them beforehand.

revam:
I've tested the docker build locally though.

Cazzar:
Are you able to provide image sizes of the different builds locally? 

Alternatively, we could just keep the .dockerignore the same and use the env `GITHUB_SHA` 

see: https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

revam:
The workflows are still untested. Will see if i can manage to test them tomorrow... I don't have enough time right now to do it.

Cazzar:
Further investigation `/p:Version={VersionNumber}` might work better, if we can get rid of the shared assembly info

ElementalCrisis:
> update the cli* and tray* workflows to also run on release, attaching the files in a zip with the release. this is to make it easier to download release zips from github for all users.

Kinda pointless and won't work if the following isn't run after compiling. 

dotnet publish -c Release -r win10-x64 --self-contained true -f net5.0-windows Shoko.TrayService

revam:
Personally I find the `SharedAssemblyInfo.cs` useful for programmatically modifying the assembly info. But if anyone have a better way to do it (it -> add a release channel and git sha digest to automated builds), then by all means, do tell.

Also do note that this PR have **nothing to do with automated build numbers/version-changes**. That will happen in another PR (if I'm the one doing it) since I don't want to delay this PR any longer.

da3dsoul:
I'm not sure if sed will work on the windows builds. We'll find out, I guess

.github/workflows/tray-framework-release.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Archive Release
+        shell: powershell
+        run: Compress-Archive .\\Shoko.Server\\bin\\Release\\net6.0-windows\\win10-x64\\publish .\\Shoko.TrayService_Framework_win10-x64.zip
+
+      - name: Upload Release
+        uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: ./ShokoServer*.zip
```
da3dsoul, line 46: This looks incorrect

.github/workflows/tray-standalone-release.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Archive Release
+        shell: powershell
+        run: Compress-Archive .\\Shoko.Server\\bin\\Release\\net6.0-windows\\win10-x64\\publish .\\Shoko.TrayService_Standalone_win10-x64.zip
+
+      - name: Upload Release
+        uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: ./ShokoServer*.zip
```
da3dsoul, line 46: This looks wrong

.github/workflows/cli-standalone-release.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Archive Release
+        shell: powershell
+        run: Compress-Archive .\\Shoko.Server\\bin\\Release\\net6.0\\${{ matrix.rid }}\\publish .\\Shoko.CLI_Standalone_${{ matrix.rid }}.zip
+
+      - name: Upload Release
+        uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: ./ShokoServer*.zip
```
da3dsoul, line 47: This too

.github/workflows/cli-framework-release.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Archive Release
+        shell: powershell
+        run: Compress-Archive .\\Shoko.Server\\bin\\Release\\net6.0\\publish .\\Shoko.CLI_Framework_any-x64.zip
+
+      - name: Upload Release
+        uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: ./ShokoServer*.zip
```
da3dsoul, line 46: And this

revam:
So I need to fix the `sed` usage on windows and maybe the `Compress-Archive` usage on linux.

revam:
Removed `sed`.

revam:
Also, the `Compress-Archive` cmdlet is available in linux.

revam:
I need to find a better way to get the previous tag... the current action module (or whatever it's called) it uses only grabs the last tag... which is not ideal when we have tags for both the server and the plugin abstractions in the same repo, also they won't conform to the required version format if they have a prefix (e.g. "v", etc.)..

revam:
I think it's ready.

revam:
It was not ready. Also, managed to actually test some aspects of the workflows locally. Will unmark it as a draft when I fix it... again.

revam:
I've tested all I could manage locally.

PR #999, opened 2022-08-25
Add base64 converter for chapter name in mediainfo

harshithmohan:
Not sure I wrote it properly.. but it works<sup>TM</sup>

### DIFF
--- a/Shoko.Server/Utilities/MediaInfoLib/MenuBase64Converter.cs
+++ b/Shoko.Server/Utilities/MediaInfoLib/MenuBase64Converter.cs
@@ -1,0 +1,57 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Shoko.Server.Utilities.MediaInfoLib
+{
+    public class MenuBase64Converter : JsonConverter
+    {
+        public override bool CanConvert(Type objectType)
+        {
+            return objectType.IsAssignableFrom(typeof(Dictionary<string,string>));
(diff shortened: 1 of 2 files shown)


harshithmohan:
I was not able to use the specific type to just target the object with @dt and #value

Shoko.Server/Utilities/MediaInfoLib/MenuBase64Converter.cs
```diff
@@ -0,0 +1,57 @@
...
+                        sanitizedDictionary[key] = value?.ToString();
+                        break;
+                    case JTokenType.Object:
+                    {
+                        var base64EncodedBytes = Convert.FromBase64String(value?["#value"]?.ToString() ?? "");
+                        sanitizedDictionary[key] = Encoding.UTF8.GetString(base64EncodedBytes).Trim();
+                        break;
+                    }
+                    default:
+                        throw new NotImplementedException();
```
harshithmohan, line 41: Was not sure what to put for default case.

Cazzar: Throw something more specific then a not implemented description, or at least provide a message

da3dsoul: If it's intentionally not implemented, I tend to go for a NotSupportedException

harshithmohan: I changed it to
`throw new NotSupportedException("Unknown type encountered in MenuStream.extra");`

PR #1002, opened 2022-09-30
Refactor CLI Project

bartimaeusnek:
+ possible breaking change: ProgramArguments-Singleton was removed (needs to be tested)

### DIFF
--- a/Shoko.CLI/Program.cs
+++ b/Shoko.CLI/Program.cs
@@ -51,9 +51,13 @@ public static class Program
 #region IConfigurationBuilderExtensions
     private static IConfigurationBuilder AddCommandLineIfNotNull(this IConfigurationBuilder configurationBuilder)
     {
-        var args = Environment.GetCommandLineArgs();
-        if (args is not null)
-            configurationBuilder.AddCommandLine(args);
+        var tmpArgs = Environment.GetCommandLineArgs();
+        var args = tmpArgs.Length > 1 ? new string[tmpArgs.Length - 1] : null;
+        if (args is null)
+            return configurationBuilder;
+        
+        Array.Copy(tmpArgs, 1, args, 0, args.Length);
+        configurationBuilder.AddCommandLine(args);
         return configurationBuilder;
     }
 
--- a/Shoko.CLI/Program.cs
+++ b/Shoko.CLI/Program.cs
@@ -1,89 +1,70 @@
-﻿using System;
+﻿#region
+using System;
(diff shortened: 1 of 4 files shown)


Shoko.CLI/Program.cs
```diff
@@ -1,89 +1,74 @@
...
-                    if (appAssembly != null)
-                    {
-                        config.AddUserSecrets(appAssembly, optional: true);
-                    }
-                }
+    private static void ConfigureServices(IServiceCollection services) => services.AddHostedService<Worker>();
 
-                config.AddEnvironmentVariables();
+#region IConfigurationBuilderExtensions
+    private static IConfigurationBuilder AddCommandLineIfNotNull(this IConfigurationBuilder configurationBuilder)
```
da3dsoul, line 52: Correct me if I'm wrong. This allows the existing command line stuff to work, but doesn't bind to the intermediate model like it used to. Is that right?

Shoko.CLI/Worker.cs
```diff
@@ -1,69 +1,101 @@
...
 
-        protected override async Task ExecuteAsync(CancellationToken stoppingToken)
-        {
-            if (!string.IsNullOrEmpty(Args?.Instance)) ServerSettings.DefaultInstance = Args.Instance;
+    private static void InitInstanceLogger() => ShokoServer.Instance.InitLogger();
+    
+    private static string? GetInstanceFromCommandLineArguments()
+    {
+        const int notFound = -1;
+        var       args     = Environment.GetCommandLineArgs();
```
da3dsoul, line 53: I personally don't care, but our current formatter settings would nuke that whitespace. I'm less anal about that stuff, as long as it makes sense for readability in the current context.

Shoko.CLI/Program.cs
bartimaeusnek: I'm not quite sure which model youre talking about. If you mean the ``services.AddSingleton(arguments);`` Singleton which did only contain the instance name, then yes, that is gone.

da3dsoul:
I don't have any issues with it. I'll let @Cazzar  be anal about formatting or whatever if he cares to. We haven't made any proper templates for code styles. I'll use whatever cazzar wants if he gives me one. I use Rider's defaults

Shoko.CLI/Program.cs
da3dsoul: Mk that's fine

PR #1005, opened 2022-10-01
Cleanup TrayService

bartimaeusnek:


Shoko.Server/Server/StartServer.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        const int notFound = -1;
+        var       args     = Environment.GetCommandLineArgs();
+        var       idx      = Array.FindIndex(args, x => string.Equals(x, "instance", StringComparison.InvariantCultureIgnoreCase));
+        if (idx is notFound)
+            return null;
+        if (idx >= args.Length - 1)
+            return null;
+        return args[idx + 1];
+    }
```
Cazzar, line 30: Given the fact that shoko's "instance" support is a really niche thing, I think this is over-engineering a solution for finding it, simply a `if argv[1] == 'instance' return argv[2]` would be fine.

Shoko.TrayService/App.xaml.cs
```diff
@@ -1,174 +1,167 @@
...
         }
+        catch (Exception) { /*ok no IPv6 available.*/ }
 
-        protected override void OnExit(ExitEventArgs e)
+        if (string.IsNullOrWhiteSpace(ip))
+            ip = GetLocalIpv4OrFallback();
+        else
+            ip = $"[{ip}]";
+        
+        var url = $"http://{ip}:{ServerSettings.Instance.ServerPort}";
```
Cazzar, line 100: Instead of doing this whole thing... why not just `http://localhost:{ServerSettings.Instance.ServerPort}`

Shoko.TrayService/Shoko.TrayService.csproj
```diff
@@ -15,6 +15,8 @@
         <Configurations>Debug;Release;ApiLogging</Configurations>
         <ApplicationIcon>..\Shoko.Server\db.ico</ApplicationIcon>
         <ErrorOnDuplicatePublishOutputFiles>false</ErrorOnDuplicatePublishOutputFiles>
+        <NoWarn>CA1416</NoWarn>
```
Cazzar, line 18: Would this be better to have only specific instances of this listed and ignored if required? Removing it on the whole project to me seems a little excessive.

Shoko.CLI/Worker.cs
```diff
@@ -4,73 +4,25 @@
...
     protected override Task ExecuteAsync(CancellationToken stoppingToken)
     {
-        SetInstanceIfNeeded();
-        InitInstanceLogger();
-        LoadSettings();
-        if (ServerCouldNotStart())
-            return Task.CompletedTask;
-        EnsureAniDBSocketInitialized();
-        AddEventHandlers();
+        StartServer.StartupServer(AddEventHandlers, ServerCouldStart);
```
Cazzar, line 21: Out of curiosity why refactor out this? If you're wanting it testable, you would need to have some sort of DI, but it is simply a static class which adds little to no benefit, given there isn't a huge amount of logic in the worker.

Shoko.TrayService/App.xaml.cs
```diff
@@ -1,174 +1,167 @@
...
-            menu.Items.Add(webui);
-            webui = new MenuItem();
-            webui.Header = "Exit";
-            webui.Click += (sender, args) => Dispatcher.Invoke(Shutdown);
-            menu.Items.Add(webui);
-            Icon.ContextMenu = menu;
-            Icon.MenuActivation = PopupActivationMode.All;
-            Icon.Visibility = Visibility.Visible;
+    private static          TaskbarIcon? _icon;
+    private static readonly Logger       Logger = LogManager.GetCurrentClassLogger();
```
Cazzar, line 26: Making variables align like this imo is pretty pointless and it also doesn't happen to occur a lot in the rest of the shoko codebase.

Shoko.TrayService/App.xaml.cs
```diff
@@ -1,174 +1,167 @@
...
-            ShokoServer.Instance.ServerShutdown += (sender, eventArgs) => Shutdown();
-            Utils.YesNoRequired += (sender, e) =>
-            {
-                e.Cancel = true;
-            };
+    private void InitialiseTaskbarIcon()
+    {
+        _icon = new TaskbarIcon{
+                                   ToolTipText = "Shoko Server", ContextMenu = CreateContextMenu(), MenuActivation = PopupActivationMode.All, Visibility = Visibility.Visible,
+                               };
```
Cazzar, line 65: If you're going to make this a multi-line object constructor, do each property as it's own line

Cazzar (changes requested):

Shoko.TrayService/Shoko.TrayService.csproj
bartimaeusnek: `CA1416: Validate platform compatibility` simply won't make any sense, in a windows only project.

Cazzar: I am aware on what it is, but, that still allows you to validate that you're not using Linux/OSX apis on a, windows-only project. There's two sides to the sword here.

Shoko.CLI/Worker.cs
bartimaeusnek: It removes duplicate code from CLI and TrayService, and merges that code into a smaller, easier to handle shape.

> If you're wanting it testable, you would need to have some sort of DI

wdym by that? You can literally just call it in a Testproject.

Shoko.Server/Server/StartServer.cs
bartimaeusnek: having a simple solution such as if[1] then [2] is prone to breaking and wont allow any further development of command line arguments, without rewriting. i also think such a solution would be too hacky, since it does not respect array bounds, and once you test on array bounds, you are really not any different from this solution.

Shoko.TrayService/App.xaml.cs
bartimaeusnek: please provide a complete .editorconfig for fomating issues, this is ReSharper default / detection of Shoko's Style.

Shoko.TrayService/App.xaml.cs
bartimaeusnek: please provide a complete .editorconfig for fomating issues, this is ReSharper default / detection of Shoko's Style.

Shoko.CLI/Worker.cs
Cazzar: If you wanted to test Worker.cs, you can't without a hard dependency to StartServer.

Shoko.TrayService/App.xaml.cs
Cazzar: That will be in the works but in the mean time, you can change this.

Shoko.TrayService/App.xaml.cs
Cazzar: That will be in the works but in the mean time, you can change this.

Cazzar (changes requested):

Shoko.CLI/Worker.cs
bartimaeusnek: But you already have a hard dependency on the Server Project from the Worker, simple because it starts the server? What exactly is the issue with that?

da3dsoul:
That was a mess. Merged as of 5037e332c55be7cba7d43f671f130ec84c36958f

PR #1007, opened 2022-10-10
File_Source DB Migrations

krbrs:
Added DB migrations for File_Source to make already present entries consistent with currently inserted values.
Please let me know if I forgot something.  Also, if somebody could test the migrations for other DBs than MariaDB that would be great!

### DIFF
--- a/Shoko.Server/Databases/MySQL.cs
+++ b/Shoko.Server/Databases/MySQL.cs
@@ -20,7 +20,7 @@ namespace Shoko.Server.Databases;
 public class MySQL : BaseDatabase<MySqlConnection>, IDatabase
 {
     public string Name { get; } = "MySQL";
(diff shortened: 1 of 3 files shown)


da3dsoul:
That was an oversight, rather than intended behavior. After thinking on it, though. I do at least want them to match. I'm still torn on whether I want the database model to be bound directly to the Enum. It would make a lot of things more simple

revam:
I'm in favor of the database model being directly tied to the enum(s). Also, we don't need to be strictly using anidb's terminology either.

da3dsoul:
Ok after some thought, for now, we will take this. (thank you). We will wait an making it an enum in the database model, only because Desktop is dependent on that, and I don't want to fix it. When we merge/separate Commons and Models, then it can be done

PR #1008, opened 2022-10-12
fixup .editorconfig

bartimaeusnek:
Adds missing properties to .editorconfig (automatically detected by Rider, non-standard only)
fixes broken properties

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,8 +5,88 @@ root = true
 
 # Don't use tabs for indentation.
 [*]
-indent_style = space
+indent_style = space 
 # (Please don't specify an indent_size here; that has too many unintended consequences.)
+
+
+# Microsoft .NET properties
+csharp_new_line_before_members_in_object_initializers = false
+csharp_new_line_before_open_brace = all
+csharp_preferred_modifier_order = private, protected, public, internal, sealed, new, override, virtual, abstract, static, extern, async, unsafe, volatile, readonly, required:suggestion
+csharp_preserve_single_line_blocks = true
+csharp_space_after_cast = false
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -45,10 +125,10 @@ dotnet_diagnostic.IDE0055.severity = warning
...
 dotnet_separate_import_directive_groups = false
 # Avoid "this." and "Me." if not necessary
-dotnet_style_qualification_for_field = false:refactoring
-dotnet_style_qualification_for_property = false:refactoring
-dotnet_style_qualification_for_method = false:refactoring
-dotnet_style_qualification_for_event = false:refactoring
+dotnet_style_qualification_for_field = false:suggestion # refactoring is not supported by all IDEs
+dotnet_style_qualification_for_property = false:suggestion
+dotnet_style_qualification_for_method = false:suggestion
+dotnet_style_qualification_for_event = false:suggestion
```
Cazzar, line 131: Are we able to bump this up from a suggestion to something higher - @da3dsoul thoughts?

da3dsoul: I think suggestion is one below warning. It'll be underlined yellow, but it doesn't build as a warning

da3dsoul: https://learn.microsoft.com/en-us/visualstudio/code-quality/use-roslyn-analyzers?view=vs-2022
That is correct according to this

Cazzar: Yeah, I would possibly be pushing it up to a warning if anything, Error seems _too_ strict right now

da3dsoul: I'm not against it

PR #1009, opened 2022-10-15
feature: add tvdb series refresh endpoints to the v3 api

revam:


revam:
Mimiced how the anidb refresh is done.
The direct refresh endpoint gives a boolean status if the refresh were performed immediately.

PR #1011, opened 2022-10-21
Implement TraktTV Ratelimiter

krbrs:
Tried implementing a rate limiter to get rid of http 429 errors on syncing to Trakt. I took the limiter from TVDB and altered it a bit.
TraktTV wants 1 Second delay on POST according to Apiary: https://trakt.docs.apiary.io/#introduction/rate-limiting

Maybe someone can have a look at what I did there - I guess there is room for improvement, lol

PR #1012, opened 2022-10-24
Implement TraktTV Ratelimiter - MK-II

krbrs:
Same as #1011  - but with a fresh repo.. I could not fix it.

Tried implementing a rate limiter to get rid of http 429 errors on syncing to Trakt. I took the limiter from TVDB and altered it a bit.
TraktTV wants 1 Second delay on POST according to Apiary: https://trakt.docs.apiary.io/#introduction/rate-limiting

Maybe someone can have a look at what I did there - I guess there is room for improvement, lol

### DIFF
--- a/Shoko.Server/Providers/TraktTV/TraktRateLimiter.cs
+++ b/Shoko.Server/Providers/TraktTV/TraktRateLimiter.cs
@@ -1,0 +1,78 @@
+﻿using System;
+using System.Diagnostics;
+using System.Threading;
+using NLog;
+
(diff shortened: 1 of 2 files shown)


krbrs:
I compiled a build and confirmed it is working; the only thing is, it spams the Log a bit with the throttling messages, as throttling occurs rather often (because it's only 1 sec wait). Maybe the messages can or should be suppressed

da3dsoul:
Oh I didn't get a notification that this left draft status

PR #1017, opened 2022-11-17
fix: add last modified and update file watcher

revam:
# Changes

- Exclude the file while setting file permissions


revam:
For other people looking at this, the discussion happened on discord.

revam:
> Thirdly, this breaks the current recommended method to re-sort files, which is to move them back to a drop source.

Also, I explicitly tried to avoid that with the initial implementation. It only affected the events fired when _the file metadata_ (e.g. last write, last access, file size changed, file permission changed, etc.) were changed.

da3dsoul:
Looks good

PR #1019, opened 2022-11-29
feature: add DateTimeImported to VideoLocal

revam:
We're currently not tracking when a file was imported. This makes ordering the "Recently Imported" panel on the dashboard difficult to do correctly, so this PR adds a nullable date that's set when a file is either automatically imported by the system or manually imported by the user. It also unsets the date if a manual link is removed by the user, since it is no longer considered as _imported_.

### Changes in this PR

- Added `DateTime? DateTimeImported` to `VideoLocal`. The naming is to make it consistent with the existing dates present on the model. (We should probably consolidate on a common naming scheme when moving to EF Core, but for now I'm adding this as-is).

- Added a migration path for each of the database types. This will simply set the import date to the file creation date if the file is currently imported.

- Updated the `CommandRequest_ProcessFile` and `CommandRequest_LinkFileManually` to set the import date.

- Updated the manually link endpoints in both v1 and v3 to reset the import date once all cross-references have been removed from the file.

- Updated the recently episode/series endpoints (in the v3 api) to use the new field to sort the results.

- Made sure cross-references are removed _before_ queuing the new manual link command in the v3 api.

- Exposed the import date on the file model in the v3 api.

- Removed one extra white-space. :slightly_smiling_face:

### DIFF
--- a/Shoko.Server/Databases/SQLServer.cs
+++ b/Shoko.Server/Databases/SQLServer.cs
@@ -22,7 +22,7 @@ namespace Shoko.Server.Databases;
 public class SQLServer : BaseDatabase<SqlConnection>, IDatabase
 {
     public string Name { get; } = "SQLServer";
-    public int RequiredVersion { get; } = 98;
+    public int RequiredVersion { get; } = 99;
 
     public void BackupDatabase(string fullfilename)
     {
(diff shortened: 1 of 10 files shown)


da3dsoul:
I would name it LastUpdated

da3dsoul:
SQL migration:
```sql
UPDATE VideoLocal SET LastUpdated = DateCreated WHERE DateCreated IS NOT NULL;
```

revam:
> I would name it LastUpdated

We already have an update field (DateTimeUpdate) that's exclusively used by the hash command. And I don't want the import date (and import order!) to shift when someone triggers a rehash.



> SQL migration:
> 
> ```sql
> UPDATE VideoLocal SET LastUpdated = DateCreated WHERE DateCreated IS NOT NULL;
> ```

Will update replace the C# part of the migration with this.

revam:
> SQL migration:
> 
> ```sql
> UPDATE VideoLocal SET LastUpdated = DateCreated WHERE DateCreated IS NOT NULL;
> ```

Actually, small problem. The current migration only updates the import date for the files with existing cross-references. Not all the files.

da3dsoul:
Fair enough on the naming

revam:
Do you have a SQL that also check the cross-reference table if a match exists, or should we keep it as C# code? @da3dsoul

da3dsoul:
I'd need to open it in a console to check

Shoko.Server/Databases/SQLServer.cs
```diff
@@ -647,7 +647,9 @@ public override bool HasVersionsTable()
         new DatabaseCommand(95, 1, "UPDATE AniDB_File SET File_Source = 'Web' WHERE File_Source = 'www'; UPDATE AniDB_File SET File_Source = 'BluRay' WHERE File_Source = 'Blu-ray'; UPDATE AniDB_File SET File_Source = 'LaserDisc' WHERE File_Source = 'LD'; UPDATE AniDB_File SET File_Source = 'Unknown' WHERE File_Source = 'unknown';"),
         new DatabaseCommand(96, 1, "ALTER TABLE AniDB_GroupStatus ALTER COLUMN GroupName nvarchar(max); ALTER TABLE AniDB_GroupStatus ALTER COLUMN EpisodeRange nvarchar(max);"),
         new DatabaseCommand(97, 1, "CREATE INDEX IX_AniDB_Episode_EpisodeType ON AniDB_Episode(EpisodeType);"),
-        new DatabaseCommand(98, 1, "ALTER TABLE AniDB_Episode_Title ALTER COLUMN Title nvarchar(max) NOT NULL;")
+        new DatabaseCommand(98, 1, "ALTER TABLE AniDB_Episode_Title ALTER COLUMN Title nvarchar(max) NOT NULL;"),
+        new(99, 1, "ALTER TABLE VideoLocal ADD DateTimeImported datetime DEFAULT NULL;"),
```
revam, line 651: Forgot to add `DatabaseCommand` to initializators in the SQLServer commands. :facepalm:

revam:
Added the SQL statements provided by @da3dsoul. Any other changes that needs to be made, or is it ok to merge this now?

revam:
Removed the now unused database fix helper.

PR #1021, opened 2022-12-02
Add plex endpoints

revam:
Added new plex endpoints in the v3 api to authenticate the user, select the active server and library, and to sync the user.
Also fixed the v2 endpoints so they do what you would expect (e.g. the sync user endpoint would only sync for that one user, etc.)

### Changes in this PR

- Fixed the v2 endpoints

- Fixed the sync all endpoint in v3

- Added user sync endpoints

- Added plex auth. and server/library endpoints.

### DIFF
--- a/Shoko.Server/API/v2/Modules/PlexWebhook.cs
+++ b/Shoko.Server/API/v2/Modules/PlexWebhook.cs
@@ -238,6 +238,7 @@ public class PlexWebhook : BaseController
     {
         return CallPlexHelper(h =>
         {
+            if (!h.IsAuthenticated) return false;
             h.InvalidateToken();
             return true;
         });
@@ -247,9 +248,8 @@ public class PlexWebhook : BaseController
     [HttpGet("sync")]
     public ActionResult Sync()
     {
-        Analytics.PostEvent("Plex", "SyncOne");
-
         _commandFactory.Create<CommandRequest_PlexSyncWatched>(c => c.User = HttpContext.GetUser()).Save();
+
         return APIStatus.OK();
     }
 
@@ -257,16 +257,25 @@ public class PlexWebhook : BaseController
     [HttpGet("sync/all")]
     public ActionResult SyncAll()
     {
-        ShokoServer.Instance.SyncPlex();
+        var users = RepoFactory.JMMUser.GetAll()
+            .Where(user => !string.IsNullOrEmpty(user.PlexToken));
+
+        foreach (var user in users)
+            _commandFactory.Create<CommandRequest_PlexSyncWatched>(c => c.User = user).Save();
+
         return APIStatus.OK();
     }
 
     [Authorize("admin")]
-    [HttpGet("sync/{id}")]
-    public ActionResult SyncForUser(int uid)
+    [HttpGet("sync/{userID}")]
+    public ActionResult SyncForUser([FromQuery] int userID)
     {
-        JMMUser user = HttpContext.GetUser();
-        ShokoServer.Instance.SyncPlex();
+        var user = RepoFactory.JMMUser.GetByID(userID);
+        if (user == null)
+            return NotFound("Unable to find a user with the given userID.");
+
+        _commandFactory.Create<CommandRequest_PlexSyncWatched>(c => c.User = user).Save();
+
         return APIStatus.OK();
     }
 
--- a/Shoko.Server/API/v3/Controllers/PlexController.cs
+++ b/Shoko.Server/API/v3/Controllers/PlexController.cs
@@ -1,0 +1,203 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using Shoko.Server.API.Annotations;
+using Shoko.Server.API.v3.Plex;
+using Shoko.Server.Plex;
+using Shoko.Server.Settings;
+
+namespace Shoko.Server.API.v3.Controllers;
+
+[ApiController]
+[Route("/api/v{version:apiVersion}/[controller]")]
+[ApiV3]
+[Authorize]
+public class PlexController : BaseController
+{
+    private const string AlreadyAuthenticated = "User has already authenticated with plex yet. Invalidate the current token before re-authenticating.";
+
+    private const string NotAuthenticated = "User has not authenticated with plex yet. Authenticate first before retrying.";
+
+    private const string NoSelectedServer = "A server has not been selected yet. Select a server first.";
+
+    private const string InvalidServerSelection = "Invalid server selection.";
+
+    private const string InvalidLibrarySelection = "Selected libraries are invalid.";
+
+    /// <summary>
+    /// Get an OAuth2 authenticate url to authenticate the current user.
+    /// </summary>
+    /// <returns></returns>
+    [HttpGet("Auth/Authenticate")]
+    public ActionResult<string> GetOAuthRequestUrl([FromQuery] bool force = false)
+    {
+        var helper = GetHelperForUser();
+
+        if (helper.IsAuthenticated && !force)
+            return BadRequest(AlreadyAuthenticated);
+
+        return helper.LoginUrl;
+    }
+
+    /// <summary>
+    /// Check if the current user is currently authenticated against the plex api.
+    /// </summary>
+    /// <returns></returns>
+    [HttpGet("Auth/IsAuthenticated")]
+    public ActionResult<bool> IsAuthenticated()
+    {
+        var helper = GetHelperForUser();
+
+        return helper.IsAuthenticated;
+    }
+
+    /// <summary>
+    /// Invalidate and remove the current plex authentication token.
+    /// </summary>
+    /// <returns></returns>
+    [HttpPost("Auth/InvalidateToken")]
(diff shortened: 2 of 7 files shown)


Cazzar:
"Fixed the v2 endpoints" What was wrong with them?

Shoko.Server/API/v2/Modules/PlexWebhook.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
```
Cazzar, line 1: Can you check your editor config, this looks to have changed the file encoding (UTF8 w/BOM to w/o BOM or the other way around)

Shoko.Server/API/v2/Modules/PlexWebhook.cs
```diff
@@ -247,26 +248,36 @@ public bool InvalidateToken()
     [HttpGet("sync")]
     public ActionResult Sync()
     {
-        Analytics.PostEvent("Plex", "SyncOne");
+        var user = HttpContext.GetUser();
+
+        _commandFactory.Create<CommandRequest_PlexSyncWatched>(c => c.User = user).Save();
```
Cazzar, line 253:
```suggestion
        _commandFactory.Create<CommandRequest_PlexSyncWatched>(c => c.User = HttpContext.GetUser()).Save();
```

Shoko.Server/API/v2/Modules/PlexWebhook.cs
```diff
@@ -247,26 +248,36 @@ public bool InvalidateToken()
...
+            _commandFactory.Create<CommandRequest_PlexSyncWatched>(c => c.User = user).Save();
+
         return APIStatus.OK();
     }
 
     [Authorize("admin")]
-    [HttpGet("sync/{id}")]
-    public ActionResult SyncForUser(int uid)
+    [HttpGet("sync/{userID}")]
+    public ActionResult SyncForUser([FromQuery] int userID)
```
Cazzar, line 273: You're changing the way this is working - is this a good idea at all? Users might be expecting a certain behaviour

Shoko.Server/API/v3/Controllers/PlexController.cs
```diff
@@ -0,0 +1,201 @@
...
+{
+    private static string AlreadyAuthenticated = "User has already authenticated with plex yet. Invalidate the current token before re-authenticating.";
+
+    private static string NotAuthenticated = "User has not authenticated with plex yet. Authenticate first before retrying.";
+
+    private static string NoSelectedServer = "A server has not been selected yet. Select a server first.";
+
+    private static string InvalidServerSelection = "Invalid server selection.";
+
+    private static string InvalidLibrarySelection = "Selected libraries are invalid.";
```
Cazzar, line 26: Use `const` for these, prevent overwriting them

Shoko.Server/API/v3/Controllers/PlexController.cs
```diff
@@ -0,0 +1,201 @@
...
+        ServerSettings.Instance.Plex.Libraries = libraries?.Select(s => s.ID).ToList() ?? new();
+        ServerSettings.Instance.SaveSettings();
+
+        return directories
+            .Select(directory => new PlexLibrary(directory, currentServer, directorySet.Contains(directory.Key)))
+            .ToList();
+    }
+
+    [NonAction]
+    public PlexHelper GetHelperForUser()
```
Cazzar, line 198:
```suggestion
    [MethodImpl(MethodImplOptions.AggressiveInlining)]
    public PlexHelper GetHelperForUser()
```

Shoko.Server/API/v3/Controllers/UserController.cs
```diff
@@ -156,6 +165,20 @@ public ActionResult ChangePasswordForCurrentUser([FromBody] User.Input.ChangePas
...
+    /// <summary>
+    /// Sync the watch states against the selected plex server for the current user.
+    /// </summary>
+    /// <returns></returns>
+    [HttpGet("Current/Plex/Sync")]
+    public ActionResult SyncPlexForCurrentUser()
+    {
+        var user = HttpContext.GetUser();
+
+        CommandFactory.Create<CommandRequest_PlexSyncWatched>(c => c.User = user).Save();
```
Cazzar, line 177: inline the `user` variable

Cazzar (changes requested):

Shoko.Server/API/v3/Controllers/PlexController.cs
revam: Will update it when i get home. 👍

revam:
> "Fixed the v2 endpoints" What was wrong with them?

The incorrect mapping of the id/uid and the sync for user by id endpoint syncing for all the users instead of just the selected user.

Shoko.Server/API/v2/Modules/PlexWebhook.cs
revam: So it seems like it automatically removes the BOM if it's not needed.

Shoko.Server/API/v2/Modules/PlexWebhook.cs
revam: I can omit the changes I made to the v2 endpoints if you don't then to change.

Shoko.Server/API/v3/Controllers/PlexController.cs
```diff
@@ -0,0 +1,203 @@
...
+        var helper = GetHelperForUser();
+
+        return helper.IsAuthenticated;
+    }
+
+    /// <summary>
+    /// Invalidate and remove the current plex authentication token.
+    /// </summary>
+    /// <returns></returns>
+    [HttpPost("Auth/InvalidateToken")]
```
Cazzar, line 60: HTTP `DELETE`

Shoko.Server/API/v3/Controllers/PlexController.cs
```diff
@@ -0,0 +1,203 @@
...
+    public ActionResult<PlexServer> CurrentDevice()
+    {
+        var helper = GetHelperForUser();
+
+        if (!helper.IsAuthenticated)
+            return BadRequest(NotAuthenticated);
+
+        var currentServer = helper.ServerCache;
+        if (currentServer == null)
+            return BadRequest(NoSelectedServer);
```
Cazzar, line 106: Should an "unlinked" be a bad request? I would expect this to be a http 204

Shoko.Server/API/v3/Controllers/PlexController.cs
```diff
@@ -0,0 +1,203 @@
...
+
+        if (!helper.IsAuthenticated)
+            return BadRequest(NotAuthenticated);
+
+        var previousServer = helper.ServerCache;
+        var currentServer = helper.GetPlexServers()
+            .FirstOrDefault(ser => ser.ClientIdentifier == server.ID);
+
+        if (currentServer == null || !currentServer.Provides.Split(',').Contains("server"))
+            return BadRequest(InvalidServerSelection);
```
Cazzar, line 129: This is too much of the "is it a server" logic in the controller - can we move this out?

Shoko.Server/API/v3/Controllers/PlexController.cs
```diff
@@ -0,0 +1,203 @@
...
+
+        helper.UseServer(currentServer);
+
+        // Reset the selected libraries if we're switching server.
+        if (previousServer != null && currentServer.ClientIdentifier != previousServer.ClientIdentifier)
+        {
+            ServerSettings.Instance.Plex.Libraries = new();
+            ServerSettings.Instance.SaveSettings();
+        }
```
Cazzar, line 138: Move this to the `helper.UseServer()` call

The reason for this means if we add other places/methods to change the server it means the expected patterns are the same.

Cazzar (changes requested):

Shoko.Server/API/v3/Controllers/PlexController.cs
revam: That's the difference between loudly complaining (this 400) and silently failing (204). Personally I want it to loudly complain, but silently failing (and returning an empty) body is also fine. Will update it to use 204 instead.

Cazzar: Your still returning a 200 series response, aka a successful request, 400 indicates something was wrongly requested on the client & using 204 or “No content” is still distinct enough if you want to determine based off the return code

Shoko.Server/API/v3/Controllers/PlexController.cs
revam: It was doing the same in the `helper.UseServer` with the exception of it **silently failing** and not informing the user that the server **were not selected** because it was an invalid device to begin with.

If we can either change the helper to return the currently selected server (so we can check if it was updated) or throw inside the helper instead of silently returning then I'm fine with removing the check here, but I still want it to complain if the selected device is invalid as a form of user input validation and also just to inform the user that they tried to select an invalid device.

Cazzar: This is a matter for consistency sake, and the problem if we want to look. At exposing the logic elsewhere we have to duplicate the logic, IMO we shouldn’t really have much decision making in controllers but defer that decision making to central classes that are agnostic of the request method.

revam:
A new PR with more internal changes is incoming. Closing this for now.

PR #1024, opened 2022-12-08
refactor: update image commands

revam:
Modernize the image commands. A second PR to add staff data and staff images
are incoming.

 ### Changes in this commit/PR

- Modernized and simplified the image commands (by my standards if that is
  anything to go by man_shrugging).

- Tried to reduce duplicate code where possible, without rewriting the whole
  system.

- **Updated Shoko.Commons** to modify the get image format helper function
  used in this PR. The master branch for Shoko.Commons needs to be moved
  accordingly afterwards.

### DIFF
--- a/Shoko.Server/Commands/Import/CommandRequest_DownloadAniDBImages.cs
+++ b/Shoko.Server/Commands/Import/CommandRequest_DownloadAniDBImages.cs
@@ -1,23 +1,16 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Globalization;
-using System.IO;
 using System.Linq;
 using System.Net;
-using System.Threading;
 using System.Xml;
 using Microsoft.Extensions.Logging;
 using Shoko.Commons.Properties;
 using Shoko.Commons.Queue;
-using Shoko.Commons.Utils;
-using Shoko.Models.Enums;
 using Shoko.Models.Queue;
 using Shoko.Models.Server;
 using Shoko.Server.Commands.Attributes;
 using Shoko.Server.Commands.Generic;
-using Shoko.Server.Extensions;
 using Shoko.Server.ImageDownload;
-using Shoko.Server.Providers.AniDB;
 using Shoko.Server.Providers.AniDB.Interfaces;
 using Shoko.Server.Repositories;
 using Shoko.Server.Server;
@@ -28,30 +21,37 @@ namespace Shoko.Server.Commands.Import;
 [Command(CommandRequestType.DownloadAniDBImages)]
 public class CommandRequest_DownloadAniDBImages : CommandRequestImplementation
 {
+    private const string DownloadImage = "Downloading Image {0}: {1}";
+
+    private const string FailedToDownloadNoID = "Images failed to download: Can\'t find {EntityType} with ID: {AnimeID}";
+
     private readonly IUDPConnectionHandler _handler;
+
     private readonly ISettingsProvider _settingsProvider;
+
     public int AnimeID { get; set; }
+
     public bool ForceDownload { get; set; }
 
     public override CommandRequestPriority DefaultPriority => CommandRequestPriority.Priority1;
 
     public override QueueStateStruct PrettyDescription => new()
     {
-        message = "Downloading Image {0}: {1}",
+        message = DownloadImage,
         queueState = QueueStateEnum.DownloadImage,
         extraParams = new[] { Resources.Command_ValidateAllImages_AniDBPosters, AnimeID.ToString() }
     };
 
     public QueueStateStruct PrettyDescriptionCharacters => new()
     {
-        message = "Downloading Image {0}: {1}",
+        message = DownloadImage,
         queueState = QueueStateEnum.DownloadImage,
         extraParams = new[] { Resources.Command_ValidateAllImages_AniDBCharacters, AnimeID.ToString() }
     };
 
     public QueueStateStruct PrettyDescriptionCreators => new()
     {
-        message = "Downloading Image {0}: {1}",
+        message = DownloadImage,
         queueState = QueueStateEnum.DownloadImage,
         extraParams = new[] { Resources.Command_ValidateAllImages_AniDBSeiyuus, AnimeID.ToString() }
     };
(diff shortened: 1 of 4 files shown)


revam:
Feel free to point out my mistakes (if any).

Shoko.Server/ImageDownload/ImageDownloadRequest.cs
```diff
@@ -1,17 +1,129 @@
...
 public class ImageDownloadRequest
 {
-    public ImageEntityType ImageType { get; set; }
-    public object ImageData { get; set; }
-    public bool ForceDownload { get; set; }
 
-    public ImageDownloadRequest(ImageEntityType imageType, object data, bool forceDownload)
+    private object ImageData { get; }
+
+    public bool ForceDownload { get; }
```
revam, line 20: We can probably remove this since it's always the same as the `ForceDownload` option in the command(s).

revam: Will update the code when i get back home.

Shoko.Server/ImageDownload/ImageDownloadRequest.cs
```diff
@@ -1,17 +1,126 @@
...
+            // Ignore all certificate failures.
+            ServicePointManager.Expect100Continue = true;
+            ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12;
+            ServicePointManager.ServerCertificateValidationCallback = delegate { return true; };
+
+            // Download the image.
+            using (var client = new HttpClient())
+            {
+                // Download the image data.
+                client.DefaultRequestHeaders.Add("user-agent", "JMM");
```
revam, line 85: Should we update or keep the "JMM" user agent?

Shoko.Server/ImageDownload/ImageDownloadRequest.cs
```diff
@@ -1,17 +1,126 @@
...
+
+                // Get the file extension.
+                var extension = Misc.GetImageFormat(bytes) switch
+                {
+                    string imageFormat => "." + imageFormat,
+                    _ => throw new WebException("The image download stream returned an invalid image"),
+                };
+
+                // Change the file extension.
+                tempPath = Path.ChangeExtension(tempPath, extension);
```
revam, line 102: I have no idea why we change the file extension _on the temporary file path_ while also **not** changing it on the final file path. To be clear, this was the behavior from before this PR, and I didn't change it to this, only copy it. 

It should probably be removed as it serves no purpose (in my eyes at least :shrug:). Checking the image format should still be done — we only need to check if it's not `null` (now after the last two commits) to check if it's a valid image — but changing the extension on the temp. file is unneeded.

Alternately we can just keep it. I won't make the final call. Either way is fine. :upside_down_face:

Shoko.Server/Commands/Import/CommandRequest_DownloadAniDBImages.cs
```diff
@@ -28,204 +25,180 @@ namespace Shoko.Server.Commands.Import;
...
-
-                        foreach (var creator in creators)
-                        {
-                            downloadUrls.Add(string.Format(_handler.ImageServerUrl, creator.PicName));
-                            fileNames.Add(creator.GetPosterPath());
-                        }
-
-                        ShokoService.CmdProcessorImages.QueueState = PrettyDescriptionCreators;
-                        break;
+                    Logger.LogWarning(FailedToDownloadNoID, "characters for anime", AnimeID);
```
da3dsoul, line 100: No. Don't move new things to resources. We are moving away from them. Also...especially don't mix and match hardcoded language strings and resources...
Please fix all occurrences

Shoko.Server/Commands/Import/CommandRequest_DownloadAniDBImages.cs
```diff
@@ -235,90 +208,30 @@ select RepoFactory.AniDB_Seiyuu.GetBySeiyuuID(xref2.SeiyuuID))
...
-        {
-            if (count + 1 >= maxretry)
-            {
-                throw;
-            }
-
-            Thread.Sleep(500);
-            RecursivelyRetryDownload(downloadURL, ref tempFilePath, count + 1, maxretry);
-        }
+        // Find out which directories needs to be created and which already
```
da3dsoul, line 220: Completely unnecessary. Directory.CreateDirectory will make all needed parts. https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.createdirectory?view=net-7.0

```cs
var dirPath = Path.GetDirectoryName(path);
Directory.CreateDirectory(dirPath);
```

Shoko.Server/ImageDownload/ImageDownloadRequest.cs
da3dsoul: I agree, it should be done on the final path

Shoko.Server/ImageDownload/ImageDownloadRequest.cs
da3dsoul: Keep. AniDB whitelists the user agent for some things, and I don't want to fuck with it

da3dsoul (changes requested):

Shoko.Server/ImageDownload/ImageDownloadRequest.cs
revam: Will try to change it (so it also updates the database entry)

Shoko.Server/Commands/Import/CommandRequest_DownloadAniDBImages.cs
revam: I'm not using resources here though. This is a `const string` message template to prevent the same line being repeated over and over in the code... but now is the usage of the `const string` repeated over and over instead. The upside is that it is now one line change to update all the messages.

revam:
I'm converting this to a draft for now. After some more thinking did I remember that the image paths stored in `AnimeStaff` table is always empty for the staff members that are not VA — because the paths are not included in the HTTP XML files and we haven't downloaded it from the UDP API yet.

I'm keeping this PR open since I intend to add the udp commands to download the missing data and also do some refactoring on the `Anidb_Seiyuu` table. Mainly renaming it to `AniDB_Creator` since it will be not just VAs anymore and also adding the creator type ('person', 'collaboration' or 'company' iirc). It will be easy to migrate for the existing VAs since they're always of type 'person'. The creator type is good to know mainly to distinguish between the studios (of type 'company') and staff members — both VAs and non-VAs — (of type 'person' or 'collaboration').

da3dsoul:
Sounds good

revam:
As I said (in the updated) description, a second PR is coming for the staff data. Meanwhile I'll use this PR for the image refactor.

Shoko.Server/ImageDownload/ImageDownloadRequest.cs
revam: Tl;dr: that's hard, since we're not storing every path in the database, and not all paths are stored the same way.

revam:
Fixed the merge conflicts from the settings rewrite.

da3dsoul:
I don't see any glaring issues. As long as it's tested, it's fine

PR #1026, opened 2022-12-22
refactor: update tags

revam:
This commit/PR changes some of the data we store about tags, update how we handle tag renaming and exposes some new data in the api. I've tested that it works with a SQLite database, but have not tried for MySQL or SQL Server.

### Changes in this PR

- Updated the Anidb_Tag and Anidb_Anime_Tag tables to better align with the data present in the anime http xml dumps. This is mostly for adding some missing-but-useful information like the verified status of the tag, parent tag id for parsing the tag hierarchy, the last updated timestamp for comparing if we want to update the tag details and to prevent older tag details from overwriting newer details. These changes also involve converting int booleans to booleans in code, moving the local spoiler boolean from the tag to the anime↔tag cross-reference where it belongs, and removes some old and now unused fields from both tables.

- Added a name override for some anidb tags, mainly "original work" → "source material" and "new" → "original work" for now. These overrides are used both internally and externally to be backwards compatible with v1 and v2, but are stored separately in the database so we can still use the source name when updating the tags.

- Tweaked (and simplified) the tag filter. Removed the logic to replace and rename tags, but kept the logic for adding the "original work" tag if no source material tag is set.

- Database migrations and a database fix to fix up the tags after the migration.

- Updated the v1 and v2 api so they only display verified tags, since unverified tags are still pending approval by a moderator on anidb. If a tag later becomes verified then it will be reflected in the apis once we've downloaded an anime http xml dump containing the updated tags.

- Updated the v3 api to only show verified tags by default, but have the ability to also show unverified tags. Also updated the v3 api model to display the new data for the anidb tags.

- Upated the tests for the tag filter.

### DIFF
--- a/Shoko.Server/Databases/MySQL.cs
+++ b/Shoko.Server/Databases/MySQL.cs
@@ -706,7 +706,7 @@ public class MySQL : BaseDatabase<MySqlConnection>, IDatabase
         new(106, 2, "UPDATE VideoLocal v INNER JOIN CrossRef_File_Episode CRFE on v.Hash = CRFE.Hash SET DateTimeImported = DateTimeCreated;"),
         new(107, 1, "ALTER TABLE AniDB_Tag ADD Verified integer NOT NULL DEFAULT 0;"),
         new(107, 2, "ALTER TABLE AniDB_Tag ADD ParentTagID integer DEFAULT NULL;"),
-        new(107, 3, "ALTER TABLE AniDB_Tag ADD TagNameOverride text DEFAULT NULL;"),
+        new(107, 3, "ALTER TABLE AniDB_Tag ADD TagNameOverride varchar(150) DEFAULT NULL;"),
         new(107, 4, "ALTER TABLE AniDB_Tag ADD LastUpdated datetime NOT NULL DEFAULT '1970-01-01 00:00:00';"),
         new(107, 5, "ALTER TABLE AniDB_Tag DROP COLUMN Spoiler;"),
         new(107, 6, "ALTER TABLE AniDB_Tag DROP COLUMN LocalSpoiler;"),
--- a/Shoko.Server/Databases/MySQL.cs
+++ b/Shoko.Server/Databases/MySQL.cs
@@ -19,7 +19,7 @@ namespace Shoko.Server.Databases;
 public class MySQL : BaseDatabase<MySqlConnection>, IDatabase
 {
     public string Name { get; } = "MySQL";
-    public int RequiredVersion { get; } = 106;
+    public int RequiredVersion { get; } = 107;
 
 
     private List<DatabaseCommand> createVersionTable = new()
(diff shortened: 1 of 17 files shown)


Shoko.Server/Mappings/AniDB_TagMap.cs
```diff
@@ -11,12 +11,13 @@ public AniDB_TagMap()
...
 
-        Map(x => x.GlobalSpoiler).Not.Nullable();
-        Map(x => x.LocalSpoiler).Not.Nullable();
-        Map(x => x.Spoiler).Not.Nullable();
-        Map(x => x.TagCount).Not.Nullable();
-        Map(x => x.TagDescription).Not.Nullable().CustomType("StringClob");
         Map(x => x.TagID).Not.Nullable();
-        Map(x => x.TagName).Not.Nullable();
+        Map(x => x.ParentTagID);
+        Map(x => x.TagNameSource).Column("TagName").Not.Nullable();
```
da3dsoul, line 16: I would name it TagName and TagNameOverride. The Override part is a modifier.

Shoko.Server/Providers/AniDB/HTTP/HttpAnimeParser.cs
```diff
@@ -473,20 +473,26 @@ private static ResponseTag ParseTag(int animeID, XmlNode node)
...
             AnimeID = animeID,
             TagID = tagID,
+            ParentTagID = parentTagID > 0 ? parentTagID : null,
             TagName = tagName,
             TagDescription = tagDescription,
             Weight = weight,
+            Verified = verified,
             LocalSpoiler = lsp,
             GlobalSpoiler = gsp,
-            Spoiler = lsp || gsp
```
da3dsoul, line 489: Are you just removing the general spoiler field?

revam: It was never used.

revam: Also, if it's a global spoiler, then it's a spoiler. The local spoiler is only for when the tag is used on that particular anime.

Shoko.Server/Mappings/AniDB_TagMap.cs
revam: I didn't want to update all the existing occurrences of `TagName` in the code to a new getter so instead did i create a new property for the source name and mapped it to the database field, and then made the original property a getter (only) that gets the override or the source if no override is available. This way we don't need to update 34 (iirc) occurrences of `TagName`.

Shoko.Server/Providers/AniDB/HTTP/HttpAnimeParser.cs
revam: > It was never used.

Outside the setters i mean.

da3dsoul: That's fine. I was just asking for clarification

Shoko.Server/Mappings/AniDB_TagMap.cs
da3dsoul: I see. That is acceptable

Shoko.Server/Databases/SQLServer.cs
```diff
@@ -649,7 +649,17 @@ public override bool HasVersionsTable()
         new DatabaseCommand(97, 1, "CREATE INDEX IX_AniDB_Episode_EpisodeType ON AniDB_Episode(EpisodeType);"),
         new DatabaseCommand(98, 1, "ALTER TABLE AniDB_Episode_Title ALTER COLUMN Title nvarchar(max) NOT NULL;"),
         new DatabaseCommand(99, 1, "ALTER TABLE VideoLocal ADD DateTimeImported datetime DEFAULT NULL;"),
-        new DatabaseCommand(99, 2, "UPDATE v SET DateTimeImported = DateTimeCreated FROM VideoLocal v INNER JOIN CrossRef_File_Episode CRFE on v.Hash = CRFE.Hash;")
+        new DatabaseCommand(99, 2, "UPDATE v SET DateTimeImported = DateTimeCreated FROM VideoLocal v INNER JOIN CrossRef_File_Episode CRFE on v.Hash = CRFE.Hash;"),
+        new DatabaseCommand(100, 1, "ALTER TABLE AniDB_Tag ADD Verified integer NOT NULL DEFAULT 0;"),
+        new DatabaseCommand(100, 2, "ALTER TABLE AniDB_Tag ADD ParentTagID integer DEFAULT NULL;"),
+        new DatabaseCommand(100, 3, "ALTER TABLE AniDB_Tag ADD TagNameOverride text DEFAULT NULL;"),
```
da3dsoul, line 655: Make TagNameOverride nvarchar(150) null to match TagName

Shoko.Server/Databases/MySQL.cs
```diff
@@ -703,7 +703,17 @@ public class MySQL : BaseDatabase<MySqlConnection>, IDatabase
         new(104, 1, "ALTER TABLE AniDB_Episode ADD INDEX IX_AniDB_Episode_EpisodeType (EpisodeType);"),
         new(105, 1, "ALTER TABLE AniDB_Episode_Title MODIFY Title TEXT NOT NULL"),
         new(106, 1, "ALTER TABLE VideoLocal ADD DateTimeImported datetime DEFAULT NULL;"),
-        new(106, 2, "UPDATE VideoLocal v INNER JOIN CrossRef_File_Episode CRFE on v.Hash = CRFE.Hash SET DateTimeImported = DateTimeCreated;")
+        new(106, 2, "UPDATE VideoLocal v INNER JOIN CrossRef_File_Episode CRFE on v.Hash = CRFE.Hash SET DateTimeImported = DateTimeCreated;"),
+        new(107, 1, "ALTER TABLE AniDB_Tag ADD Verified integer NOT NULL DEFAULT 0;"),
+        new(107, 2, "ALTER TABLE AniDB_Tag ADD ParentTagID integer DEFAULT NULL;"),
+        new(107, 3, "ALTER TABLE AniDB_Tag ADD TagNameOverride text DEFAULT NULL;"),
```
da3dsoul, line 709: varchar(150) to match TagName

da3dsoul:
Ok, after those two changes, looks good

revam:
Also updated the tests and added back some of the custom behavior that I broke.

Shoko.Server/Utilities/TagFilter.cs
```diff
@@ -422,6 +422,15 @@ public static bool IsTagBlackListed(string tag, Filter flags)
     {
         tag = tag.Trim().ToLowerInvariant();
         var inverted = flags.HasFlag(Filter.Invert);
+
+        // Always remove the "original work" tag. It will be added back if it's
+        // not supposed to be filtered out and there are no other source
+        // material tag.
```
revam, line 428: Should be "tags" on this line or "is" on the line above, but _meh_

revam:
Fixed the tests (again) and ensured we always have lower case letters

Is this ready for merge now? @da3dsoul

da3dsoul:
Do we want lowercase tag names?

revam:
I made it so it ensures the tag name is always lower case because the tag filter expects the values to be lower case (as indicated by the hash sets. Alternatively I can change it so it changes it to lower case the value right before it's used in the tag filter. Actually, I'll do that instead. I'll also tweak the tests to test that it does not matter if the input tags names are upper cased, lower cased or a mix of the two.

da3dsoul:
I'll review it again later. I had adjusted the case stuff intentionally to improve the speed. I'll also add unit tests to take casing into account

da3dsoul:
Ok, looks good

PR #1036, opened 2023-01-23
Refactor episode updates

revam:
This refactor is mainly to prevent episodes moving between different series from throwing errors. I changed it so the code tracks when an episode was last updated _at the remote_ (or in the entry stored in the cache), and made it so it conditionally updates the episode record and/or updates the episode title records accordingly.

Also, for the best compatibility with updating existing episode entries did I add a database migration to reset the last updated timestamp on all existing anidb episode records. This won't affect anything other than it will force the episode records to be updated on the first anime refresh/update — be it from the cache or online — after this PR has been merged.

### DIFF
--- a/Shoko.Server/Databases/MySQL.cs
+++ b/Shoko.Server/Databases/MySQL.cs
@@ -19,7 +19,7 @@ namespace Shoko.Server.Databases;
 public class MySQL : BaseDatabase<MySqlConnection>, IDatabase
 {
     public string Name { get; } = "MySQL";
-    public int RequiredVersion { get; } = 112;
+    public int RequiredVersion { get; } = 113;
 
 
     private List<DatabaseCommand> createVersionTable = new()
@@ -727,6 +727,7 @@ public class MySQL : BaseDatabase<MySqlConnection>, IDatabase
         new(112, 2, "ALTER TABLE AniDB_Anime DROP COLUMN AnimePlanetID;"),
         new(112, 3, "ALTER TABLE AniDB_Anime DROP COLUMN AnimeNfo;"),
         new(112, 4, "ALTER TABLE AniDB_Anime ADD LainID INT NULL"),
+        new(113, 1, DatabaseFixes.FixEpisodeDateTimeUpdated),
     };
 
     private DatabaseCommand linuxTableVersionsFix = new("RENAME TABLE versions TO Versions;");
--- a/Shoko.Server/Databases/DatabaseFixes.cs
+++ b/Shoko.Server/Databases/DatabaseFixes.cs
@@ -876,4 +876,116 @@ public class DatabaseFixes
 
         logger.Info($"Done updating anidb tags for {animeList.Count} anidb anime entries.");
     }
+
+    public static void FixEpisodeDateTimeUpdated()
+    {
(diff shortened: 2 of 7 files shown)


Shoko.Server/Databases/MySQL.cs
```diff
@@ -720,6 +720,7 @@ public class MySQL : BaseDatabase<MySqlConnection>, IDatabase
         new(110, 1, "CREATE TABLE `AniDB_FileUpdate` ( `AniDB_FileUpdateID` INT NOT NULL AUTO_INCREMENT, `FileSize` BIGINT NOT NULL, `Hash` varchar(50) NOT NULL, `HasResponse` BIT NOT NULL, `UpdatedAt` datetime NOT NULL, PRIMARY KEY (`AniDB_FileUpdateID`) );"),
         new(110, 2, "ALTER TABLE `AniDB_FileUpdate` ADD INDEX `IX_AniDB_FileUpdate` (`FileSize` ASC, `Hash` ASC) ;"),
         new(110, 3, DatabaseFixes.MigrateAniDB_FileUpdates),
+        new(111, 1, "Update AniDB_Episode SET DateTimeUpdated = '1970-01-01 00:00:00';"),
```
Cazzar, line 723: Why are we resetting this time for all the episodes?

revam: Yes, I could have worded it better.

revam: The migration can be removed if you or @da3dsoul think it should be removed. (That's why I made it a separate commit, in case someone else deemed it unnecessary.)

da3dsoul: I said this before, but why use an individual date for episodes? We don't get a date from AniDB, and the date will only be updated when the anime is. Just use the AniDB_AnimeUpdate date

revam: And I said it before too, we do get individual dates for episodes for when they were last updated. They are stored in the xml files and I mapped them to be used in this PR.

Also, said date can be used to check which of the episode entries — comparing the record stored in the database and the entry in the xml file (which can be loaded from a stall/older cache if banned) — is the freshest, and conditionally update the episode records (e.g. set the overview, rating, length, etc., move the episode to another anime, etc.) if the xml file contains a newer representation than the one stored in the database.

da3dsoul: I forgot

revam: So, should we remove the database migration? I will reset the dates on my own install either way, but I thought it would be more helpful for others if they was just reset as part of a migration.

revam: > So, should we remove the database migration?

Still waiting for an answer. If nobody responds then is it safe to assume the answer is "no, it's fine to let it be" and I can just merge this now?

da3dsoul: I would make a helper function that pulls it from the xml

revam:
I'll get to it... when I have time and energy to do it. 'It' being "updating the database migration."

PR #1039, opened 2023-01-29
Fix: PathEndsWithQuery parsing path as intended

fearnlj01:
`/api/v3/File/PathEndsWith` was receiving the path as a null paramater when passed as a query rather than path. (Working fine when passed as a path)

I don't fully understand my own change, but I couldn't find any other instances of `[FromRoute, FromQuery]` being used so I tweaked & tested.

This seems to fix the aforementioned issue, at least on my equipment.


da3dsoul:
To be fair, I don't think we have any other instances that support both, so that would be why you didn't see it elsewhere.

revam:
My bad

PR #1042, opened 2023-02-28
Adds additional censorship when logging

fearnlj01:
The first commit uses regex to avoid logging the auth key when utilising AVDump. As @Cazzar has noted in #support (https://discord.com/channels/96234011612958720/96235819366371328/1079930915054170172), it probably makes more sense to just simply log the filename here instead, however I daren't make such a change myself.

The second commit just additionally censors settings (when dumping) that have names ending in `key` & `token` (rather than just `password`). Admittedly, my inclusion of `token` in this change is a bit of a guess since I don't use Plex nor Trakt, however I'd imagine that users might expect these credentials to be kept secret if sharing the logfile?

### DIFF
--- a/Shoko.Server/Utilities/AVDumpHelper.cs
+++ b/Shoko.Server/Utilities/AVDumpHelper.cs
@@ -1,6 +1,7 @@
-﻿using System;
+using System;
 using System.Diagnostics;
 using System.IO;
+using System.Text.RegularExpressions;
 using NLog;
 using SharpCompress.Common;
 using SharpCompress.Readers;
@@ -194,7 +195,9 @@ public static class AVDumpHelper
 
             var filenameArgs = GetFilenameAndArgsForOS(file);
 
-            logger.Info($"Dumping File with AVDump: {filenameArgs.Item1} {filenameArgs.Item2}");
+            // Censor the --Auth section for logging, replacing only the password
+            var censoredArgs = Regex.Replace(filenameArgs.Item2, "(--Auth=([^:]+):)([^ ]+)", "$1***HIDDEN***");
+            logger.Info($"Dumping File with AVDump: {filenameArgs.Item1} {censoredArgs}");
 
             var pProcess = new Process
             {
--- a/Shoko.Server/Utilities/AVDumpHelper.cs
+++ b/Shoko.Server/Utilities/AVDumpHelper.cs
@@ -206,7 +206,7 @@ public static class AVDumpHelper
         }
         catch (Exception ex)
         {
-            logger.Error($"An error occurred while AVDumping the file \"file\":\n{ex}");
+            logger.Error($"An error occurred while AVDumping the file \"{file}\":\n{ex}");
             return $"An error occurred while AVDumping the file:\n{ex}";
         }
     }
--- a/Shoko.Server/Utilities/AVDumpHelper.cs
+++ b/Shoko.Server/Utilities/AVDumpHelper.cs
@@ -197,8 +197,7 @@ public static class AVDumpHelper
             // Logs only the last argument to AVDump, the filename
             logger.Info($"Dumping File with AVDump: \"{startInfo.ArgumentList[^1]}\"");
 
-            using var pProcess = new Process();
-            pProcess.StartInfo = startInfo;
+            using var pProcess = new Process { StartInfo = startInfo };
             pProcess.Start();
             var strOutput = pProcess.StandardOutput.ReadToEnd();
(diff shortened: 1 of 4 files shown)


Cazzar:
I would more say log the filename as regex, while I can be a fan of it can be hard to maintain and make sure isn't working as well, plus worst case we can reconstruct the parameters .

As for the other reasons I don't think there's much of a need to do those specifically, the plex token is actually never set (nor logged as it's done as a HTTP header to plex) (The setting is actually just a thing that existed for a short period before it was moved to the database)

as for the simple matter of censoring, at a thought maybe instead of a string-match would it be better to make it attribute driven (cc @da3dsoul thoughts?) for example 

```csharp
[LogCensor] public string SomeKey {get; set;}
```

fearnlj01:
New commits refactor the AVDump process somewhat to use a list of arguments rather than a single argument string. Regex is completely avoided this way!

Tested for function on Windows and it seems to play nice however I haven't yet had the chance to test on any other OS. I will try to get a VM spun up later today to test this on Ubuntu.

If we wanted to log the 'full' command line it should be equally trivial to change with this refactor.

As far as censorship for the settings file dump is concerned, I could remove this from this PR (although the AVDump key would still be dumped if so) - just let me know.

fearnlj01:
> as for the simple matter of censoring, at a thought maybe instead of a string-match would it be better to make it attribute driven (cc @da3dsoul thoughts?) for example
> 
> ```cs
> [LogCensor] public string SomeKey {get; set;}
> ```

Since this was easier than I thought to do... I've got an implementation for this ready as well - Will await a further thumbs up to making the censorship attribute driven before adding this to the PR. Hopefully storing the new attribute in `/Shoko.Server/Settings/Attributes/LogCensorAttribute.cs` is agreeable?

Shoko.Server/Utilities/AVDumpHelper.cs
```diff
@@ -192,56 +192,54 @@ public static string DumpFile(string file)
...
-                    WindowStyle = ProcessWindowStyle.Hidden,
-                    RedirectStandardOutput = true,
-                    CreateNoWindow = true
-                }
-            };
+            // Logs only the last argument to AVDump, the filename
+            logger.Info($"Dumping File with AVDump: \"{startInfo.ArgumentList[^1]}\"");
 
+            using var pProcess = new Process();
+            pProcess.StartInfo = startInfo;
```
Cazzar, line 201:
```suggestion
            using var pProcess = new Process { StartInfo = startInfo };
```

Uses object constructor to set the start info at the same time as creation.

da3dsoul:
I think nlog just has a filtering system, and we could add values to it when needed, that could prevent keys from showing up in logged URLs and stuff

fearnlj01:
Swapped out the filtering with the latest two commits to filter out (via nlog config) any error message containing `password`, `token` and `key` when the log message starts with `Settings.`. Restarted a few times and the dumped settings look good to me.

For the sake of including the links... The nlog documentation pages I used for this was ['Filtering log messages'](https://github.com/nlog/nlog/wiki/Filtering-log-messages) & ['When filter'](https://github.com/nlog/nlog/wiki/When-filter).

Haven't added the filters to the console output, only the file output - not sure if it'd be best to filter both?

Shoko.Server/nlog.config
```diff
@@ -13,7 +13,13 @@
     <logger name="Shoko.Server.API.Authentication.CustomAuthHandler" maxlevel="Info"
             final="true"/> <!-- Auth log spam (blackhole) -->
 
-    <logger name="*" minlevel="Info" writeTo="file"/>
+    <logger name="*" minlevel="Info" writeTo="file">
```
da3dsoul, line 16: Have you tested this? I'm not sure how some nlog.config stuff works with our code overrides in https://github.com/ShokoAnime/ShokoServer/blob/73542e3fb10c64b2f05e76c9c6e93c2a70e373af/Shoko.Server/Utilities/Utils.cs#L75

fearnlj01: I've given it a test and it works for file based logging (in line with previous comments)...

Realistically, given that troubleshooting can be performed using `server-settings.json`, I'm not entirely sure why I only bothered implementing this for the file target (especially if I was proposing removing the censorship added the original way).

I've now implemented the previous changes in `Utils.cs` with the above filter rules applying to the file log target, and additionally the SignalR & console log targets. Will get this pushed onto the branch shortly.

da3dsoul:
Looks good

PR #1043, opened 2023-03-02
Fix: Pass Tag (not Version) to GetUrlAndUpdate

fearnlj01:
Tried updating to latest WebUI and was receiving 404's when trying to update.

Did a bit of digging and it looks like `GetUrlAndUpdate` expects the tag name to be passed to it, rather than the version. I've tested my changes and it "works on my machine™".


revam (approved):
1. My bad, forgot about that
2. I will trust you that it "works on my machine" for you. (Since it should.)

PR #1051, opened 2023-03-30
Update Dockerfiles

krbrs:
Remove obsolete webui volume as it's in the already exported home directory - if I'm not wrong.

This should solve:

- QNAP Container Station keeps bugging to map it to a volume or share
- Portainer creating an empty volume


revam:
Personally i don't mind keeping or removing the volumes.

@da3dsoul @Cazzar thoughts on removing it?

Cazzar:
There’s your answer @revam

PR #1052, opened 2023-04-01
Fix AniDB languages not being updated properly

Mik1ll:
Noticed this while testing my renamer, when a file has either no dub or no sub AniDB returns "none" in the UDP response for the track languages. This is being incorrectly saved to the DB and will return as "Unknown" in the renamer interface. Fixed by ignoring "none" and returning an empty list.

The languages pulled fron AniDB were not syncing with the DB correctly when using update file info. CreateLanguages in CommandRequest_GetFile.cs did not create a transaction and commit it. Observed duplicate language entries being created, and none deleted upon multiple updates.

The condition for updating the languages was if the returned count of languages from AniDB was > 0, changed to only check if the response was null, tracks can be deleted on AniDB and should be deleted in DB.

### DIFF
--- a/Shoko.Server/Commands/AniDB/CommandRequest_GetFile.cs
+++ b/Shoko.Server/Commands/AniDB/CommandRequest_GetFile.cs
@@ -138,34 +138,36 @@ public class CommandRequest_GetFile : CommandRequestImplementation
     public void CreateLanguages(ResponseGetFile response)
     {
         using var session = DatabaseFactory.SessionFactory.OpenSession();
-        if ((response?.AudioLanguages?.Count ?? 0) >
-            0) //Only create relations if the origin of the data if from Raw (WebService/AniDB)
+        using var trans = session.BeginTransaction();
+        // Only update languages if we got a response
+        if (response?.AudioLanguages is not null)
         {
-            // Delete old if changed
-            var fileLanguages = RepoFactory.CrossRef_Languages_AniDB_File.GetByFileID(response.FileID);
-            RepoFactory.CrossRef_Languages_AniDB_File.DeleteWithOpenTransaction(session, fileLanguages);
+            // Delete old
+            var toDelete = RepoFactory.CrossRef_Languages_AniDB_File.GetByFileID(response.FileID);
+            RepoFactory.CrossRef_Languages_AniDB_File.DeleteWithOpenTransaction(session, toDelete);
 
+            // Save new
             var toSave = response.AudioLanguages.Select(language => language.Trim().ToLower())
                 .Where(lang => lang.Length > 0)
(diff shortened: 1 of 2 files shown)


Shoko.Server/Commands/AniDB/CommandRequest_GetFile.cs
```diff
@@ -138,34 +138,36 @@ protected override void Process()
     public void CreateLanguages(ResponseGetFile response)
     {
         using var session = DatabaseFactory.SessionFactory.OpenSession();
-        if ((response?.AudioLanguages?.Count ?? 0) >
-            0) //Only create relations if the origin of the data if from Raw (WebService/AniDB)
+        using var trans = session.BeginTransaction();
+        // Only update languages if we got a response
+        if (response?.AudioLanguages is not null)
```
revam, line 143: This will remove all existing languages even if the response doesn't contain any (new) languages (to replace the list with).

If this is the intended behavior then this change is fine, otherwise it would be wise to revert this line.

Shoko.Server/Commands/AniDB/CommandRequest_GetFile.cs
```diff
@@ -138,34 +138,36 @@ protected override void Process()
...
             var toSave = response.AudioLanguages.Select(language => language.Trim().ToLower())
                 .Where(lang => lang.Length > 0)
                 .Select(lang => new CrossRef_Languages_AniDB_File { LanguageName = lang, FileID = response.FileID })
                 .ToList();
-
             RepoFactory.CrossRef_Languages_AniDB_File.SaveWithOpenTransaction(session, toSave);
         }
 
-        if ((response?.SubtitleLanguages?.Count ?? 0) > 0)
+        if (response?.SubtitleLanguages is not null)
```
revam, line 157: Same as the other comment.

Shoko.Server/Providers/AniDB/UDP/Info/RequestGetFile.cs
```diff
@@ -222,10 +222,10 @@ protected override UDPResponse<ResponseGetFile> ParseResponse(UDPResponse<string
                     }
 
                     // audio languages
-                    var alangs = parts[10].Split(new[] { '\'' }, StringSplitOptions.RemoveEmptyEntries).ToList();
+                    var alangs = parts[10].Split(new[] { '\'' }, StringSplitOptions.RemoveEmptyEntries).Where(lang => lang != "none").ToList();
```
revam, line 225: Did you confirm that it may include "none" as a value in the list? Just asking to confirm.

revam (commented):
The changes looks okay, but can you look over the comments before we merge it?

revam:
I forgot to read the PR message, nvm me.

PR #1059, opened 2023-04-19
Resolves searches returning null

fearnlj01:
I've tested these changes using the `/api/v3/Series/AniDB/Search` endpoint and it seems to get things working.

Given that the initial values for `current.Distance` & `current.Index` were always null-coalesced to `int.MaxValue`, `result.Distance` & `index` would always be less and as such the initial `null` value was always returned as the search result.

### DIFF
--- a/Shoko.Server/Utilities/SeriesSearch.cs
+++ b/Shoko.Server/Utilities/SeriesSearch.cs
@@ -125,14 +125,14 @@ public static class SeriesSearch
                     if (fuzzy)
                     {
                         var result = DiceFuzzySearch(title, search, t);
-                        if (result.Index == -1 || result.Distance > (current?.Distance ?? int.MaxValue))
+                        if (result.Index == -1 || result.Distance >= (current?.Distance ?? int.MaxValue))
                             return current;
 
                         return result;
                     }
 
                     var index = title.IndexOf(search, StringComparison.OrdinalIgnoreCase);
-                    if (index == -1 || index > (current?.Index ?? int.MaxValue))
+                   if (index == -1 || index >= (current?.Index ?? int.MaxValue))
                         return current;
 
                     return new()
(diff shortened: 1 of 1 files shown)


Shoko.Server/Utilities/SeriesSearch.cs
```diff
@@ -125,14 +125,14 @@ private static ParallelQuery<SearchResult<T>> SearchCollection<T>(ParallelQuery<
                     if (fuzzy)
                     {
                         var result = DiceFuzzySearch(title, search, t);
-                        if (result.Index == -1 || result.Distance < (current?.Distance ?? int.MaxValue))
+                        if (result.Index == -1 || result.Distance > (current?.Distance ?? int.MaxValue))
```
revam, line 128:
```suggestion
                        if (result.Index == -1 || result.Distance >= (current?.Distance ?? int.MaxValue))
```

Shoko.Server/Utilities/SeriesSearch.cs
```diff
@@ -125,14 +125,14 @@ private static ParallelQuery<SearchResult<T>> SearchCollection<T>(ParallelQuery<
...
                             return current;
 
                         return result;
                     }
 
                     var index = title.IndexOf(search, StringComparison.OrdinalIgnoreCase);
-                    if (index == -1 || index < (current?.Index ?? int.MaxValue))
+                    if (index == -1 || index > (current?.Index ?? int.MaxValue))
```
revam, line 135:
```suggestion
                   if (index == -1 || index >= (current?.Index ?? int.MaxValue))
```

revam (approved):
Oh fudge. I actually ~~broke~~"fixed" this logic a few commits back, since i thought it was inverted. It was not (well, it was inverted, but it's supposed to be inverted, since it is supposed to skip it).

On that note, we don't want to replace our current match if the new match have the same distance or same index, so I've added two suggestions to bring back the greater than or equal to, instead of just the greater than.

PR #1068, opened 2023-06-16
Add workflow for daily build discord notification

harshithmohan:


.github/workflows/build-notifier.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Get current version
+        id: release_info
+        uses: revam/gh-action-get-tag-and-version@v1
+        with:
+          branch: true
+          prefix: v
+          prefixRegex: "[vV]?"
+
+      - id: commit_info
+        name: Shorten Commit Hash for Sentry Release
```
revam, line 33: Update the name.

.github/workflows/build-notifier.yml
```diff
@@ -0,0 +1,50 @@
...
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const sha = context.sha.substring(0, 7);
+            core.setOutput("sha", sha);
+            
+      - name: Notify discord
+        uses: sarisia/actions-status-discord@v1
```
revam, line 41: Can we use a discord bot/webhook that allow us to send plain messages (so notnembeds, but also with support for markdown), and have a few buttons (i know bots can add those) that link to our downloads page, github page, and/or docker, etc.?

.github/workflows/build-notifier.yml
```diff
@@ -0,0 +1,50 @@
...
+        with:
+          webhook: ${{ secrets.DISCORD_WEBHOOK }}
+          nodetail: true
+          title: New Daily Build!
+          description: |
+            **Version**: `${{ steps.release_info.outputs.version }}-dev-${{ steps.commit_info.outputs.sha }}`
+            
+            Latest Commit:
+            ${{ github.event.workflow_run.head_commit.message }}
```
revam, line 50: Should we include all the messages since last run of the workflow (so it will include the changes since the last run)?

No, I haven't tested if it's possibe yet, but i think it will be possible if we can store persistent variables between runs somehow or somewhere.

revam (commented):
Just some thoughts, also update the name of that one step please.

.github/workflows/build-notifier.yml
harshithmohan: I forgot...

.github/workflows/build-notifier.yml
harshithmohan: It's not possible without tagging every release or if you store previous release commit somewhere

.github/workflows/build-notifier.yml
harshithmohan: If you can find one. This one does markdown and we can add links.
Just no buttons. AFAIK, all webhooks are embeds. We would need a bot for what you want

revam: For now let's leave ifnoff the table. Maybe in a future revision to how we push the updates.

revam (approved):
LGTM

PR #1069, opened 2023-06-23
fix: fix up anidb episode removal

revam:


revam:
Missing the database migration for already broken data.

PR #1081, opened 2023-08-20
Add file relocation and renaming api for v3

revam:
I've tested the new file endpoints to a degree (on linux using sqlite), moving files around, renaming them, creating hard and soft copies, and squashed a few bugs while I was at it.

---

So this PR adds renaming to the v3 api, and cleans up the move/rename internals, adapts the v1 api to the new internals, among other things.


Shoko.Server/Mappings/VideoLocal_PlaceMap.cs
```diff
@@ -15,5 +15,7 @@ public VideoLocal_PlaceMap()
         Map(x => x.FilePath).Not.Nullable();
         Map(x => x.ImportFolderID).Not.Nullable();
         Map(x => x.ImportFolderType).Not.Nullable();
+        Map(x => x.AllowAutoRelocation).Not.Nullable();
```
da3dsoul, line 18: Why? Per file overrides on this seem completely unnecessary

revam: To prevent it from being automagically moved, e.g. you place a copy at one location for it to stay there, and not be moved away by shoko. If we had a better renaming system that allowed the user to specify _all_ the paths you would want for a file then I don't it will be needed, no. But until then then I have a personal need for this per-file-location setting to exist.

revam: An good example is for Toriko EP1 / One Piece EP420(?), where _I_ would want to place a hard copy in both directories, and not have shoko touch them.

da3dsoul: Well it won't get moved if it's in a destination. For predictability and reproducibility, you want your script to give it exactly one place it should go.

revam: Also, I think i forgot to expose a way to toggle the per-file location settings (allow auto relocate and allow auto delete) in the api…

da3dsoul: That's a fair example, I guess. It's a very niche feature.

Cazzar: This sounds a bit like over-engineering an issue that isn't really there.

sdaqo:
Will this PR make it possible to have one source import folder and another destination import folder where the files don't get moved but linked or copied instead. If yes awesome. Does it work well enough to test it?

da3dsoul:
No. Copying isn't really a good idea anyway. If you want copies, then use a copy on complete script in your downloader.
We can't implement hard linking because .net doesn't have an API for it. Shoko can move hard links if you make a hard link via a completion script in your downloader.

revam:
In case you didn't know @da3dsoul, i added and have tested adding hard linking support in this PR. They won't be made automatically by any renaming script, but it's possible to create them through the REST v3 API in this PR as of right now.

revam:
Also when hard linking (or just creating another location in general) then the two locations will automatically be excempt from auto deletion.

revam:
There are a few things to change up, so I've converted this PR to a draft for now, since I'm focusing on the other PR for adding TMDB support.

revam:
> Will this PR make it possible to have one source import folder and another destination import folder where the files don't get moved but linked or copied instead? If yes awesome. Does it work well enough to test it?

While it might make that possible it won't make it possible with a traditional renamer, you would have to use the file events directly and also either use the REST api or reference the shoko dlls directly and not using the abstraction.

revam:
Closed by cfaffe5. And no. No Hard linking were added. Only the file relocation and renamer/renamer scripts.

PR #1082, opened 2023-08-20
[WIP] Add TMDB support

revam:


.editorconfig
```diff
@@ -193,8 +193,8 @@ dotnet_naming_rule.instance_fields_should_be_camel_case.style = instance_field_s
 
 dotnet_naming_symbols.instance_fields.applicable_kinds = field
 
-dotnet_naming_style.instance_field_style.capitalization = camel_case
-dotnet_naming_style.instance_field_style.required_prefix = _
+dotnet_naming_style.instance_field_style.capitalization = pascal_case
+dotnet_naming_style.instance_field_style.required_prefix =
```
Cazzar, line 197: What's happened here?

da3dsoul: Indeed. No pascal case

Shoko.Plugin.Abstractions/Extensions/LanguageExtensions.cs
```diff
@@ -119,11 +122,27 @@ public static TitleLanguage GetTitleLanguage(this string lang)
             "YI" or "YID" => TitleLanguage.Yiddish,
             "YO" or "YOR" => TitleLanguage.Yoruba,
             "ZU" or "ZUL" => TitleLanguage.Zulu,
-            "UNK" => TitleLanguage.Unknown,
-            _ => TitleLanguage.Unknown,
+            "UR" or "URD" => TitleLanguage.Urdu,
+            "GREEK (ANCIENT)" => TitleLanguage.Greek,
```
da3dsoul, line 126: What's all this? The format changed entirely

Shoko.Server/API/v1/Implementations/ShokoServiceImplementation/ShokoServiceImplementation.cs
```diff
@@ -938,84 +938,72 @@ public string EnableDisableImage(bool enabled, int imageID, int imageType)
...
 
                     poster.Enabled = enabled ? 1 : 0;
                     RepoFactory.TvDB_ImagePoster.Save(poster);
-                    animeID = RepoFactory.CrossRef_AniDB_TvDB.GetByTvDBID(poster.SeriesID).FirstOrDefault()?.AniDBID ?? 0;
+                    foreach (var xref in RepoFactory.CrossRef_AniDB_TvDB.GetByTvDBID(poster.SeriesID))
+                        animeIDs.Add(xref.AniDBID);
                     break;
 
-                case ImageEntityType.TvDB_FanArt:
+                case CL_ImageEntityType.TvDB_FanArt:
```
da3dsoul, line 977: Why rename this? I'm against CL_ shit. We have a ton of clients. It should be more specific if it needs to be separate

Shoko.Server/API/v3/Controllers/ImageController.cs
```diff
@@ -20,64 +21,38 @@ public class ImageController : BaseController
...
+    /// <param name="source">AniDB, TvDB, TMDB, Shoko, etc.</param>
+    /// <param name="type">Poster, Fanart, Banner, Thumbnail, etc.</param>
+    /// <param name="value">The image ID.</param>
     /// <returns>200 on found, 400/404 if the type or source are invalid, and 404 if the id is not found</returns>
     [HttpGet("{source}/{type}/{value}")]
     [ProducesResponseType(typeof(FileStreamResult), 200)]
     [ProducesResponseType(404)]
     public ActionResult GetImage([FromRoute] Image.ImageSource source, [FromRoute] Image.ImageType type,
-        [FromRoute] string value)
+        [FromRoute] int value)
```
da3dsoul, line 32: This looks like a breaking change. Please comment

.editorconfig
revam: I changed it to get the IDE to shut up about it's "suggestions" to rename the fields. My IDE was suggesting to renaming _all_ properties to use camel case and a prefix. If you can somehow tweak the rules to only apply the prefix and camel case to the _private_ properties, then be my guest, i won't invest time into finding the correct settings to use atm.

Shoko.Server/API/v1/Implementations/ShokoServiceImplementation/ShokoServiceImplementation.cs
revam: It was renamed since only v1 will be using it by the end. I also added a new ImageEntityType void of data sources mixed in, so we can use it with a separate data source field when needed. I _can_ rewrite the v1 layer to use the new enum in combination with a data source, but I didn't feel like touching the v1 (and v2) api layer more than I needed to.

.editorconfig
revam: And yes, I can remove the commit that changed these settings from this PR, but it's staying for now, so i can keep my sanity… for now.

Shoko.Server/API/v3/Controllers/ImageController.cs
revam: Also, the way we use the image metadata in the web ui makes it so this change won't break the web ui. Shokofin might break, but I can fix that in the "unstable" channel if it does break it.

Shoko.Server/API/v1/Implementations/ShokoServiceImplementation/ShokoServiceImplementation.cs
da3dsoul: If it's just Desktop, then making it obey Desktop's naming scheme seems correct

Shoko.Server/API/v3/Controllers/ImageController.cs
revam: Looking at @harshithmohan's v3 branch for Metadata, then it doesn't look like it will break since it's just using the JSON values as is without any type checking (which is done in Shokofin, since it's C#).

https://github.com/harshithmohan/ShokoMetadata.bundle/blob/apiv3/Contents/Code/__init__.py#L481

Shoko.Server/API/v1/Implementations/ShokoServiceImplementation/ShokoServiceImplementation.cs
revam: I moved the logic into the image utils helper.

PR #1087, opened 2023-09-15
TraktTVHelper: Fix wrong match for continuous ep numbering across seasons

krbrs:
This fixes wrong ep numbers being used for sync with series like One Piece, where the entry was migrated to TMDB as Trakt's source and where continuous numbering across seasons is used (instead of starting over at 1, as it was with TVDB as Trakt's source before).

This change in numbering broke the old logic behind season/ep matching between AniDB and Trakt.

Should not be a problem with Specials, still I changed the logic here too, as they have been the same before. Also, who knows if the same will happen to Specials one day...

Also some housekeeping/spellchecking on TraktTVHelper.cs

### DIFF
--- a/Shoko.Server/Providers/TraktTV/TraktTVHelper.cs
+++ b/Shoko.Server/Providers/TraktTV/TraktTVHelper.cs
@@ -604,7 +604,7 @@ public class TraktTVHelper
                     traktSummary.CrossRefTraktV2.Count > 0)
                 {
                     // find the xref that is right
-                    // relies on the xref's being sorted by season number and then episode number (desc)
+                    // relies on the xrefs being sorted by season number and then episode number (desc)
                     var traktCrossRef =
                         traktSummary.CrossRefTraktV2.OrderByDescending(a => a.AniDBStartEpisodeNumber).ToList();
 
@@ -625,7 +625,7 @@ public class TraktTVHelper
                         }
                     }
 
-                    // we have found the starting epiosde numbder from AniDB
+                    // we have found the starting episode number from AniDB
                     // now let's check that the Trakt Season and Episode Number exist
                     if (foundStartingPoint)
                     {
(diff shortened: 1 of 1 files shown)


Shoko.Server/Providers/TraktTV/TraktTVHelper.cs
```diff
@@ -1822,9 +1848,7 @@ public void SyncCollectionToTrakt()
                 TraktTVRateLimiter.Instance.EnsureRate();
                 SendData(url, json, "POST", BuildRequestHeaders(), ref retData);
             }
-
-
-            _logger.LogTrace("Test");
+
```
revam, line 1851: Very minor but you can remove the empty newline too.

Shoko.Server/Providers/TraktTV/TraktTVHelper.cs
```diff
@@ -641,14 +641,27 @@ public void ScanForMatches()
...
 
-                        if (dictTraktSeasons != null && dictTraktSeasons.ContainsKey(xrefBase.TraktSeasonNumber))
+                        if (dictTraktSeasons != null && dictTraktSeasons.TryGetValue(xrefBase.TraktSeasonNumber, out var traktSeason))
                         {
-                            var episodeNumber = dictTraktSeasons[xrefBase.TraktSeasonNumber] +
-                                                (ep.EpisodeNumber + xrefBase.TraktStartEpisodeNumber - 2) -
-                                                (xrefBase.AniDBStartEpisodeNumber - 1);
-                            if (dictTraktEpisodes.ContainsKey(episodeNumber))
+                            int episodeNumber;
+                            {
```
revam, line 647: Why are you creating another scope just for the if/else statements?

Shoko.Server/Providers/TraktTV/TraktTVHelper.cs
```diff
@@ -641,14 +641,27 @@ public void ScanForMatches()
...
+                                if (xrefBase.TraktStartEpisodeNumber == xrefBase.AniDBStartEpisodeNumber)
+                                {
+                                    // The Trakt and AniDB start episode numbers match
+                                    episodeNumber = (traktSeason - xrefBase.TraktStartEpisodeNumber ) + ep.EpisodeNumber;
+                                }
+
```
revam, line 653: Please remove the empty line between the if/else statements if you won't have a comment on the line. Functionality it works, but it's not in line with how we write if/else statements in this project.

Shoko.Server/Providers/TraktTV/TraktTVHelper.cs
```diff
@@ -708,14 +721,27 @@ public void ScanForMatches()
...
 
-                    if (dictTraktSeasons != null && dictTraktSeasons.ContainsKey(xrefBase.TraktSeasonNumber))
+                    if (dictTraktSeasons != null && dictTraktSeasons.TryGetValue(xrefBase.TraktSeasonNumber, out var traktSeason))
                     {
-                        var episodeNumber = dictTraktSeasons[xrefBase.TraktSeasonNumber] +
-                                            (ep.EpisodeNumber + xrefBase.TraktStartEpisodeNumber - 2) -
-                                            (xrefBase.AniDBStartEpisodeNumber - 1);
-                        if (dictTraktEpisodes != null && dictTraktEpisodes.ContainsKey(episodeNumber))
+                        int episodeNumber;
+                        {
```
revam, line 727: Same as above.

Shoko.Server/Providers/TraktTV/TraktTVHelper.cs
```diff
@@ -708,14 +721,27 @@ public void ScanForMatches()
...
+                            if (xrefBase.TraktStartEpisodeNumber == xrefBase.AniDBStartEpisodeNumber)
+                            {
+                                // The Trakt and AniDB start episode numbers match
+                                episodeNumber = (traktSeason - xrefBase.TraktStartEpisodeNumber ) + ep.EpisodeNumber;
+                            }
+
```
revam, line 733: Same as above.

revam (commented):
Functionality it looks fine, the only thing i found were minor nit-picks.

Shoko.Server/Providers/TraktTV/TraktTVHelper.cs
krbrs: It did some weird things when I didn't do this before, going to re-check this and resolve

krbrs: looks like I had an error somewhere there before which is now gone. going to commit

revam (approved):
LGTM 👍

PR #1094, opened 2023-11-12
Fix image validation for TVDB episode images

LordFlashmeow:
Valid images were being deleted and redownloaded instead of invalid images.


revam (approved):
LGTM :+1:

PR #1098, opened 2024-01-12
Work on CheckForTraktTokenUpdate

krbrs:
Have it read the expiration time from settings and not just refresh the token every 24h, only if it's going to expire in 45 days

### DIFF
--- a/Shoko.Server/Import/Importer.cs
+++ b/Shoko.Server/Import/Importer.cs
@@ -1438,49 +1438,41 @@ public static class Importer
         try
         {
             var settings = Utils.SettingsProvider.GetSettings();
-            if (!settings.TraktTv.Enabled)
+            if (!settings.TraktTv.Enabled || string.IsNullOrEmpty(settings.TraktTv.TokenExpirationDate))
             {
                 return;
             }
 
             var traktHelper = Utils.ServiceContainer.GetRequiredService<TraktTVHelper>();
-            // by updating the Trakt token regularly, the user won't need to authorize again
-            var freqHours = 24; // we need to update this daily
 
-            var sched =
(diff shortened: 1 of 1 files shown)


krbrs:
I could not test expiration and renewal yet. But at least I get a proper output of the token valid date in the logs while it's still valid.

If someone can look into this code and optimize it if needed (or give feedback), that would be great

Shoko.Server/Import/Importer.cs
```diff
@@ -1438,49 +1438,41 @@ public static void CheckForTraktTokenUpdate(bool forceRefresh)
...
-            var freqHours = 24; // we need to update this daily
 
-            var sched =
-                RepoFactory.ScheduledUpdate.GetByUpdateType((int)ScheduledUpdateType.TraktToken);
-            if (sched != null)
+            // Convert the Unix timestamp to DateTime directly
+            var expirationDate = DateTimeOffset.FromUnixTimeSeconds(long.Parse(settings.TraktTv.TokenExpirationDate)).DateTime;
+
+            // Check if force refresh is requested or the token is expired
+            if (forceRefresh || DateTime.Now.Add(TimeSpan.FromDays(45)) >= expirationDate)
```
da3dsoul, line 1452: Why 45 days?

krbrs: https://discord.com/channels/96234011612958720/1148092356592160768/1195127627397214331

Initially I had it set to 30 days, but half would be 45 days so I set this. We can set it to whatever is fine

da3dsoul: Half of what

da3dsoul: Fair enough. Looks good then

PR #1104, opened 2024-02-13
Update TagFilter.cs: Remove Some Obsolete Tags / Add Some New Ones

natyusha:
Mainly adding some parent tags to `AniDBHelpers` but found some tags that no longer exist and some other obvious tags to add along the way. Also, swapped "nevada" to `Setting` and "original work" to `AniDBHelpers`


revam:
you do know that we rename the "new" tag to "original work" and the "original work" tag to "source material" (not quite sure what we rename it to, but something other than "original work") on our side before using the tags, right? so moving the "original work" string (AKA the "new" tag) out of the stribg collection for sources is wrong imo. Also, do correct me if I'm wrong.

da3dsoul:
Revam addressed my concerns. Original Work and New were very misleading if seen in Shoko, so they were replaced. It is also considered an anime original (original work) if no other source tag is given, since it's somewhat common to be omitted.

natyusha:
I didn't realize you were renaming tags in the backend as I was just looking to quickly remove some parent tags from Plex with the `AniDBHelpers` filters and didn't look into it. I have 1174 series that have the original work tag which seemed excessive at the time but makes sense now.

I can see the reasoning here but there isn't any clear indication that certain tags are being renamed (that I know of). The web ui `/webui/collection/series/seriesid/tags` page also shows the description for those tags which don't match anidb. Maybe appending an asterisk or something to the tag name in the ui to let users know would be a good idea but I'm not sure.

revam (approved):

PR #1112, opened 2024-04-01
Create compressed MyList backup before syncing

Mik1ll:
* Backups are timestamped
* Deletes oldest backups when there are more than 30

### DIFF
--- a/Shoko.Server/Scheduling/Jobs/AniDB/SyncAniDBMyListJob.cs
+++ b/Shoko.Server/Scheduling/Jobs/AniDB/SyncAniDBMyListJob.cs
@@ -141,7 +141,7 @@ public class SyncAniDBMyListJob : BaseJob
             totalItems, missingFiles, filesToRemove.Count, watchedItems, modifiedItems);
     }
 
-    private static async Task CreateMyListBackup(HttpResponse<List<ResponseMyList>> response)
+    private async Task CreateMyListBackup(HttpResponse<List<ResponseMyList>> response)
     {
         var serialized = JsonConvert.SerializeObject(response.Response, Formatting.Indented);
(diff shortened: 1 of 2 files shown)


Shoko.Server/Scheduling/Jobs/AniDB/SyncAniDBMyListJob.cs
```diff
@@ -142,6 +141,40 @@ public override async Task Process()
...
+        // Backup rotation depeonds on filename being universally sortable ("u" format specifier)
+        var archivePath = Path.Join(myListDirectory.FullName, DateTimeOffset.UtcNow.ToString("u").Replace(':', '_') + ".zip");
+        await using var backupFs = new FileStream(archivePath, FileMode.OpenOrCreate);
+        using var archive = new ZipArchive(backupFs, ZipArchiveMode.Create);
+        archive.CreateEntryFromFile(currentBackupPath, Path.GetFileName(currentBackupPath));
+
+        // Delete oldest backups when more than 30 exist
+        // Only gets zip files that start with ISO 8601 date format YYYY-MM-DD
+        var backupFiles = myListDirectory.GetFiles("????-??-?? *.zip").OrderByDescending(f => f.Name).ToList();
+        var retainedBackupCount = 30;
```
Cazzar, line 163: Can this be a variable in the config file?

We don't have to expose it in the frontends, but it would be convenient to be able to configure how many backups are stored.

Mik1ll: Sure, I haven't added a config option before, so not sure how many files need to edited considering DTO, mapping, and empty value population. I can take a look at it in a few days.

PR #1113, opened 2024-04-03
Fix: Emit the actual duration of an AniDB ban

fearnlj01:
Apparently, only the seconds component of the ban duration was emitted, rather than the total duration in seconds... I was a little confused when the value was always returning `0`!

This looks to get things working as expected on my end (including in the current WebUI), but I'm not too sure if changing this might cause any issues elsewhere...


da3dsoul:
Facepalm good catch

PR #1116, opened 2024-04-11
Rename OriginalFilename

Mik1ll:
Original filename is not representative of the string's meaning. The original is not returned by AniDB api, an internal version of the filename is

### DIFF
--- a/Shoko.Plugin.Abstractions/DataModels/IAniDBFile.cs
+++ b/Shoko.Plugin.Abstractions/DataModels/IAniDBFile.cs
@@ -17,9 +17,9 @@ namespace Shoko.Plugin.Abstractions.DataModels
         /// </summary>
         string Source { get; }
         /// <summary>
-        /// The Filename as released, according to AniDB. It's usually correct.
+        /// The file name used internally by AniDB. It has a consistant format.
         /// </summary>
-        string OriginalFilename { get; }
+        string FileName { get; }
         /// <summary>
         /// Description of the file on AniDB. This will often be blank, and it's generally not useful
         /// </summary>
--- a/Shoko.Plugin.Abstractions/Shoko.Plugin.Abstractions.csproj
+++ b/Shoko.Plugin.Abstractions/Shoko.Plugin.Abstractions.csproj
@@ -12,7 +12,7 @@
         <RepositoryUrl>https://github.com/ShokoAnime/ShokoServer</RepositoryUrl>
         <PackageTags>plugins, shoko, anime, metadata, tagging</PackageTags>
         <PackageReleaseNotes>File Events</PackageReleaseNotes>
-        <Version>3.0.0-alpha2</Version>
+        <Version>3.0.0-alpha3</Version>
         <Configurations>Debug;Release</Configurations>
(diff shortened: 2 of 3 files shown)


revam (approved):
This may break any plugins which uses this field, but doubt there are many, if any at all.

Cazzar:
Also, given that it's an API break, you want more than just updating the build metadata @revam @Mik1ll

Shoko.Plugin.Abstractions/DataModels/IAniDBFile.cs
```diff
@@ -17,9 +17,9 @@ public interface IAniDBFile
         /// </summary>
         string Source { get; }
         /// <summary>
-        /// The Filename as released, according to AniDB. It's usually correct.
+        /// The file name used internally by AniDB. It has a consistant format.
```
Cazzar, line 20: image
Are you sure this is even consistent?

Shoko.Plugin.Abstractions/DataModels/IAniDBFile.cs
```diff
@@ -17,9 +17,9 @@ public interface IAniDBFile
...
         /// </summary>
-        string OriginalFilename { get; }
+        string FileName { get; }
```
Cazzar, line 22: I suggest we think of a better name of variable in this case

Shoko.Plugin.Abstractions/Shoko.Plugin.Abstractions.csproj
```diff
@@ -12,7 +12,7 @@
         <RepositoryUrl>https://github.com/ShokoAnime/ShokoServer</RepositoryUrl>
         <PackageTags>plugins, shoko, anime, metadata, tagging</PackageTags>
         <PackageReleaseNotes>File Events</PackageReleaseNotes>
-        <Version>3.0.0-alpha2</Version>
+        <Version>3.0.0-alpha3</Version>
```
Cazzar, line 15: This is more than a metadata bump

revam:
will it be fine to add the original field back as an obsolete field be enough to avoid the api breakage? Also, if FileName isn't enough, should it be appended by 'Internal' or something similar?

Shoko.Plugin.Abstractions/DataModels/IAniDBFile.cs
revam: older versions of shoko overwrote the file name locally with the file name at the time of linking. as for newer names, i simply don't know.

Mik1ll: Something changed in how Shoko stored the data, it seems that Shoko wasn't using what was returned by AniDB before.   None of those files appear to be in the format AniDB returns.

Cazzar: Then simply put, we can't trust that to be the case in all cases, since we have people with Shoko instances that old, we need to consider these.

Telling people to re-import all their data isn't going to be feasible, or reasonable at the least, without us having a way of correcting this to the "correct" data.

PR #1118, opened 2024-04-13
Split the anidb address & port settings for HTTP and UDP

harshithmohan:
As per the discussion in #support. To have the ability to use a different address for the HTTP API

- Moved UDP address and port settings into a separate property.
- Desktop will only be able to view and edit the HTTP address and port settings

### DIFF
--- a/Shoko.Server/Providers/AniDB/UDP/AniDBUDPConnectionHandler.cs
+++ b/Shoko.Server/Providers/AniDB/UDP/AniDBUDPConnectionHandler.cs
@@ -94,8 +94,8 @@ public class AniDBUDPConnectionHandler : ConnectionHandler, IUDPConnectionHandle
     public async Task<bool> Init(string username, string password, string serverName, ushort serverPort, ushort clientPort)
     {
         var settings = SettingsProvider.GetSettings();
-        settings.AniDb.ServerAddress = serverName;
-        settings.AniDb.ServerPort = serverPort;
+        settings.AniDb.UDPServerAddress = serverName;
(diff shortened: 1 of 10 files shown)


Shoko.Server/Providers/AniDB/UDP/AniDBUDPConnectionHandler.cs
```diff
@@ -114,10 +114,10 @@ private async Task InitInternal()
         }
 
         var settings = SettingsProvider.GetSettings();
-        ArgumentNullException.ThrowIfNull(settings.AniDb?.ServerAddress);
-        if (settings.AniDb.ServerPort <= 0) throw new ArgumentException("AniDB Server Port is invalid");
+        ArgumentNullException.ThrowIfNull(settings.AniDb?.UDPServerAddress);
+        if (settings.AniDb.UDPServerPort <= 0) throw new ArgumentException("AniDB Server Port is invalid");
```
fearnlj01, line 118: Would it be better to tweak the exception to specifically state that it's the `UDP` Server Port that's invalid given that ServerPort would be a different setting?

Cazzar: This case will never hit.

https://dotnetfiddle.net/NnHiBB

```csharp
ushort a = -1;
```
is a compiler error.

harshithmohan: So the line can just be removed then?

PR #1119, opened 2024-04-13
Split the anidb address & port settings for HTTP and UDP

harshithmohan:
As per the discussion in #support. To have the ability to use a different address for the HTTP API

- Combined HTTP address and port into `HTTPServerUrl`
- Moved UDP address and port settings into a separate property.
- Desktop will only be able to view and edit the HTTP address and port settings
- Too much regex

### DIFF
--- a/Shoko.Server/API/v1/Implementations/ShokoServiceImplementation/ShokoServiceImplementation.cs
+++ b/Shoko.Server/API/v1/Implementations/ShokoServiceImplementation/ShokoServiceImplementation.cs
@@ -479,9 +479,9 @@ public partial class ShokoServiceImplementation : Controller, IShokoServer
                 }
             }
 
-            var serverAddressMatches = Regex.Match(settings.AniDb.HTTPServerUrl, @"(?<=https?://)(?<address>.*):(?<port>\d+)");
-            var oldAniDB_ServerAddress = serverAddressMatches.Groups["address"].Value;
-            var oldAniDB_ServerPort = serverAddressMatches.Groups["port"].Value;
+            var serverAddressGroups = Regex.Match(settings.AniDb.HTTPServerUrl, @"(?<=https?://)(?<address>.*):(?<port>\d+)").Groups;
+            var oldAniDB_ServerAddress = serverAddressGroups["address"].Value;
+            var oldAniDB_ServerPort = serverAddressGroups["port"].Value;
             var newAniDB_HTTPServerUrl = $"{contractIn.AniDB_ServerAddress}:{contractIn.AniDB_ServerPort}";
 
             if (contractIn.AniDB_ServerAddress != oldAniDB_ServerAddress || contractIn.AniDB_ServerPort != oldAniDB_ServerPort)
--- a/Shoko.Server/API/v1/Implementations/ShokoServiceImplementation/ShokoServiceImplementation.cs
+++ b/Shoko.Server/API/v1/Implementations/ShokoServiceImplementation/ShokoServiceImplementation.cs
@@ -51,6 +51,8 @@ public partial class ShokoServiceImplementation : Controller, IShokoServer
     private readonly ISchedulerFactory _schedulerFactory;
     private readonly ActionService _actionService;
 
+    private readonly Regex _urlRegex = new Regex(@"(?<=https?://)(?<address>.*):(?<port>\d+)", RegexOptions.Compiled);
+
     public ShokoServiceImplementation(TvDBApiHelper tvdbHelper, TraktTVHelper traktHelper, MovieDBHelper movieDBHelper, ISchedulerFactory schedulerFactory, ISettingsProvider settingsProvider, ILogger<ShokoServiceImplementation> logger, ActionService actionService, AnimeGroupCreator groupCreator, JobFactory jobFactory)
     {
         _tvdbHelper = tvdbHelper;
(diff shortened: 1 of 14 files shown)


Shoko.Server/Extensions/ModelClients.cs
```diff
@@ -22,8 +23,8 @@ public static CL_ServerSettings ToContract(this IServerSettings settings)
         {
             AniDB_Username = settings.AniDb.Username,
             AniDB_Password = settings.AniDb.Password,
-            AniDB_ServerAddress = settings.AniDb.ServerAddress,
-            AniDB_ServerPort = settings.AniDb.ServerPort.ToString(),
+            AniDB_ServerAddress = Regex.Match(settings.AniDb.HTTPServerUrl, @"(https?://)*[^:]+").ToString(),
```
fearnlj01, line 26: Should the regex be `@"(https?://).*?[^:]+")` instead? Unless c# regex works differently to what I'm used to, it's the `https://` capture group that it tries to match 0+ times otherwise.

whacked a few test urls into https://regex101.com/ using the above and I _think_ my suggestion isn't baseless

fearnlj01: Testing the existing regex on single line strings... Does work as expected :)

I'm the fool that tried being lazy and testing on multiline strings without adapting the regex accordingly.

harshithmohan: Either is right if you didn't know what I was expecting it to do. Such is the way of regex 😔

harshithmohan:
I'll revert the last commit if it's not ok

Shoko.Server/Providers/AniDB/HTTP/RequestGetAnime.cs
```diff
@@ -13,21 +13,19 @@ public class RequestGetAnime : HttpRequest<ResponseGetAnime>
     private readonly HttpXmlUtils _xmlUtils;
     private readonly HttpAnimeParser _parser;
     private readonly string _aniDBUrl;
-    private readonly ushort _aniDBPort;
 
     public int AnimeID { get; set; }
 
     protected override string BaseCommand =>
-        $"http://{_aniDBUrl}:{_aniDBPort}/httpapi?client=animeplugin&clientver=1&protover=1&request=anime&aid={AnimeID}";
+        $"{_aniDBUrl}/httpapi?client=animeplugin&clientver=1&protover=1&request=anime&aid={AnimeID}";
```
Cazzar, line 20: Instead of us needing to set the URL on every single one, would it be worth setting this? https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.baseaddress?view=net-8.0

harshithmohan: I'll go cry in a corner wondering why I didn't think of that 😔

Shoko.Server/API/v1/Implementations/ShokoServiceImplementation/ShokoServiceImplementation.cs
```diff
@@ -486,13 +479,30 @@ public CL_Response SaveServerSettings(CL_ServerSettings contractIn)
                 }
             }
 
-            if (contractIn.AniDB_ServerAddress != settings.AniDb.ServerAddress)
+            var serverAddressGroups = Regex.Match(settings.AniDb.HTTPServerUrl, @"(?<=https?://)(?<address>.*):(?<port>\d+)").Groups;
```
da3dsoul, line 482: I would move that to a private readonly compiled regex.

Something like:
```cs
private readonly Regex _urlRegex = new Regex("(?<=https?://)(?<address>.*):(?<port>\d+)", RegexOptions.Compiled);
```

Called with
```cs
serverAddressGroups = _urlRegex.Match(settings.AniDb.HTTPServerUrl).Groups;
```

I'm going from memory, so it might not be quite right

Shoko.Server/Extensions/ModelClients.cs
```diff
@@ -22,8 +23,8 @@ public static CL_ServerSettings ToContract(this IServerSettings settings)
         {
             AniDB_Username = settings.AniDb.Username,
             AniDB_Password = settings.AniDb.Password,
-            AniDB_ServerAddress = settings.AniDb.ServerAddress,
-            AniDB_ServerPort = settings.AniDb.ServerPort.ToString(),
+            AniDB_ServerAddress = Regex.Match(settings.AniDb.HTTPServerUrl, @"https?://[^:]+").ToString(),
```
da3dsoul, line 26: private readonly compiled regex

Shoko.Server/Settings/SettingsMigrations.cs
```diff
@@ -87,4 +88,26 @@ private static string MigrateAutoGroupRelationsAlternateToAlternative(string set
         var regex = new Regex(@"(?<=""AutoGroupSeriesRelationExclusions""\s*:\s*\[)(?<value>[^\]]+)", RegexOptions.Compiled);
         return regex.Replace(settings, match => match.Groups["value"].Value.Replace("alternate", "alternative", StringComparison.InvariantCultureIgnoreCase));
     }
+
+    private static string MigrateAniDBServerAddresses(string settings)
+    {
+        ushort serverPort = 9001;
+
+        var serverPortRegex = new Regex(@"(?<=""AniDb""\s*:\s*\{.*)""ServerPort""\s*:\s*(?<value>\d+)", RegexOptions.Compiled | RegexOptions.Singleline);
```
da3dsoul, line 96: These are fine to leave as is. Reusing compiled regexes helps speed up hot paths. This only runs once

da3dsoul:
Otherwise looks good

PR #1122, opened 2024-04-29
Fix: Add timeouts to async AniDB UDP calls to prevent queue deadlocks

fearnlj01:
I'm not entirely confident that this will fix the longstanding issue with queue deadlocks, however while debugging, the changes have appeared to play nice when manually cancelling the token, resulting in a second attempt at the UDP call being made as expected.

I'm not exactly too sure how to simulate a "natural timeout" of the cancellation token and thus haven't tested the changes in this manner.

As I don't understand it fully, I also want to note that `payload`/`result` are accepted by `SendToAsync` & `ReceiveFromAsync` as `ReadOnlyMemory<byte>` rather than `ArraySegment<byte>` when using a CancellationToken. Given that "it works" experimentally, I figured I'd PR regardless and mention it here in case this is an issue.

### DIFF
--- a/Shoko.Server/Providers/AniDB/UDP/AniDBSocketHandler.cs
+++ b/Shoko.Server/Providers/AniDB/UDP/AniDBSocketHandler.cs
@@ -22,6 +22,7 @@ public class AniDBSocketHandler : IAniDBSocketHandler
     private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(1, 1);
     private bool Locked { get; set; }
     public bool IsLocked => Locked;
+    private static int TimeoutMs => 30000; // 30 seconds
 
     public AniDBSocketHandler(ILoggerFactory loggerFactory, string host, ushort serverPort, ushort clientPort)
     {
@@ -45,13 +46,12 @@ public class AniDBSocketHandler : IAniDBSocketHandler
             {
                 result = await SendUnsafe(payload, result);
             }
-            catch (SocketException e)
+            catch (Exception e) when ((e is SocketException se && se.SocketErrorCode == SocketError.TimedOut) ||
+                                      (e is TaskCanceledException tce && tce.CancellationToken.IsCancellationRequested))
             {
-                if (e.SocketErrorCode == SocketError.TimedOut)
-                {
-                    // we will retry once and only once.
-                    result = await SendUnsafe(payload, result);
-                }
+                // catches timeouts in both synchronous and asynchronous calls
+                // we will retry once and only once.
+                result = await SendUnsafe(payload, result);
             }
         }
         catch (Exception e)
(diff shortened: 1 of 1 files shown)


fearnlj01:
Just to add some further notes on this... I accidentally left a breakpoint set overnight on the catch of `SocketException`/`TaskCancelledException`s. Was wondering why I got a timeout trying to open the WebUI just now and it appears to have hit this exception of it's own accord.

Logs look like the below after resuming.
image

Not sure if I should amend then PR to explicitly try catch `OperationCancelledException` as well?

da3dsoul:
The unknown command thing is happening because of an invalid login state. I'll try to fix it, but it'll be a bit before I have time to

da3dsoul:
The other stuff is good

fearnlj01:
> The unknown command thing is happening because of an invalid login state. I'll try to fix it, but it'll be a bit before I have time to

This part is my bad - My testing involved restarting the server over and over to make sure I wasn't just assuming my changes worked as expected so the login state was persistently borked upon server restarts.

I ~~was too lazy~~ couldn't figure out a quick way to not impact the login command for a realistic screenshot so this was the unfortunate byproduct.

Shoko.Server/Providers/AniDB/UDP/AniDBSocketHandler.cs
```diff
@@ -45,15 +46,18 @@ public async Task<byte[]> Send(byte[] payload)
...
-                {
-                    // we will retry once and only once.
-                    result = await SendUnsafe(payload, result);
-                }
+                // catches timeouts in both synchronous and asynchronous calls
+                // we will retry once and only once.
+                result = await SendUnsafe(payload, result);
             }
         }
+        catch (TaskCanceledException e) when (e.CancellationToken.IsCancellationRequested)
```
revam, line 57: The `when` here is unneeded impo. Unless we have another cancellation token source that might also cancel it.

Shoko.Server/Providers/AniDB/UDP/AniDBSocketHandler.cs
```diff
@@ -22,6 +22,7 @@ public class AniDBSocketHandler : IAniDBSocketHandler
     private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(1, 1);
     private bool Locked { get; set; }
     public bool IsLocked => Locked;
+    private static int TimeoutMs => 30000; // 30 seconds
```
revam, line 25: Maybe add one timeout for send and one for receive? They can both be set to 30 seconds for now, but it will make it easier to configure them individually in the future, if the need arise.

revam (commented):
My 2 cents.

Shoko.Server/Providers/AniDB/UDP/AniDBSocketHandler.cs
fearnlj01: I'm also personally in agreement here. The `when` was added due to an overabundance of caution. Will remove it now, not like it'd be too hard to re-add in the future if it does become relevant 👍

Shoko.Server/Providers/AniDB/UDP/AniDBSocketHandler.cs
fearnlj01: Will action in next commit.

revam (approved):

Shoko.Server/Providers/AniDB/UDP/AniDBSocketHandler.cs
```diff
@@ -22,6 +22,8 @@ public class AniDBSocketHandler : IAniDBSocketHandler
     private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(1, 1);
     private bool Locked { get; set; }
     public bool IsLocked => Locked;
+    private static int SendTimeoutMs => 30000; // 30 seconds
+    private static int ReceiveTimeoutMs => 30000; // 30 seconds
```
Cazzar, line 26: For future reference, what would be better is using TimeSpan objects to specify this, then you can use [TimeSpan.TotalMilliseconds](https://learn.microsoft.com/en-us/dotnet/api/system.timespan.totalmilliseconds?view=net-8.0)

PR #1123, opened 2024-05-02
Add `/Episode/Watched` endpoint

fearnlj01:
The `/Episode/Watched` endpoint allows for a list of episodes to be marked as watched in bulk, regardless of what series they belong to.

Most practically, this could be utilised by the Web UI to allow marking a selection of episodes as watched, rather than having to mark everything that matches a search filter as watched.

I have tested the changes, and everything appears to have worked as (I) expected it to do so. For reference, requests with invalid episode IDs are rejected rather than filtering the invalid input out to avoid any misleading responses to the endpoint.

It goes without saying, but like with the other episode watched endpoints, if there are no files linked to the episode to mark as watched, it won't actually impact anything but will return a status of `200`.

### DIFF
--- a/Shoko.Server/API/v3/Controllers/EpisodeController.cs
+++ b/Shoko.Server/API/v3/Controllers/EpisodeController.cs
@@ -488,8 +488,8 @@ public class EpisodeController : BaseController
         return Ok();
     }
 
-    private void SetWatchedStatusOnEpisode(SVR_AnimeEpisode episode, bool watched)
-        => episode.ToggleWatchedStatus(watched, true, DateTime.Now, true, User.JMMUserID, true);
+    private void SetWatchedStatusOnEpisode(SVR_AnimeEpisode episode, bool? watched)
+        => episode.ToggleWatchedStatus(watched ?? false, true, DateTime.Now, true, User.JMMUserID, true);
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


Cazzar:
Does this really provide much value over having something hit the API many times? Given it's local, we don't implement any rate limiting, and it should be able to handle plenty

fearnlj01:
I was more so considering that this endpoint would show value when used with a long running series like Detective Conan/One Piece.

I worked on this under the assumption that client devices would probably struggle making 100+ requests all at the same time. I'll do some provisional testing with a local build of the webui to actually validate if this is the case or not.

Cazzar:
The only devices I could forsee having those kind of issues would be cheaper, low-end mobile devices, and WebUI already has enough issues in mobile viewports.

fearnlj01:
Closing this PR to allow for a more optimised approach to achieving this on a per-series basis.

No point in keeping clutter around!

bigretromike:
I would see it being usefull when you retrospective want to sync your collection.
Like when I was thinking that API endpoint for syncing watched state for episode was working and was using it for 3 years to end up looking it up and all those episodes didn't get synced, some series wasn't completed. but yeah , there would have to be a interface to check all those if this is just a bulk update - if not there is value there.

PR #1126, opened 2024-05-11
Plex webhook episode detection for non-episode videos

LordFlashmeow:
~~Use title + episode number (or TVDB season + episode #) for non-Episode type videos.~~

Use Plex API to fetch the filename of the watched video file (for non-episode type videos). 

This should allow the Plex webhook to mark specials and other video types as watched.

There is a corner case where the user has multiple episodes with the same file name in a series, handling that is TBD.

### DIFF
--- a/Shoko.Server/API/v2/Modules/PlexWebhook.cs
+++ b/Shoko.Server/API/v2/Modules/PlexWebhook.cs
@@ -182,7 +182,19 @@ public class PlexWebhook : BaseController
         if (episodeType != EpisodeType.Episode ||
             metadata.Index == 0) //metadata.index = 0 when it's something else.
         {
-            return (null, anime); //right now no clean way to detect the episode. I could do by title.
+            var nameMatches = anime
+                .GetAnimeEpisodes().Where(a => a.AniDB_Episode != null)
+                .Where(a => a.EpisodeTypeEnum == episodeType)
+                .Where(a => a.Title.Equals(metadata.Title))
+                .Where(a => a.AniDB_Episode.EpisodeNumber == episodeNumber
+                    || (a.TvDBEpisode?.SeasonNumber == series && a.TvDBEpisode?.EpisodeNumber == episodeNumber)).ToList();
+
+            if (nameMatches.Count == 1) return (nameMatches.First(), anime);
+
+            _logger.LogInformation(
+            $"Unable to work out the metadata for {metadata.Guid} using title {metadata.Title}.");
+
+            return (null, anime);
         }
 
 
@@ -204,7 +216,7 @@ public class PlexWebhook : BaseController
 
         //catch all
         _logger.LogInformation(
-            $"Unable to work out the metadata for {metadata.Guid}, this might be a clash of multipl episodes linked, but no tvdb link.");
+            $"Unable to work out the metadata for {metadata.Guid}, this might be a clash of multiple episodes linked, but no tvdb link.");
         return (null, anime);
     }
 
--- a/Shoko.Server/API/v2/Modules/PlexWebhook.cs
+++ b/Shoko.Server/API/v2/Modules/PlexWebhook.cs
@@ -14,6 +14,7 @@ using Shoko.Commons.Extensions;
 using Shoko.Models.Enums;
 using Shoko.Models.Plex.Collection;
 using Shoko.Models.Plex.Connections;
+using Shoko.Models.Plex.Libraries;
 using Shoko.Models.Plex.TVShow;
 using Shoko.Models.Server;
 using Shoko.Server.API.v2.Models.core;
@@ -98,7 +99,7 @@ public class PlexWebhook : BaseController
     private void TraktScrobble(PlexEvent evt, ScrobblePlayingStatus type, JMMUser user)
     {
         var metadata = evt.Metadata;
-        var (episode, _) = GetEpisode(metadata, user);
+        var (episode, _) = GetEpisodeForUser(metadata, user);
 
         if (episode == null) return;
 
@@ -120,7 +121,7 @@ public class PlexWebhook : BaseController
     private void Scrobble(PlexEvent data, SVR_JMMUser user)
     {
         var metadata = data.Metadata;
-        var (episode, anime) = GetEpisode(metadata, user);
+        var (episode, anime) = GetEpisodeForUser(metadata, user);
         if (episode == null)
         {
             _logger.LogInformation(
@@ -140,7 +141,7 @@ public class PlexWebhook : BaseController
     #endregion
 
     [NonAction]
-    private (SVR_AnimeEpisode, SVR_AnimeSeries) GetEpisode(PlexEvent.PlexMetadata metadata, JMMUser user)
+    private (SVR_AnimeEpisode, SVR_AnimeSeries) GetEpisodeForUser(PlexEvent.PlexMetadata metadata, JMMUser user)
     {
         var guid = new Uri(metadata.Guid);
         if (guid.Scheme != "com.plexapp.agents.shoko" && guid.Scheme != "com.plexapp.agents.shokorelay")
--- a/Shoko.Server/API/v2/Modules/PlexWebhook.cs
+++ b/Shoko.Server/API/v2/Modules/PlexWebhook.cs
@@ -194,19 +194,16 @@ public class PlexWebhook : BaseController
         {
             if (episode == null)
             {
-                _logger.LogInformation(
-                    $"Failed to get anime episode from plex using key {metadata.Key}.");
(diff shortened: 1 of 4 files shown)


Shoko.Server/API/v2/Modules/PlexWebhook.cs
```diff
@@ -182,7 +182,19 @@ private void Scrobble(PlexEvent data, SVR_JMMUser user)
         if (episodeType != EpisodeType.Episode ||
             metadata.Index == 0) //metadata.index = 0 when it's something else.
         {
-            return (null, anime); //right now no clean way to detect the episode. I could do by title.
+            var nameMatches = anime
+                .GetAnimeEpisodes().Where(a => a.AniDB_Episode != null)
+                .Where(a => a.EpisodeTypeEnum == episodeType)
+                .Where(a => a.Title.Equals(metadata.Title))
+                .Where(a => a.AniDB_Episode.EpisodeNumber == episodeNumber
+                    || (a.TvDBEpisode?.SeasonNumber == series && a.TvDBEpisode?.EpisodeNumber == episodeNumber)).ToList();
```
Cazzar, line 190: Not a huge fan of this - the reason is doing this "a or b" could cause erroneous matches.

Shoko.Server/API/v2/Modules/PlexWebhook.cs
```diff
@@ -182,7 +182,19 @@ private void Scrobble(PlexEvent data, SVR_JMMUser user)
         if (episodeType != EpisodeType.Episode ||
             metadata.Index == 0) //metadata.index = 0 when it's something else.
         {
-            return (null, anime); //right now no clean way to detect the episode. I could do by title.
+            var nameMatches = anime
+                .GetAnimeEpisodes().Where(a => a.AniDB_Episode != null)
+                .Where(a => a.EpisodeTypeEnum == episodeType)
+                .Where(a => a.Title.Equals(metadata.Title))
```
Cazzar, line 188: What about the situation where a user changes the episode title?

Shoko.Server/API/v2/Modules/PlexWebhook.cs
```diff
@@ -179,10 +184,31 @@ private void Scrobble(PlexEvent data, SVR_JMMUser user)
                 break;
         }
 
+        var key = metadata.Key.Split("/").LastOrDefault();  // '/library/metadata/{key}'
+
+        var plexEpisode = (SVR_Episode)GetPlexEpisodeData(key, user);
+        var episode = plexEpisode?.AnimeEpisode;
```
Cazzar, line 190: Can we add this with a guard against people not having a valid plex connection?

The original design was that the webhook did not require user plex login, so it was entirely working off it's own data.

Happy to add this in, but I think it should be at least guarded to allow for users to not have their own plex config.

Shoko.Server/API/v2/Modules/PlexWebhook.cs
```diff
@@ -179,10 +184,31 @@ private void Scrobble(PlexEvent data, SVR_JMMUser user)
...
+
         if (episodeType != EpisodeType.Episode ||
             metadata.Index == 0) //metadata.index = 0 when it's something else.
         {
-            return (null, anime); //right now no clean way to detect the episode. I could do by title.
+            if (episode == null)
+            {
+                _logger.LogInformation(
+                    $"Failed to get anime episode from plex using key {metadata.Key}.");
```
Cazzar, line 198:
```suggestion
                _logger.LogInformation($"Failed to get anime episode from plex using key {metadata.Key}.");
```

Code style - there's no real need to have it on multiple lines here.

Shoko.Server/API/v2/Modules/PlexWebhook.cs
```diff
@@ -179,10 +184,31 @@ private void Scrobble(PlexEvent data, SVR_JMMUser user)
...
+                return (null, anime);
+            }
+
+            if (episode.EpisodeTypeEnum == episodeType
+                && anime.GetAnimeEpisodes().Contains(episode))
```
Cazzar, line 203:
```suggestion
            if (episode.EpisodeTypeEnum == episodeType && anime.GetAnimeEpisodes().Contains(episode))
```
Code style again.

Shoko.Server/API/v2/Modules/PlexWebhook.cs
```diff
@@ -179,10 +184,31 @@ private void Scrobble(PlexEvent data, SVR_JMMUser user)
...
+            {
+                return (episode, anime);
+            }
+
+            _logger.LogInformation(
+            $"Unable to work out the metadata for {metadata.Guid}.");
```
Cazzar, line 209:
```suggestion
            _logger.LogInformation($"Unable to work out the metadata for {metadata.Guid}.");
```

Shoko.Server/Repositories/Cached/AnimeEpisodeRepository.cs
```diff
@@ -56,8 +56,9 @@ public SVR_AnimeEpisode GetByAniDBEpisodeID(int epid)
     /// Get the AnimeEpisode 
     /// </summary>
     /// <param name="name">The filename of the anime to search for.</param>
+    /// <param name="size">The size of the file in bytes</param>
     /// <returns>the AnimeEpisode given the file information</returns>
-    public SVR_AnimeEpisode GetByFilename(string name)
+    public SVR_AnimeEpisode GetByFilename(string name, long? size = null)
```
Cazzar, line 61: Consider renaming this, or adding an extra helper to make this clearer.

Shoko.Server/Plex/Collection/SVR_PlexLibrary.cs
```diff
@@ -23,4 +23,13 @@ public Episode[] GetEpisodes()
             .DeserializeObject<MediaContainer<MediaContainer>>(data, Helper.SerializerSettings)
             .Container.Metadata;
     }
+
+    public Episode[] GetEpisode(string ratingKey)
+    {
+        var (_, data) = Helper.RequestFromPlexAsync($"/library/metadata/{ratingKey}").GetAwaiter()
+            .GetResult();
```
Cazzar, line 30: https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.getawaiter?view=net-8.0#remarks

generally speaking you shouldn't really be using `.GetAwaiter().GetResult()`

Shoko.Server/API/v2/Modules/PlexWebhook.cs
```diff
@@ -273,7 +299,7 @@ public async Task<ActionResult> SyncForUser(int id)
 
     [Authorize]
     [HttpGet("libraries")]
-    public ActionResult<Directory[]> GetLibraries()
+    public ActionResult<Shoko.Models.Plex.Libraries.Directory[]> GetLibraries()
```
Cazzar, line 302:
```suggestion
    public ActionResult<Directory[]> GetLibraries()
```

Use an import, or if there is a clash, an alias.

Cazzar (changes requested):

Shoko.Server/Repositories/Cached/AnimeEpisodeRepository.cs
LordFlashmeow: Should I split it into 2 functions, `GetByFilename(string name)` and `GetByFilenameAndSize(string name, long size)`?

Shoko.Server/API/v2/Modules/PlexWebhook.cs
LordFlashmeow: Can you clarify the scenario you're describing? The user has plex webhooks enabled, but hasn't linked their plex account to Shoko?

If that's the case, I suppose I could fall back original behavior of only matching episodes.

Shoko.Server/API/v2/Modules/PlexWebhook.cs
```diff
@@ -60,20 +62,29 @@ public ActionResult WebhookPost([FromForm] [ModelBinder(BinderType = typeof(Plex
         }
 
         _logger.LogTrace($"{payload.Event}: {payload.Metadata?.Guid}");
+
+        var user = User;
+        user ??= RepoFactory.JMMUser.GetAll().FirstOrDefault(u => payload.Account.Title.FindIn(u.GetPlexUsers()));
```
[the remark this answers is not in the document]
LordFlashmeow: How would that be done? (Sorry, I haven't worked with C# before)

ElementalCrisis:
@LordFlashmeow can you update your PR please and resolve conflicts.

ElementalCrisis:
@LordFlashmeow I'm going to close this PR since there's been no activity on it for a long time...

PR #1127, opened 2024-05-13
Fix: Get release groups correctly for WebUI

fearnlj01:
Ignore the typo amendments, the WebUIController one was killing me every time I opened swagger.

Fix is to make sure the file summary endpoint gets group ids as expected.

### DIFF
--- a/Shoko.Server/API/v3/Models/Shoko/WebUI.cs
+++ b/Shoko.Server/API/v3/Models/Shoko/WebUI.cs
@@ -201,7 +201,7 @@ public class WebUI
                 .DistinctBy(anidbFile => anidbFile.Hash)
                 .ToDictionary(anidbFile => anidbFile.Hash);
             var releaseGroups = anidbFiles.Values
-                .Select(anidbFile => anidbFile.ReleaseGroup.GroupID)
+                .Select(anidbFile => anidbFile.GroupID)
                 .Distinct()
                 .Where(groupID => groupID != 0)
                 .Select(RepoFactory.AniDB_ReleaseGroup.GetByGroupID)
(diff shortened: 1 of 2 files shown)


fearnlj01:
I am also confirming that I have tested this, so it does indeed actually work as far as per my own expectations.

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
```diff
@@ -201,10 +201,10 @@ public WebUISeriesFileSummary(SVR_AnimeSeries series, HashSet<EpisodeType> episo
                 .DistinctBy(anidbFile => anidbFile.Hash)
                 .ToDictionary(anidbFile => anidbFile.Hash);
             var releaseGroups = anidbFiles.Values
-                .Select(anidbFile => anidbFile.AniDB_FileID)
+                .Select(anidbFile => anidbFile.ReleaseGroup.GroupID)
```
revam, line 204: If you're going to use the object helper instead of the id on the anidb file model, then you might as well rewrite it to select the model here, then distinct by the group id, then filter on the model for a group id of 0, then drop the select step after that.

revam: the alternative is to leave most of it as is, and just select the group id on the anidb file model so we don't do the group call early (and save a lot of repo lookups like we do currently).

revam: actually, lets go with that, since it will save on db calls (or more accurately, won't introduce _more_ db calls, even if we're just hitting the cache).

revam (commented):
:+1:

PR #1128, opened 2024-05-13
Feat: Add overview property to the webui's filesummary endpoint

fearnlj01:
The motivation for this addition is to ensure that, regardless of what the WebUI requests, we always have the same information available for the file overview panel.

I've tested the changes on my end, but I've not got much mixed media, so I cannot explicitly testify to the robustness of the changes, just that it works on 'typical' & empty series...

Separately, I have provisional changes ready for the series redesign PR for the web ui, so the changes in this PR should actually be utilised if everything on that end gets merged.

### DIFF
--- a/Shoko.Server/API/v3/Models/Shoko/WebUI.cs
+++ b/Shoko.Server/API/v3/Models/Shoko/WebUI.cs
@@ -208,7 +208,7 @@ public class WebUI
                 .Where(releaseGroup => releaseGroup != null)
                 .ToDictionary(releaseGroup => releaseGroup.GroupID);
             // We only care about files with exist and have actual media info and with an actual physical location. (Which should hopefully exclude nothing.)
-            var files = crossRefs
+            var filesWithXrefAndLocation = crossRefs
                 .Where(xref => episodes.ContainsKey(xref.EpisodeID))
                 .Select(xref =>
                 {
@@ -216,6 +216,14 @@ public class WebUI
                     var location = file?.GetBestVideoLocalPlace();
                     if (file?.Media == null || location == null)
                         return null;
+
+                    return new { file, xref, location };
+                })
+                .Where(f => f != null);
+            var files = filesWithXrefAndLocation
+                .Select(fileWithEpisode =>
+                {
+                    var (file, xref, location) = (fileWithEpisode.file, fileWithEpisode.xref, fileWithEpisode.location);
 
                     var media = new MediaInfo(file, file.Media);
                     var episode = episodes[xref.EpisodeID];
@@ -383,7 +391,42 @@ public class WebUI
                 .ThenBy(episode => episode.Number)
                 .Select(episode => new Episode.AniDB(episode.AniDB))
                 .ToList();
+
+
+            var releaseGroupNames = releaseGroups.Values
+                .Select(group => group.GroupNameShort)
+                .ToList();
+
+            var sourcesByType = filesWithXrefAndLocation
+                .Select(f => new
+                {
+                    f.file,
+                    EpisodeType = episodes[f.xref.EpisodeID].Type,
+                    FileSource = (CrossRefSource)f.xref.CrossRefSource == CrossRefSource.AniDB && anidbFiles.ContainsKey(f.xref.Hash) ? anidbFiles[f.xref.Hash].File_Source : null
+                })
+                .Where(f => f != null)
+                .GroupBy(f => f.EpisodeType)
+                .Select(fileGroup =>
+                {
+                    var sources = fileGroup
+                        .GroupBy(f => f.FileSource)
+                        .Select(f => new SourceGrouping { Type = f.Key, Count = f.Count() })
+                        .ToList();
+                    return new SourcesByType { Type = fileGroup.Key, Sources = sources };
+                })
+                .ToList();
+
+            var totalFileSize = files
+                .Sum(fileWrapper => fileWrapper.Episode.Size);
+            Overview = new FileSummaryOverview()
+            {
+                ReleaseGroups = releaseGroupNames,
+                SourcesByType = sourcesByType,
+                TotalFileSize = totalFileSize,
+            };
         }
+
+        public FileSummaryOverview Overview;
 
         public List<EpisodeGroupSummary> Groups;
 
@@ -677,6 +720,44 @@ public class WebUI
             /// ED2K File Hash.
             /// </summary>
             public string ED2K;
+        }
+
+        public class FileSummaryOverview
+        {
+            /// <summary>
+            /// The list of all AniDB episode sources, and their associated file counts
+            /// </summary>
+            public List<SourcesByType> SourcesByType;
+
+            /// <summary>
+            /// The total size of all locally available files for a series
+            /// </summary>
+            public long TotalFileSize;
+
+            /// <summary>
+            /// A summarised list of all the locally available release groups for a series.
+            /// </summary>
+            public List<string> ReleaseGroups;
+        }
+
+        public class SourcesByType
+        {
+            /// <summary>
+            /// The type of episode.
+            /// </summary>
+            [JsonConverter(typeof(StringEnumConverter))]
+            public EpisodeType Type;
+
+            /// <summary>
+            /// The source of the file for the episode
+            /// </summary>
+            public List<SourceGrouping> Sources;
+        }
+
+        public class SourceGrouping
+        {
+            public string Type;
+            public int Count;
         }
 
         /// <summary>
--- a/Shoko.Server/API/v3/Models/Shoko/WebUI.cs
+++ b/Shoko.Server/API/v3/Models/Shoko/WebUI.cs
@@ -214,16 +214,15 @@ public class WebUI
                 {
                     var file = RepoFactory.VideoLocal.GetByHash(xref.Hash);
                     var location = file?.GetBestVideoLocalPlace();
-                    if (file?.Media == null || location == null)
-                        return null;
 
-                    return new { file, xref, location };
+                    return (file, xref, location);
                 })
-                .Where(f => f != null);
+                .Where(t => t.file?.Media != null && t.location != null)
+                .ToList();
             var files = filesWithXrefAndLocation
-                .Select(fileWithEpisode =>
+                .Select(tuple =>
                 {
-                    var (file, xref, location) = (fileWithEpisode.file, fileWithEpisode.xref, fileWithEpisode.location);
+                    var (file, xref, location) = tuple;
 
                     var media = new MediaInfo(file, file.Media);
                     var episode = episodes[xref.EpisodeID];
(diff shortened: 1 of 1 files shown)


Shoko.Server/API/v3/Models/Shoko/WebUI.cs
```diff
@@ -217,6 +217,14 @@ public WebUISeriesFileSummary(SVR_AnimeSeries series, HashSet<EpisodeType> episo
                     if (file?.Media == null || location == null)
                         return null;
 
+                    return new { file, xref, location };
+                })
+                .Where(f => f != null);
+            var files = filesWithXrefAndLocation
```
fearnlj01, line 223: I've got no idea as to how expensive retreiving the SVR_VideoLocal & SVR_VideoLocal_Place records are, so I've gone down the 'safer' route of reusing these records.

If this is unnecessary, I'm more than willing to amend and re-request everything for the overview calculation rather than modifying this section.

harshithmohan:
Can you what it looks like on the webui?

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
```diff
@@ -217,6 +217,14 @@ public WebUISeriesFileSummary(SVR_AnimeSeries series, HashSet<EpisodeType> episo
                     if (file?.Media == null || location == null)
                         return null;
 
+                    return new { file, xref, location };
+                })
+                .Where(f => f != null);
```
revam, line 222: at least add .ToList() after the .Where() to execute the query amd save the results in memory, so we don't have to run it twice. 😔

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
```diff
@@ -679,6 +722,44 @@ public class EpisodeDetails
...
+
+            /// <summary>
+            /// The source of the file for the episode
+            /// </summary>
+            public List<SourceGrouping> Sources;
+        }
+
+        public class SourceGrouping
+        {
+            public string Type;
```
revam, line 759: no doc. comments here? 🫠

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
```diff
@@ -217,6 +217,14 @@ public WebUISeriesFileSummary(SVR_AnimeSeries series, HashSet<EpisodeType> episo
                     if (file?.Media == null || location == null)
                         return null;
 
+                    return new { file, xref, location };
+                })
+                .Where(f => f != null);
+            var files = filesWithXrefAndLocation
+                .Select(fileWithEpisode =>
+                {
+                    var (file, xref, location) = (fileWithEpisode.file, fileWithEpisode.xref, fileWithEpisode.location);
```
revam, line 226: you can use a tuplenin the .Select() above to get deconstruction working here (again)

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
```diff
@@ -217,6 +217,14 @@ public WebUISeriesFileSummary(SVR_AnimeSeries series, HashSet<EpisodeType> episo
                     if (file?.Media == null || location == null)
                         return null;
 
+                    return new { file, xref, location };
```
revam, line 220: replace with `return (file, xref, location);` to get deconstructing working (again).

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
```diff
@@ -383,8 +391,43 @@ public WebUISeriesFileSummary(SVR_AnimeSeries series, HashSet<EpisodeType> episo
...
+                .ToList();
+
+            var sourcesByType = filesWithXrefAndLocation
+                .Select(f => new
+                {
+                    f.file,
+                    EpisodeType = episodes[f.xref.EpisodeID].Type,
+                    FileSource = (CrossRefSource)f.xref.CrossRefSource == CrossRefSource.AniDB && anidbFiles.ContainsKey(f.xref.Hash) ? anidbFiles[f.xref.Hash].File_Source : null
+                })
+                .Where(f => f != null)
```
revam, line 407: f won't ever be null here

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
```diff
@@ -383,8 +391,43 @@ public WebUISeriesFileSummary(SVR_AnimeSeries series, HashSet<EpisodeType> episo
...
+
+
+            var releaseGroupNames = releaseGroups.Values
+                .Select(group => group.GroupNameShort)
+                .ToList();
+
+            var sourcesByType = filesWithXrefAndLocation
+                .Select(f => new
+                {
+                    f.file,
```
revam, line 403: why 8snf.file included here if it's not used? am i missing something?

revam (commented):
that was all for now from me.

fearnlj01:
> Can you what it looks like on the webui?

Just to quickly reply to this as it's a quick one... (I'll pick up on the reset when I'm back home)

As pictured below, this shouldn't change how anything is presented in the WebUI (But always stays the same, even if not grouping by ReleaseGroup/FileSource...).
image


An example response is as per the below:
```json
{
  "Overview": {
    "SourcesByType": [
      {
        "Type": "Normal",
        "Sources": [
          { "Type": "Web", "Count": 11 },
          { "Type": "Unknown", "Count": 1 }
        ]
      }
    ],
    "TotalFileSize": 17355339539,
    "ReleaseGroups": ["SomeAnimeGroup"]
  },
  "Groups": [
    {
      "GroupName": "SomeAnimeGroup",
      "GroupNameShort": "SomeAnimeGroup",
      "FileVersion": 1,
      "FileLocation": "F:\\Anime\\SomeAnime2",
      "VideoResolution": "1080p",
      "VideoWidth": 1920,
      "VideoHeight": 1080,
      "AudioLanguages": ["jpn"],
      "SubtitleLanguages": ["eng"],
      "RangeByType": {
        "Normal": { "Count": 12, "Range": "01-12", "FileSize": 17355339539 }
      }
    }
  ],
  "MissingEpisodes": []
}
```

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
revam: I didn't see this comment earlier, but yo answer; it's fine to do what you did… as long as you cache the result and don't run the query twice/thrice/etc.. If you append `.ToList()` after the last LINQ operation (the `.Where(…)` above in this case) then it will execute the query and turn it into a list which can more easily be reused for future LINQ query operations,

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
```diff
@@ -383,8 +391,43 @@ public WebUISeriesFileSummary(SVR_AnimeSeries series, HashSet<EpisodeType> episo
                 .ThenBy(episode => episode.Number)
                 .Select(episode => new Episode.AniDB(episode.AniDB))
                 .ToList();
+
+
+            var releaseGroupNames = releaseGroups.Values
+                .Select(group => group.GroupNameShort)
```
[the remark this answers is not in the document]
revam:
```suggestion
                .Select(group => group.GroupName)
```

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
revam:
```suggestion
                    return (file, xref, location);
```

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
revam:
```suggestion
                .Where(tuple => tuple.file != null);
```

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
```diff
@@ -217,6 +217,14 @@ public WebUISeriesFileSummary(SVR_AnimeSeries series, HashSet<EpisodeType> episo
                     if (file?.Media == null || location == null)
                         return null;
 
+                    return new { file, xref, location };
+                })
+                .Where(f => f != null);
+            var files = filesWithXrefAndLocation
+                .Select(fileWithEpisode =>
```
revam, line 224:
```suggestion
                .Select(tuple =>
```

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
revam:
```suggestion
                    var (file, xref, location) = tuple;
```

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
revam:
```suggestion
```

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
revam:
```suggestion
```

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
revam:
```suggestion
            /// <summary>
            /// The file source.
            /// </summary>
            [JsonConverter(typeof(StringEnumConverter))]
            public FileSource Source;
```

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
```diff
@@ -679,6 +722,44 @@ public class EpisodeDetails
...
+            /// <summary>
+            /// The source of the file for the episode
+            /// </summary>
+            public List<SourceGrouping> Sources;
+        }
+
+        public class SourceGrouping
+        {
+            public string Type;
+            public int Count;
```
revam, line 760:
```suggestion
           /// <summary>
           /// Amount of files with this file source.
           /// </summary>
           public int Count;
```

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
```diff
@@ -383,8 +391,43 @@ public WebUISeriesFileSummary(SVR_AnimeSeries series, HashSet<EpisodeType> episo
...
+            var releaseGroupNames = releaseGroups.Values
+                .Select(group => group.GroupNameShort)
+                .ToList();
+
+            var sourcesByType = filesWithXrefAndLocation
+                .Select(f => new
+                {
+                    f.file,
+                    EpisodeType = episodes[f.xref.EpisodeID].Type,
+                    FileSource = (CrossRefSource)f.xref.CrossRefSource == CrossRefSource.AniDB && anidbFiles.ContainsKey(f.xref.Hash) ? anidbFiles[f.xref.Hash].File_Source : null
```
revam, line 405:
```suggestion
                   FileSource = (CrossRefSource)f.xref.CrossRefSource == (int)CrossRefSource.AniDB && anidbFiles.ContainsKey(f.xref.Hash) ? File.ParseFileSource(anidbFiles[f.xref.Hash].File_Source) : FileSource.Unknown,
```

revam (commented):
A few suggestions I would had done if i were writing this.

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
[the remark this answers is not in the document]
fearnlj01: Originally I was planning to send both `GroupName` & `GroupNameShort`... However, when I took a second look at how it ends up being displayed [in the WebUI](https://github.com/ShokoAnime/Shoko-WebUI/blob/8cd9fc7c5319563306a489ba94cfe9e15a3412c9/src/pages/collection/series/SeriesFileSummary.tsx#L149), I decided that given that the overview only shows the short name anyway, there was no need to complicate things any further.

Shoko.Server/API/v3/Models/Shoko/WebUI.cs
fearnlj01: "Oops". Spent a while going in circles to get this all to work and this is a remnant of my botched housekeeping.

fearnlj01:
I've not gone 100% inline with feedback, but should've addressed everything whilst making things a bit more readable in the process.

ElementalCrisis:
@revam review and merge please if everything is good to go.

revam (approved):

PR #1129, opened 2024-05-19
Implement notify system from AniDB UDP API

Terrails:
Implements the AniDB [notify system](https://wiki.anidb.net/UDP_API_Definition#Notify_Commands) which will be useful to read and handle file moved messages.

There are 2 new tables:
- `AniDB_NotifyQueue` used to add any new messages and notifications returned by `NOTIFYLIST`
- `AniDB_Message` used to store message content

Currently there is no need for something like a `AniDB_Notification` table as there is no real use for it. Unread notifications are only added to the `AniDB_NotifyQueue` table and there is nothing that fetches further notification content at the moment.

And 3 new config options:
- `Notification_UpdateFrequency` used to configure fetch interval (is set to `ScheduledUpdateFrequency.Never` by default). It would be useful to add shorter intervals to the `ScheduledUpdateFrequency` enum in `Shoko.Models` since the minimum is 6 hours. A 1 hour option would most likely be ideal since the API does say it can be fetched at as low as every 20 mins.
- `Notification_Acknowledge` if fetched notifications should be automagically marked as read upon retrieval.
- `Notification_HandleMovedFiles` if `moved file: <fileId>` messages should force an update on the file.

`GetAniDBNotifyJob` should run in the configured intervals or when the new  `/Action/GetAniDBNotifications/` v3 API endpoint is ran. ~~This PR just lays down the groundwork for fetching and saving unread message content to the database. There is already a `IsFileMoved` flag in `AniDB_Message` that will be used for future behavior.~~ It includes the logic to refetch AniDB data for a moved file now.


Shoko.Server/Databases/MySQL.cs
```diff
@@ -247,7 +247,11 @@ public class MySQL : BaseDatabase<MySqlConnection>
...
+            "CREATE TABLE `CrossRef_AniDB_Trakt` ( `CrossRef_AniDB_TraktID` INT NOT NULL AUTO_INCREMENT, `AnimeID` int NOT NULL, `TraktID` varchar(100) character set utf8, `TraktSeasonNumber` int NOT NULL, `CrossRefSource` int NOT NULL, PRIMARY KEY (`CrossRef_AniDB_TraktID`) ) ; "),
+        new DatabaseCommand(1, 111,
+            "CREATE TABLE `AniDB_NotifyQueue` ( `AniDB_NotifyQueueID` INT NOT NULL AUTO_INCREMENT, `Type` int NOT NULL, `ID` int NOT NULL, `Added` datetime NOT NULL, PRIMARY KEY (`AniDB_NotifyQueueID`) ) ; "),
+        new DatabaseCommand(1, 112,
+            "CREATE TABLE `AniDB_Message` ( `AniDB_MessageID` INT NOT NULL AUTO_INCREMENT, `MessageID` int NOT NULL, `FromUserID` int NOT NULL, `FromUserName` varchar(100) character set utf8 NOT NULL, `Date` datetime NOT NULL, `Type` int NOT NULL, `Title` text character set utf8 NOT NULL, `Body` text character set utf8 NOT NULL, `Flags` int NOT NULL, PRIMARY KEY (`AniDB_MessageID`) ) ;"),
```
Cazzar, line 254: Remind me @da3dsoul shouldn't these be in the `patchCommands` array with the `RequiredVersion` updated?

IIRC these are only ran on a new install, not an upgrade.

Shoko.Server/Models/AniDB_Message.cs
```diff
@@ -0,0 +1,72 @@
...
+    public string FromUserName { get; set; }
+    public DateTime Date { get; set; }
+    public AniDBMessageType Type { get; set; }
+    public string Title { get; set; }
+    public string Body { get; set; }
+    public AniDBMessageFlags Flags { get; set; }
+
+    #endregion
+
+    #region Disabled Auto Matching
```
revam, line 23: at least rename the region if you're going to copy paste :pensive:

Shoko.Server/Scheduling/Jobs/AniDB/GetAniDBMessageJob.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        // add flag if its a file moved system message
+        if (message.Type == AniDBMessageType.System && message.Title.ToLower().StartsWith("file moved:"))
+        {
+            message.IsFileMoved = true;
+        }
+
+        // acknowledge if enabled
+        var settings = _settingsProvider.GetSettings();
+        if (settings.AniDb.Notification_Acknowledge)
```
revam, line 71: can you fork this out to a separate job, then just schedule that job so we're not doing more than needed UDP commands per job? unless @da3dsoul means it's better to keep it in this job.

Shoko.Server/Scheduling/Jobs/AniDB/GetAniDBMessageJob.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            var requestAck = _requestFactory.Create<RequestAcknowledgeNotify>(
+                r =>
+                {
+                    r.Type = AniDBNotifyType.Message;
+                    r.ID = MessageID;
+                }
+            );
+            var responseAck = requestAck.Send();
```
revam, line 80: also the message for the db might never be saved to disk if it throws here because it detects a ban.

revam (commented):
that's all from me for now.

Shoko.Server/Databases/MySQL.cs
harshithmohan: Yep

revam:
Also, before I forget. We might need a guard to prevent _too often_ requests for the endpoint, since iirc it had a rate limit of 20 (or was it 30) minutes per request, and we want shoko to adhered to that rule.

revam:
@Terrails can you add the job to act on the File Moved flag to this PR? And also an option to auto schedule it if a setting is set, and an endpoint to schedule all files that have been moved but not acted on yet from an action endpoint?

Terrails:
> Also, before I forget. We might need a guard to prevent _too often_ requests for the endpoint, since iirc the endpoint have a rate limit of 20 minutes per request, and we want shoko to adhered to that rule.
> 
> Edit: _Wiki says 20_.

Shouldn't that already be somewhat handled by https://github.com/ShokoAnime/ShokoServer/blob/906219e999809c1e138b792c3b1085d6355928c0/Shoko.Server/Services/ActionService.cs#L940-L962
It does stop it from running unless it was forced, which happens only when the API endpoint is manually executed. The automatic scheduling should not run more often than configured as the forced refresh flag is false.
https://github.com/ShokoAnime/ShokoServer/blob/906219e999809c1e138b792c3b1085d6355928c0/Shoko.Server/Server/ShokoServer.cs#L343

Terrails:
> @Terrails can you add the job to act on the File Moved flag to this PR? And also an option to auto schedule it if a setting is set, and an endpoint to schedule all files that have been moved but not acted on yet from an action endpoint?

Should be implemented now. I assume that `ProcessFileJob` has to be executed to fetch the new file data
https://github.com/ShokoAnime/ShokoServer/blob/aacf117c7825a0fb39a97e26fb8b4c45199f7ed0/Shoko.Server/Scheduling/Jobs/Shoko/ProcessFileMovedMessageJob.cs#L43-L56
And then if it is successful it is flagged as handled.
The job also has a new endpoint now and should automatically run after new notifications are fetched from AniDB
https://github.com/ShokoAnime/ShokoServer/blob/aacf117c7825a0fb39a97e26fb8b4c45199f7ed0/Shoko.Server/Services/ActionService.cs#L965-L967
It might be a good idea to add a force flag to the API endpoint though.

Shoko.Server/Mappings/AniDB_MessageMap.cs
```diff
@@ -0,0 +1,24 @@
...
+    public AniDB_MessageMap()
+    {
+        Table("AniDB_Message");
+        Not.LazyLoad();
+        Id(x => x.AniDB_MessageID);
+
+        Map(x => x.MessageID).Not.Nullable();
+        Map(x => x.FromUserId).Not.Nullable();
+        Map(x => x.FromUserName).Not.Nullable();
+        Map(x => x.Date).Not.Nullable();
```
revam, line 18: Please rename the field to be less generic (E.g. "SentAt", "ReceivedAt", etc.). Also please also add a date for when it was fetched by Shoko ("E.g. "FetchedAt" or "CreatedAt", etc.).

Shoko.Server/Models/AniDB_Message.cs
```diff
@@ -0,0 +1,86 @@
...
+
+public class AniDB_Message
+{
+    #region DB Columns
+
+    public int AniDB_MessageID { get; set; }
+    public int MessageID { get; set; }
+    public int FromUserId { get; set; }
+    public string FromUserName { get; set; }
+    public DateTime Date { get; set; }
```
revam, line 14: See comment above.

revam (changes requested):

revam (approved):

revam (approved):

PR #1132, opened 2024-06-03
fix: import directory being deleted

Terrails:
Due to `ReturnSpecialDirectories` being set to `true`, the `IEnumerable` contained `\\.` and `\\..` for current and all subdirectories. This resulted in the path with `\\.` at the end not being seen as equal as `dropFolder.ImportFolderLocation` in `RecursiveDeleteEmptyDirectories`. Simply setting this flag to `false` should resolve this and another issue like duplicate subdirectory entries. Also made the delete function run only when the `IEnumerable` is not empty.

### DIFF
--- a/Shoko.Server/Services/VideoLocal_PlaceService.cs
+++ b/Shoko.Server/Services/VideoLocal_PlaceService.cs
@@ -410,9 +410,10 @@ public class VideoLocal_PlaceService
 
             if (dropFolder.IsDropSource == 1 && deleteEmpty)
             {
-                var directories = dropFolder.BaseDirectory.EnumerateDirectories("*", new EnumerationOptions() { RecurseSubdirectories = true, ReturnSpecialDirectories = true, IgnoreInaccessible = true })
+                var directories = dropFolder.BaseDirectory.EnumerateDirectories("*", new EnumerationOptions() { RecurseSubdirectories = true, ReturnSpecialDirectories = false, IgnoreInaccessible = true })
                     .Select(dirInfo => dirInfo.FullName);
-                RecursiveDeleteEmptyDirectories(directories, dropFolder.ImportFolderLocation);
+                if (directories.Any())
+                    RecursiveDeleteEmptyDirectories(directories, dropFolder.ImportFolderLocation);
             }
 
             ShokoEventHandler.Instance.OnFileMoved(dropFolder, destFolder, oldPath, newFilePath, videoLocalPlace);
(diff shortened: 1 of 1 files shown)


Shoko.Server/Services/VideoLocal_PlaceService.cs
```diff
@@ -410,9 +410,10 @@ public MoveFileResult MoveFile(SVR_VideoLocal_Place videoLocalPlace, bool delete
 
             if (dropFolder.IsDropSource == 1 && deleteEmpty)
             {
-                var directories = dropFolder.BaseDirectory.EnumerateDirectories("*", new EnumerationOptions() { RecurseSubdirectories = true, ReturnSpecialDirectories = true, IgnoreInaccessible = true })
+                var directories = dropFolder.BaseDirectory.EnumerateDirectories("*", new EnumerationOptions() { RecurseSubdirectories = true, ReturnSpecialDirectories = false, IgnoreInaccessible = true })
                     .Select(dirInfo => dirInfo.FullName);
-                RecursiveDeleteEmptyDirectories(directories, dropFolder.ImportFolderLocation);
+                if (directories.Any())
```
revam, line 415: Revert the added if statement. we don't want to run the enumerator twice if we can avoid it, and the internals of the method can handle an empty enunerator just fine.

Shoko.Server/Services/VideoLocal_PlaceService.cs
```diff
@@ -410,9 +410,10 @@ public MoveFileResult MoveFile(SVR_VideoLocal_Place videoLocalPlace, bool delete
 
             if (dropFolder.IsDropSource == 1 && deleteEmpty)
             {
-                var directories = dropFolder.BaseDirectory.EnumerateDirectories("*", new EnumerationOptions() { RecurseSubdirectories = true, ReturnSpecialDirectories = true, IgnoreInaccessible = true })
+                var directories = dropFolder.BaseDirectory.EnumerateDirectories("*", new EnumerationOptions() { RecurseSubdirectories = true, ReturnSpecialDirectories = false, IgnoreInaccessible = true })
```
revam, line 413: Just remove the option if it's false by default. (Do check please, since i can't right now.)

revam (changes requested):

Shoko.Server/Services/VideoLocal_PlaceService.cs
Terrails: The docs do not mention anything but a quick run without the option confirmed its false by default.

revam (approved):

PR #1145, opened 2024-07-16
Fix group refreshing for good

cxooda:
Attempting to delete the temp group here would throw NHibernate.StaleStateException : Unexpected row count: 0; expected: 1

That is because it's not being added to the db in the first place, or as line 76 in AnimeGroupCreator.cs states: "We won't use AnimeGroupRepository.Save because we don't need to perform all the extra stuff since this is for temporary use only"


cxooda:
closed because I didn't realize this was already fixed at the time
https://github.com/ShokoAnime/ShokoServer/commit/09b0dbad93769fc603cf94fe3186f192f502f178

PR #1146, opened 2024-07-16
Fix group filter watch states not corresponding to actual watch states

cxooda:


ElementalCrisis:
@revam please review as it seems @da3dsoul is busy at the moment.

revam (changes requested):
My two cents since @ElementalCrisis forced my hand.

The rest of the code seems fine.

cxooda:
@revam done

Shoko.Server/Filters/FilterExtensions.cs
```diff
@@ -286,11 +287,28 @@ public static FilterableUserInfo ToFilterableUserInfo(this SVR_AnimeGroup group,
             .Select(a => RepoFactory.VideoLocalUser.GetByUserIDAndVideoLocalID(userID, a.VideoLocalID)?.WatchedDate).Where(a => a != null).OrderBy(a => a)
             .ToList();
 
+        // we only want to filter by watched states from files that we actually have and exclude trailers/credits, etc
+        int GetEpCount(bool getWatched)
+        {
+            var count = 0;
+            foreach (var ep in series.SelectMany(s => s.AnimeEpisodes))
+            {
+                if (ep.IsHidden || (ep.EpisodeTypeEnum != EpisodeType.Episode && ep.EpisodeTypeEnum != EpisodeType.Special)) continue;
```
revam, line 296: `SVR_AnimeSeries.AnimeEpisodes` already filters out hidden episodes, so we can remove the first check. For reference, to also get all episodes including the hidden ones episodes we would use `SVR_AnimeSeries.AllAnimeEpisodes`, but that's not needed in this case. And the enum checks can be simplified to a single `!(a is x or y)` statement. Also, the reason I am suggesting a `!(a is x or y)` over a `a is not x and y` or `a is not x or y` is because I have bad blood with them from Shokofin. 😔

```suggestion
                if (!(ep.EpisodeTypeEnum is EpisodeType.Episode or EpisodeType.Special)) continue;
```

revam (changes requested):
I think this will be the last review (hopefully).

cxooda:
@revam
Didn't want to open the pattern can of worms here, but since you suggested it: This is probably the most readable way to do it like that. Alternatively, `a is not x or not y` would also work.

revam (approved):

da3dsoul:
Thanks @revam

PR #1150, opened 2024-07-19
Update MediaInfo to 24.06

maxpiva:
Straightfoward.


Cazzar:
We really should make it so our build pipeline actually pulls these instead of versioning the binaries
=

PR #1153, opened 2024-07-21
Fix nullable properties in plugin abstractions

hidden4003:


Shoko.Plugin.Abstractions/Attributes/RenamerAttribute.cs
```diff
@@ -8,6 +8,7 @@ public class RenamerAttribute : Attribute
         public RenamerAttribute(string renamerId)
         {
             RenamerId = renamerId;
+            _desc = renamerId;
```
revam, line 11: Looking at how it's used, then it should probably be set to `null` and not the renamer id if not explicitly set by a renamer, or we can change up the `Description` to not use a private field and instead just be set to the renamer id initially like you did here.

Shoko.Plugin.Abstractions/Attributes/RenamerAttribute.cs
```diff
@@ -19,4 +20,4 @@ public string Description
             set => _desc = value;
         }
     }
-}
\ No newline at end of file
+}
```
revam, line 23: I don't like the removal of the trailing line endings, but I also won't force you to redo the PR to remove the changes. Same with the BOM(?) changes at the start of the files.

revam (changes requested):
the `_desc` change is the only one that's blocking for me. I'll assume the compiler complained because we were exiting the constructor without initializing all the properties, so you set it to the `renamerId` because that was what `Description` falls back to when not set.

I'll also assume we cannot just use `new()` because we target `netstandard2.0` so it's not available, but otherwise I would had commented to suggest updating the two lists to to use the newer syntax.

Shoko.Plugin.Abstractions/Attributes/RenamerAttribute.cs
hidden4003: I will look at it Visual Studio 2022 just does it by itself but it also says `User preferences overridden by .editorconfig` so I assumed those formatting changes came from there.

Shoko.Plugin.Abstractions/Attributes/RenamerAttribute.cs
hidden4003: I just set it to renamerId because of how the getter for it was coded, can set to empty string or make nullable.

revam: Please mark it as nullable then.

Shoko.Plugin.Abstractions/Attributes/RenamerAttribute.cs
```diff
@@ -8,6 +8,7 @@ public class RenamerAttribute : Attribute
         public RenamerAttribute(string renamerId)
         {
             RenamerId = renamerId;
+            _desc = renamerId;
         }
```
Cazzar, line 12:
```suggestion
        public RenamerAttribute(string renamerId, string description = "")
        {
            RenamerId = renamerId;
            _desc = description;
        }
```

Shoko.Plugin.Abstractions/Attributes/RenamerAttribute.cs
Cazzar: I'd suggest doing this instead: https://github.com/ShokoAnime/ShokoServer/pull/1153/files#r1685748964

Shoko.Plugin.Abstractions/Attributes/RenamerAttribute.cs
hidden4003: Adding extra line at the end is set in `.editorconfig` https://github.com/ShokoAnime/ShokoServer/blob/2451fb2591f696c2793327a162e9af08b41c756c/.editorconfig#L93 and BOM shouldn't be there according to that as well, maybe that one file had a BOM?

hidden4003:
I removed the extra private property `_desc` since `?? renamerId` will never happen now with constructor changes.

PR #1156, opened 2024-07-27
Adjust periodic checks frequency

hidden4003:


Shoko.Server/Scheduling/QuartzStartup.cs
```diff
@@ -30,7 +30,7 @@ public static async Task ScheduleRecurringJobs(bool replace)
         // Also give it a high priority, since it affects Acquisition Filters
         // StartJobNow gives a priority of 10. We'll give it 20 to be even higher priority
         await ScheduleRecurringJob<CheckNetworkAvailabilityJob>(
-            triggerConfig: t => t.WithPriority(20).WithSimpleSchedule(tr => tr.WithIntervalInMinutes(5).RepeatForever()).StartNow(), replace: true, keepSchedule: false);
+            triggerConfig: t => t.WithPriority(20).WithSimpleSchedule(tr => tr.WithIntervalInMinutes(30).RepeatForever()).StartNow(), replace: true, keepSchedule: false);
```
revam, line 33: if we're going to tweak this interval, then we should also recheck if we have internet access on every timed out http request in the jobs and when we receive a (potentially fake) UDP ban exception. otherwise we can have _a lot_ of jobs running in those 30 minutes that will fail and probably not be rescheduled immediately. We should had probably done the change to add checks upon receiving timeouts earlier, but better late than never.

hidden4003: is this not the check in case of request fail ? https://github.com/ShokoAnime/ShokoServer/blob/187d643aba4d67cb3b824afb50ea1e9c8a769064/Shoko.Server/Providers/AniDB/UDP/AniDBUDPConnectionHandler.cs#L268-L275

revam: Didn't know we had a check already for UDP commands. Thrn we only need to add it to the rest of the jobs, e.g. when fetching images, communicating with TvDB, Trakt or TMDB, etc.

hidden4003: I don't see an issue with it, generally your internet is not that flaky so for 99% of users checking every 5 minutes is just excessive and a waste of resources. Regarding other jobs I also see warnings like this one 

> warning SYSLIB0014: 'WebClient.WebClient()' is obsolete: 'WebRequest, HttpWebRequest, ServicePoint, and WebClient are obsolete. Use HttpClient instead.'

so some of that code probably need to be refactored, I also don't know how to add those checks in some common place without adding that to every invocation or exception handler which is too messy. 

So I think this should go as is and time increase between checks is not a big impact for majority of users, rather an improvement, and issue needs to be created for refactor of fetch related jobs/code.

revam: for now you've convinced me. we should still perhaps add the other checks latter at some point.

revam (approved):

revam:
If @da3dsoul isn't against the changes then he can merge it. 👍

PR #1162, opened 2024-08-13
Fix file relocation without rename

ale183:
With "Move on import" enabled and "Rename on import" disabled the file remains in the import folder for two reasons:
- in `WebAOMRenamer.cs` the variable `success` is always false, unless `GetNewFileName` is called (only if rename is enabled)
- in `VideoLocal_PlaceService.cs` there is a check on `result.FileName`, which is always null if the rename is disabled, and the file is never moved

This commit should fix that behavior.


revam:
I'll need to look at the new renamer internals for the latter change, but the success change seems to be fine. Will check when I get home.

revam (approved):

PR #1164, opened 2024-08-16
Make `TMDB_Show` and `TMDB_Episode` repos cached

harshithmohan:


Shoko.Server/Repositories/Cached/TMDB_EpisodeRepository.cs
```diff
@@ -0,0 +1,63 @@
...
+using NutzCode.InMemoryIndex;
+using Shoko.Server.Databases;
+using Shoko.Server.Models.TMDB;
+
+#nullable enable
+namespace Shoko.Server.Repositories.Cached;
+
+public class TMDB_EpisodeRepository : BaseCachedRepository<TMDB_Episode, int>
+{
+    private PocoIndex<int, TMDB_Episode, int> TmdbShowIds;
```
revam, line 12: isn't your IDE complaining that the property is null after the constructor?

revam (approved):
other tham that one comment then it looks fine.

PR #1167, opened 2024-08-21
Fix: Key collision in images in SVR_AnimeSeries

Momongaka:


revam:
@da3dsoul @Momongaka this is a band aid for a deeper issue.

da3dsoul:
Did you want a lookup instead?

revam:
> Is it? The error was thrown because there were multiple posters

there should only be a single _preferred_ poster per series. having multiple is an issue.

PR #1172, opened 2024-08-26
Add setting for UDP ping frequency

harshithmohan:


Shoko.Server/Providers/AniDB/UDP/Generic/UDPRequest.cs
```diff
@@ -136,8 +134,16 @@ protected virtual UDPResponse<string> ParseResponse(string response, bool return
...
                     break;
                 }
+            // 506 INVALID SESSION
+            // 598 UNKNOWN COMMAND
+            case UDPReturnCode.INVALID_SESSION:
             case UDPReturnCode.UNKNOWN_COMMAND:
-                throw new UnexpectedUDPResponseException(response: response, code: status, request: Command);
+                if (status == UDPReturnCode.UNKNOWN_COMMAND)
+                {
+                    Logger.LogWarning("AniDB returned UNKNOWN COMMAND. Try to increase UDPPingFrequency in the settings!");
```
hidden4003, line 143: Should be along the lines: `AniDB returned UNKNOWN COMMAND which likely means your session has expired. If you are behind NAT. please check your router's settings for how long it keeps track of active connections and adjust UDPPingFrequency in the settings accordingly.`

Cazzar: the "If you are behind NAT" can probably be removed though @hidden4003 because 99.99% of connections will be behind some level of NAT.

PR #1176, opened 2024-09-13
Fix cross-platform plex sync, get episode data using both parent folder and filename

harshithmohan:
Tested with plex on linux and shoko on windows, but should work the other way also.

- Plex on linux, shoko on windows
  - First replace -> no change, path doesn't have back slashes
  - Second replace -> forward slashes are replaced with back slashes
- Plex on windows, shoko on linux
  - First replace -> back slashes are replaced with forward slashes
  - Second replace -> no change, already has forward slashes
- Both plex and shoko on linux
  - First replace -> no change. Except for extreme cases where the filename somehow has `\` which it shouldn't
  - Second replace -> no change
- Both plex and shoko on windows
  - First and second replace -> no change

### DIFF
--- a/Shoko.Server/Plex/TVShow/SVR_Episode.cs
+++ b/Shoko.Server/Plex/TVShow/SVR_Episode.cs
@@ -1,4 +1,6 @@
-﻿using System.IO;
+﻿using System;
+using System.IO;
+using System.Linq;
 using Shoko.Models.Plex.TVShow;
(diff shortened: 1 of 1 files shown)


harshithmohan:
Closing this after a smol discussion in dev-chat 😔

Cazzar:
Reverting this, because there's a much more reliable method then that to work out proper paths.

Issue #958, opened 2015-06-12
Feature Request: New Filter Condition - Filter by Import Folder/Folder


Darken5:
Filter option to show/not show series by which "Import Folder" or just what "Folder" they are in.

revam:
\*_7 years later_\*

Still no time to take a look at it? Or is it too involved? @da3dsoul 

And on a more serious not; should I close this issue or do you want to implement it?

da3dsoul:
Adding group filters is a pita, so I've not done it in 7 years. Leave it open, because it's not done, and it should be eventually.

PR #1184, opened 2024-10-12
Add import folder filters

revam:
Add two new filters to filter by import folder IDs or names. Both are implemented as string sets because I didn't want to add a new filter category for number sets for the IDs.

Fixes/Implements #958.

@da3dsoul if you're fine with it as-is, then please merge. Otherwise I'll be awaiting your review.


da3dsoul:
I don't see any issues. If it works, then all good

PR #1198, opened 2024-11-12
Switch Params for Trakt Title Search

krbrs:
Fix Error 400 for TraktSearch by switching the params for the search URI to the correct order


harshithmohan:
For reference if anybody wants to know.
It's used here

https://github.com/ShokoAnime/ShokoServer/blob/master/Shoko.Server%2FProviders%2FTraktTV%2FTraktTVHelper.cs#L1044

PR #1200, opened 2024-11-13
Fix: Avoid the logout timer trying to acquire the lock twice

fearnlj01:


fearnlj01:
Woops, didn't mean to submit that without comment... But some experimentation suggested that `LogoutTimerElapsed` would acquire a lock, preventing the actual logout command from being able to lock.

(I think)

da3dsoul:
Have you tested that it at least logs out as expected?

fearnlj01:
I've done so when I had the logout timer set to two minutes, but can give it a more natural run on what I'm running now.

Will confirm again in ~5 minutes

fearnlj01:
Confirming that it logs out as expected in the "normal" test run, and allows for an immediate re-login afterwards without any errors 👍

PR #1201, opened 2024-11-15
Move CheckTraktToken to a Job

krbrs:
Move CheckTraktToken to a Job that runs every 12h instead of spamming the log every few minutes.

The refresh of a fairly old token worked on my dev machine, but I'd be happy if someone else could check

### DIFF
--- a/Shoko.Server/Providers/TraktTV/TraktTVHelper.cs
+++ b/Shoko.Server/Providers/TraktTV/TraktTVHelper.cs
@@ -235,7 +235,7 @@ public class TraktTVHelper
 
     #region Authorization
 
-    public void RefreshAuthToken()
+    public bool RefreshAuthToken()
     {
         var settings = _settingsProvider.GetSettings();
         try
@@ -248,7 +248,7 @@ public class TraktTVHelper
                 settings.TraktTv.RefreshToken = string.Empty;
                 settings.TraktTv.TokenExpirationDate = string.Empty;
 
-                return;
+                return false;
             }
 
             var token = new TraktV2RefreshToken { refresh_token = settings.TraktTv.RefreshToken };
@@ -272,7 +272,7 @@ public class TraktTVHelper
 
                 settings.TraktTv.TokenExpirationDate = expireDate.ToString();
 
-                return;
+                return true;
             }
 
             settings.TraktTv.AuthToken = string.Empty;
@@ -286,11 +286,13 @@ public class TraktTVHelper
             settings.TraktTv.TokenExpirationDate = string.Empty;
 
             _logger.LogError(ex, "Error in TraktTVHelper.RefreshAuthToken");
+            return false;
         }
         finally
         {
             Utils.SettingsProvider.SaveSettings();
         }
+        return false;
     }
 
     #endregion
--- a/Shoko.Server/Scheduling/Jobs/Trakt/CheckTraktTokenJob.cs
+++ b/Shoko.Server/Scheduling/Jobs/Trakt/CheckTraktTokenJob.cs
@@ -23,8 +23,6 @@ public class CheckTraktTokenJob : BaseJob
     public override string TypeName => "Check and Refresh Trakt Token";
     public override string Title => "Checking and Refreshing Trakt Token";
 
-    public override Dictionary<string, object> Details => new() { { "Action", "Check and Refresh Trakt Token" } };
-
     public override Task Process()
     {
         try
--- a/Shoko.Server/Scheduling/Jobs/Trakt/CheckTraktTokenJob.cs
+++ b/Shoko.Server/Scheduling/Jobs/Trakt/CheckTraktTokenJob.cs
@@ -1,0 +1,70 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Shoko.Server.Scheduling.Acquisition.Attributes;
+using Shoko.Server.Scheduling.Attributes;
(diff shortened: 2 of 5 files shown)


Shoko.Server/Scheduling/Jobs/Trakt/CheckTraktTokenJob.cs
```diff
@@ -0,0 +1,68 @@
...
+                return Task.CompletedTask;
+            }
+
+            // Convert the Unix timestamp to DateTime
+            var expirationDate = DateTimeOffset.FromUnixTimeSeconds(long.Parse(settings.TraktTv.TokenExpirationDate)).DateTime;
+
+            // Check if the token needs refreshing
+            if (ForceRefresh || DateTime.Now.Add(TimeSpan.FromDays(45)) >= expirationDate)
+            {
+                _traktHelper.RefreshAuthToken();
```
revam, line 44: Change the refresh method to return a bool to indicate if the refresh was successful, then log if it was successful or not below.

revam (changes requested):
A side effect of running it in the queue is that if the queue is paused for too long at the end of the expiring period then it will miss the timeframe to refresh the token, while running it on a timer outside the queue is not affected by this.

@da3dsoul can correct me if I'm wrong about this though.

Shoko.Server/Scheduling/Jobs/Trakt/CheckTraktTokenJob.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            _logger.LogInformation("Processing {Job}", nameof(CheckTraktTokenJob));
+            var settings = _settingsProvider.GetSettings();
+            if (!settings.TraktTv.Enabled || string.IsNullOrEmpty(settings.TraktTv.TokenExpirationDate))
+            {
+                _logger.LogInformation("Trakt is not enabled or no token expiration date is set");
+                return Task.CompletedTask;
+            }
+
+            // Convert the Unix timestamp to DateTime
```
revam, line 38: We should also probably add a check if the date is in the past, and unset the token if it is instead of trying to refresh.

krbrs:
> A side effect of running it in the queue is that if the queue is paused for too long at the end of the expiring period then it will miss the timeframe to refresh the token, while running it on a timer outside the queue is not affected by this.
> 
> 
> 
> @da3dsoul can correct me if I'm wrong about this though.

My token expired 3 months ago and was refreshed properly, so you don't have to unset it because you would have to link it again while the app is still approved on Trakt. I'll look into the other things, currently it would work and is just migrated code

Shoko.Server/Scheduling/Jobs/Trakt/CheckTraktTokenJob.cs
krbrs: If everybody is okay with my reply below, I'll mark this as resolved

Shoko.Server/Scheduling/Jobs/Trakt/CheckTraktTokenJob.cs
krbrs: done.

Shoko.Server/Providers/TraktTV/TraktTVHelper.cs
```diff
@@ -286,11 +286,13 @@ public void RefreshAuthToken()
             settings.TraktTv.TokenExpirationDate = string.Empty;
 
             _logger.LogError(ex, "Error in TraktTVHelper.RefreshAuthToken");
+            return false;
```
revam, line 289: this one is unnecessary since both the last path in the above try block and this block both lead out of the try…catch statement, and would hit the return statement below.

Shoko.Server/Scheduling/Jobs/Trakt/CheckTraktTokenJob.cs
```diff
@@ -41,9 +41,11 @@ public override Task Process()
...
             if (ForceRefresh || DateTime.Now.Add(TimeSpan.FromDays(45)) >= expirationDate)
             {
-                _traktHelper.RefreshAuthToken();
-                var newExpirationDate = DateTimeOffset.FromUnixTimeSeconds(long.Parse(settings.TraktTv.TokenExpirationDate)).DateTime;
-                _logger.LogInformation("Trakt token refreshed successfully. New expiry date: {Date}", newExpirationDate);
+                if (_traktHelper.RefreshAuthToken())
+                {
+                    var newExpirationDate = DateTimeOffset.FromUnixTimeSeconds(long.Parse(settings.TraktTv.TokenExpirationDate)).DateTime;
+                    _logger.LogInformation("Trakt token refreshed successfully. New expiry date: {Date}", newExpirationDate);
+                }
```
revam, line 48: We might want to log when it fails to refresh too.

revam (changes requested):
Last batch of requested changes. :pensive:

Shoko.Server/Providers/TraktTV/TraktTVHelper.cs
krbrs: I had to add it because the IDE would give me an error that the return state is missing :)

Shoko.Server/Scheduling/Jobs/Trakt/CheckTraktTokenJob.cs
krbrs: I removed that one because we have the catch statement that catches the error ? But I can add something again if you want.

revam (approved):

Issue #1202, opened 2024-11-18
NRE from TMDB Cast/Crew when a 'TMDB Person' does not yet exist


fearnlj01:
﻿## VERSION INFORMATION

Server Version: c24465c06fc6a5a1c15f057da78ed716263d2f08 ([current diff from `master`](https://github.com/ShokoAnime/ShokoServer/compare/c24465c06fc6a5a1c15f057da78ed716263d2f08...master))

## DESCRIPTION

[`Shoko.Server.Models.TMDB.TMDB_Crew`](https://github.com/ShokoAnime/ShokoServer/blob/c24465c06fc6a5a1c15f057da78ed716263d2f08/Shoko.Server/Models/TMDB/TMDB_Cast.cs) & [`Shoko.Server.Models.TMDB.TMDB_Cast`](https://github.com/ShokoAnime/ShokoServer/blob/c24465c06fc6a5a1c15f057da78ed716263d2f08/Shoko.Server/Models/TMDB/TMDB_Crew.cs) will both throw an exception if `GetTmdbPerson` is called but cannot find a corresponding person.

In turn, [`Shoko.Server.API.v3.Models.Common.Role`](https://github.com/ShokoAnime/ShokoServer/blob/c24465c06fc6a5a1c15f057da78ed716263d2f08/Shoko.Server/API/v3/Models/Common/Role.cs) is dependent upon these methods for the constructors taking TMDB data as a parameter, however does not account for the possibility of an exception being thrown.

In practice, this becomes an issue when using an endpoint such as `/api/v3/Tmdb/Episode/{episodeID}` with the option to retrieve Cast/Crew (as Shokofin does). If for whatever reason, there is something saved in `TMDB_Episode_Cast`, but no corresponding entries in the ~~other `TMDB` tables~~ the `TMDB_Person` table... a NRE occurs (which is currently a 'critical' issue in Shokofin's implementation, preventing the creation of a new library using the VFS).  
  
  
I've got no idea how my database ended up in the state it has done... But updating TMDB show info via the currently available Web UI actions does not end up resolving the issue.

I can make a copy of my SQLite DB at time of writing this issue available upon request.

## STEPS TO REPRODUCE

Using a copy of my DB... Making the following request:
`GET http://localhost:8111/api/v3/TMDB/Episode/5101834?include=Titles,Overviews,Images,CrossReferences,Cast,Crew,FileCrossReferences`

Cazzar:
[...]

PR #1203, opened 2024-11-19
fix: add DB migration & endpoint to download missing TMDB People

fearnlj01:
Optionally allows for the complete removal of the person/cast/crew entry if unable to update it.

This allows for a user-driven resolution to the problems that issue #1202 demonstrate.

### DIFF
--- a/Shoko.Server/API/v3/Controllers/TmdbController.cs
+++ b/Shoko.Server/API/v3/Controllers/TmdbController.cs
@@ -1478,6 +1478,48 @@ public partial class TmdbController : BaseController
         return NoContent();
     }
 
+    /// <summary>
+    /// Download any missing TMDB Persons.
+    /// </summary>
+    /// <param name="removeErrors">Removes all references to the bad TMDB Person if unable to successfully download the person</param>
+    /// <returns> <see cref="OkResult"/> with a summary of found/updated/skipped/deleted actions. </returns>
+    [HttpGet("Person/DownloadMissing")]
+    public async Task<ActionResult> RepairMissingTmdbPersons([FromQuery] bool removeErrors = false)
+    {
+        var personIds = new HashSet<int>();
+        var (errorCount, removedCount) = (0, 0);
+        
+        var people = RepoFactory.TMDB_Person.GetAll().Select(person => person.TmdbPersonID).ToList();
+
+        RepoFactory.TMDB_Episode_Cast.GetAll().AsParallel().Where(p =>
+            !people.Contains(p.TmdbPersonID)).ForEach(p => personIds.Add(p.TmdbPersonID));
+        RepoFactory.TMDB_Episode_Crew.GetAll().AsParallel().Where(p => 
+            !people.Contains(p.TmdbPersonID)).ForEach(p => personIds.Add(p.TmdbPersonID));
+        
+        RepoFactory.TMDB_Movie_Cast.GetAll().AsParallel().Where(p => 
+            !people.Contains(p.TmdbPersonID)).ForEach(p => personIds.Add(p.TmdbPersonID));
+        RepoFactory.TMDB_Movie_Crew.GetAll().AsParallel().Where(p => 
+            !people.Contains(p.TmdbPersonID)).ForEach(p => personIds.Add(p.TmdbPersonID));
+        
+        foreach (var id in personIds)
+            try
+            {
+                await _tmdbMetadataService.UpdatePerson(id, forceRefresh: true);
+            }
+            catch (NullReferenceException)
+            {
+                _logger.LogInformation("Unable to find TMDB Person record {@Id} on TMDB", id);
+                errorCount++;
+                if (removeErrors && await _tmdbMetadataService.PurgePerson(id, forceRemoval: true))
+                    removedCount++;
+            }
+        
+        var updateCount = personIds.Count - errorCount;
+        var skippedCount = errorCount - removedCount;
+
+        return Ok($"(Found/Updated/Skipped/Deleted) ({personIds.Count}/{updateCount}/{skippedCount}/{removedCount}) missing TMDB person(s).");
+    }
+    
     #endregion
 
     #region Online (Search / Bulk / Single)
--- a/Shoko.Server/API/v3/Controllers/TmdbController.cs
+++ b/Shoko.Server/API/v3/Controllers/TmdbController.cs
@@ -1479,17 +1479,15 @@ public partial class TmdbController : BaseController
     }
 
     /// <summary>
-    /// Download any missing TMDB Persons.
+    /// Download any missing TMDB Person records.
     /// </summary>
-    /// <param name="removeErrors">Removes all references to the bad TMDB Person if unable to successfully download the person</param>
-    /// <returns> <see cref="OkResult"/> with a summary of found/updated/skipped/deleted actions. </returns>
     [HttpGet("Person/DownloadMissing")]
-    public async Task<ActionResult> RepairMissingTmdbPersons([FromQuery] bool removeErrors = false)
+    public ActionResult RepairMissingTmdbPersonRecords()
     {
-        var result = await _tmdbMetadataService.RepairMissingPersons();
-        return Ok($"(Found/Updated/Skipped/Deleted) ({result.Found}/{result.Updated}/{result.Skipped}/{result.Removed}) missing TMDB person(s).");
+        Task.Run(() => _tmdbMetadataService.RepairMissingPersonRecords()).ConfigureAwait(false);
+        return Ok();
     }
-    
+
     #endregion
 
     #region Online (Search / Bulk / Single)
--- a/Shoko.Server/API/v3/Controllers/TmdbController.cs
+++ b/Shoko.Server/API/v3/Controllers/TmdbController.cs
@@ -1479,12 +1479,12 @@ public partial class TmdbController : BaseController
     }
 
     /// <summary>
-    /// Download any missing TMDB Person records.
+    /// Download any missing TMDB People.
     /// </summary>
     [HttpGet("Person/DownloadMissing")]
-    public ActionResult RepairMissingTmdbPersonRecords()
+    public ActionResult RepairMissingTmdbPeople()
     {
-        Task.Run(() => _tmdbMetadataService.RepairMissingPersonRecords()).ConfigureAwait(false);
+        Task.Run(() => _tmdbMetadataService.RepairMissingPeople()).ConfigureAwait(false);
         return Ok();
     }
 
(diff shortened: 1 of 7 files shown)


fearnlj01:
This is only a workaround to the original issue... but I needed this to fix my own database so figured that a PR wouldn't hurt, even if it gets closed.

In my own testing, I encountered an NRE for a person ID which appears to no longer exist on TMDB, hence this being the only thing that's caught.

revam (commented):
This would do better as a db migration once the core issue is fixed.

fearnlj01:
Confirming that this works fine™ in my SQLite setup.

I removed ~125 TMDB_Person records prior to launching the server, and after the migration occured, hitting the new endpoint resulted in no missing records being found.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -2031,6 +2031,50 @@ private void PurgeCompany(int companyId, bool removeImageFiles = true)
 
     #region People
 
+    public async Task<(int Found, int Updated, int Skipped, int Removed)> RepairMissingPersons(bool removeErrors = false)
```
revam, line 2034: The plural of 'Person' is 'People.' Engrish is weird. I know.

Shoko.Server/API/v3/Controllers/TmdbController.cs
```diff
@@ -1478,6 +1478,18 @@ public async Task<ActionResult> DownloadImagesForTmdbShowByShowID(
...
 
+    /// <summary>
+    /// Download any missing TMDB Persons.
+    /// </summary>
+    /// <param name="removeErrors">Removes all references to the bad TMDB Person if unable to successfully download the person</param>
+    /// <returns> <see cref="OkResult"/> with a summary of found/updated/skipped/deleted actions. </returns>
+    [HttpGet("Person/DownloadMissing")]
+    public async Task<ActionResult> RepairMissingTmdbPersons([FromQuery] bool removeErrors = false)
+    {
+        var result = await _tmdbMetadataService.RepairMissingPersons();
```
revam, line 1489: Since this a long running task then run it inside `Task.Run` and immediately return, and do the logging inside the metadata service instead.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -2031,6 +2031,50 @@ private void PurgeCompany(int companyId, bool removeImageFiles = true)
...
+            .Where(p => !people.Contains(p.TmdbPersonID))
+            .ForAll(p => missingIds.Add(p.TmdbPersonID));
+        
+        _tmdbMovieCast.GetAll().AsParallel()
+            .Where(p => !people.Contains(p.TmdbPersonID))
+            .ForAll(p => missingIds.Add(p.TmdbPersonID));
+        _tmdbMovieCrew.GetAll().AsParallel()
+            .Where(p => !people.Contains(p.TmdbPersonID))
+            .ForAll(p => missingIds.Add(p.TmdbPersonID));
+
```
revam, line 2054: Add logging before running the loop here.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -2031,6 +2031,50 @@ private void PurgeCompany(int companyId, bool removeImageFiles = true)
...
+                if (removeErrors && await PurgePerson(personId, removeImageFiles: true, forceRemoval: true))
+                {
+                    removedCount++;
+                    continue;
+                }
+                skippedCount++;
+            }
+        }
+
```
revam, line 2074: Add logging for the final result/changes here.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -2031,6 +2031,50 @@ private void PurgeCompany(int companyId, bool removeImageFiles = true)
 
     #region People
 
+    public async Task<(int Found, int Updated, int Skipped, int Removed)> RepairMissingPersons(bool removeErrors = false)
+    {
+        var missingIds = new HashSet<int>();
+        var (updateCount, skippedCount, removedCount) = (0, 0, 0);
+        
+        var people = _tmdbPeople.GetAll().Select(person => person.TmdbPersonID).ToList();
```
revam, line 2039: Create a hash-set instead for faster `Contains` look-ups compared to a list

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -2031,6 +2031,50 @@ private void PurgeCompany(int companyId, bool removeImageFiles = true)
...
+        
+        _tmdbEpisodeCast.GetAll().AsParallel()
```
revam, line 2041: Drop the parallel in these, and do a `for each(…¹)` loop instead of `.ForAll`. we don't need to use _all_ available threads for this. you can convert the `.Where` into an if statement inside the `for each`. be it an `if(…¹) continue;` statement or an `if(…¹) …²` statement.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -2031,6 +2031,50 @@ private void PurgeCompany(int companyId, bool removeImageFiles = true)
...
+            try
+            {
+                await UpdatePerson(personId, forceRefresh: true);
+                updateCount++;
+            }
+            catch (NullReferenceException)
+            {
+                _logger.LogDebug("Unable to update TMDB Person ({@PersonId})", personId);
+
+                if (removeErrors && await PurgePerson(personId, removeImageFiles: true, forceRemoval: true))
```
revam, line 2066: It would be better to keep a reference to all the cast/crew roles the person has credits for, and update those, instead of just removing the person. Also don't like that we're just checking for an NPE. Would be better to handle it gracefully instead of throwing, so we should probably update the code below, on line 2051, to expect the remote person to be null-able, and then schedule an update for all needed entities to naturally remove the person from the local database once all references to them are gone.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -2072,11 +2116,11 @@ private async Task DownloadPersonImages(int personId, ProfileImages images, bool
         await _imageService.DownloadImagesByType(images.Profiles, ImageEntityType.Person, ForeignEntityType.Person, personId, settings.TMDB.MaxAutoStaffImages, [], forceDownload);
     }
 
-    public async Task<bool> PurgePerson(int personId, bool removeImageFiles = true)
+    public async Task<bool> PurgePerson(int personId, bool removeImageFiles = true, bool forceRemoval = false)
```
revam, line 2119: Really don't like the forced removal. as said earlier, we should be attempting to fix it more gracefully if possible, if the person we're missing has any cast/crew credits locally, instead of just attempting to purge all crew/cast relations. by updating the linked entities it will naturally remove the person once all references are gone.

revam (changes requested):
A few things before we merge things. If you're unsure about something then you can add a comment here on GH or ask on discord.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -2105,6 +2100,28 @@ private void PurgeCompany(int companyId, bool removeImageFiles = true)
...
+        foreach (var staff in _tmdbEpisodeCrew.GetByTmdbPersonID(tmdbPersonId))
+            showIds.Add(staff.TmdbShowID);
+
+        foreach (var staff in _tmdbMovieCast.GetByTmdbPersonID(tmdbPersonId))
+            movieIds.Add(staff.TmdbMovieID);
+        foreach (var staff in _tmdbMovieCrew.GetByTmdbPersonID(tmdbPersonId))
+            movieIds.Add(staff.TmdbMovieID);
+
+        foreach (var showId in showIds)
+            await UpdateShow(showId, downloadCrewAndCast: true);
```
revam, line 2120: If you schedule the updates to run in the queue, then you can just await the method and let the queue auto scale these in parallel (if possible).

fearnlj01:
Updated in line with feedback now.

One issue I can forsee... If a TMDB show has been updated in the last calendar hour when the migration runs, it may not update as expected in the event TMDB returns `null` for a person.

N.B. I've used `PersonRecords` in favour of `People`; This is with the intention of keeping it as close to the name for the underlying table as possible. (I will change accordingly if so desired)

revam:
> Updated in line with feedback now.
> 
> One issue I can forsee... If a TMDB show has been updated in the last calendar hour when the migration runs, it may not update as expected in the event TMDB returns `null` for a person.

Just run it with force refresh set to true.

> N.B. I've used `PersonRecords` in favour of `People`; This is with the intention of keeping it as close to the name for the underlying table as possible. (I will change accordingly if so desired)

Personally, i would had just used peole.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -2031,6 +2031,38 @@ private void PurgeCompany(int companyId, bool removeImageFiles = true)
 
     #region People
 
+    public async Task RepairMissingPeople()
+    {
+        var missingIds = new HashSet<int>();
+        var (updateCount, skippedCount) = (0, 0);
```
revam, line 2037: Didn't see it earlier, but please don't use tuple destruction to assign two new variables and instead assign them separately.

revam (approved):

Shoko.Server/Databases/MySQL.cs
```diff
@@ -849,6 +849,7 @@ public class MySQL : BaseDatabase<MySqlConnection>
         new(139, 10, "ALTER TABLE Trakt_Show ADD COLUMN TmdbShowID INT NULL;"),
         new(139, 11, DatabaseFixes.CleanupAfterRemovingTvDB),
         new(139, 12, DatabaseFixes.ClearQuartzQueue),
+        new (140, 1, DatabaseFixes.RepairMissingTMDBPersons)
```
revam, line 852:
```suggestion
        new(140, 1, DatabaseFixes.RepairMissingTMDBPersons)
```

Shoko.Server/Databases/SQLite.cs
```diff
@@ -774,6 +774,7 @@ public override void CreateDatabase()
         new(123, 10, "ALTER TABLE Trakt_Show ADD COLUMN TmdbShowID INTEGER NULL;"),
         new(123, 11, DatabaseFixes.CleanupAfterRemovingTvDB),
         new(123, 12, DatabaseFixes.ClearQuartzQueue),
+        new (124, 1, DatabaseFixes.RepairMissingTMDBPersons)
```
revam, line 777:
```suggestion
        new(124, 1, DatabaseFixes.RepairMissingTMDBPersons)
```

revam (commented):
Only one nit-pick left, but I'll just fix that through GH.

PR #1210, opened 2024-12-13
Update datetime to datetime2 on TMDB Tables for MSSQL

krbrs:
Fixes `The conversion of a datetime2 data type to a datetime data type resulted in an out-of-range value.` on MSSQL


krbrs:
> ## Suspect Issues
> This pull request was deployed and Sentry observed the following issues:
> 
> * ‼️ **NHibernate.Exceptions.GenericADOException: could not execute native bulk manipulation query:ALTER TABLE AniDB_Creator DROP CONSTRAINT [SQL: ...** `Databases/SQLServer.cs in void SQLServer.DropDe...` [View Issue](https://shoko-anime.sentry.io/issues/6145283823/?referrer=github-pr-bot)
> 
> Did you find this useful? React with a 👍 or 👎

This was actually me running the schema changes on an already altered DB where the constraint was deleted by hand.
After adding it by hand and restarting server the problem was gone :) Also the prod DB change went well.

PR #1228, opened 2025-02-24
Add Release Info Providers [WIP]

revam:
First PoC of release providers. Still largely untested so far. Booting up, the DB migration, and file linking using the existing APIv3 endpoints are tested, the rest remains untested as of right now.

Things in this PR (from the top of my head):

- Added a new `IVideoReleaseService`, `IReleaseInfo`, `IReleaseVideoCrossReference`, `IReleaseMediaInfo`, `VideoReleaseEventArgs` to the plugin abstraction.

- Added new `/api/v3/ReleaseInfo/*` endpoints, allowing RESTful clients to also interact with the newly added `IVideoReleaseService`.

- File linking has been soft-deprecated and is no longer possible in APIv1.

- File linking in APIv3 have been converted to use the new service, and as a result the artificial limit of not allowing the user to remove AniDB releases is gone. A release is simply a release now.

- `AniDB_File`, `AniDB_FileUpdate`, `AniDB_ReleaseGroup`, `CrossRef_Languages_AniDB_File`, & `CrossRef_Subtitles_AniDB_File` models/tables/repos are gone, and their functionality replaced by the new `StoredReleaseInfo` & `StoredReleaseInfo_MatchAttempt`. The AniDB file has also been removed from the abstraction.

- Currently I've assigned every existing link as a "manual link", because the user is now able to edit every link we store if they so desire, and this was the simplest way to show all the links in the current Web UI.

- There is currently an example plugin in the server code, but that should probably be either;
  - moved out to it's own repository,
  - moved into the server, or
  - made to be built & included as a standalone plugin included with the server _somehow_.


da3dsoul:
This is...a lot, so I'll need to look at it more later. One thing I see first off is the complication and kind of hacky handling of the scheduling and ProcessFile. I would split the jobs if possible, and make it so that it has a flow like so:
```
Discover
Hash
ProcessFile
    Check the state of the data and what providers can/need to update, prolly via an interface/abstraction
    Schedule the relevant jobs for each provider. AniDB will have one, which can be handled in scheduler via the exclusion types. Ashen might have one. Maybe an NFO one? It's extensible, after all.
Get Provider File Info
    The job mentioned before. It can do the job that Process File did and orchestrate other things. We can make helpers or a base "Get Provider File Info" in the abstractions for providers to extend.
...
```
The plugin abstractions might need to provide a hook to add Acquisition Filters, jobs, etc

da3dsoul:
@Cazzar can you comment on some of the design? We aren't nitpicking code quality yet.

though if we were, stop making constructors for models. It'll mess up Entity Framework, and I'm going to get rid of them anyway. Use object initializers.
i.e.
```cs
public StoredReleaseInfo(IVideo video, IReleaseInfo releaseInfo)
...
```
If models should be models. If processing needs to be done, it should be in a service/factory. I don't know if your "embedded" models will work. We will see. I'm not sure how Entity Framework will handle loading of relationships through them.

Shoko.Server/Models/Release/StoredReleaseInfo.cs
```diff
@@ -0,0 +1,285 @@
...
+
+    string? IHashes.SHA1 => this[HashAlgorithmName.SHA1];
+
+    string? IHashes.CRC => this[HashAlgorithmName.CRC32];
+
+    public string? this[HashAlgorithmName algorithm]
+    {
+        get
+        {
+            var (ed2k, md5, sha1, crc32) = Hashes?.Split('|', StringSplitOptions.TrimEntries) ?? [];
```
da3dsoul, line 187: I'm not sure this will work. We can't guarantee that a plugin provided hash won't contain a pipe. I would store these in a separate table with a string Type, string or blob Hash, and any IDs necessary to link

da3dsoul: Maybe a few columns for "extra data". Kind of like how some hashes require a salt or whatever to reproduce.

Shoko.Server/API/v3/Models/Shoko/Episode.cs
```diff
@@ -119,7 +119,7 @@ public class Episode : BaseModel
     [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
     public IEnumerable<FileCrossReference.EpisodeCrossReferenceIDs>? CrossReferences { get; set; }
 
-    public Episode(HttpContext context, SVR_AnimeEpisode episode, HashSet<DataSource>? includeDataFrom = null, bool includeFiles = false, bool includeMediaInfo = false, bool includeAbsolutePaths = false, bool withXRefs = false)
+    public Episode(HttpContext context, SVR_AnimeEpisode episode, HashSet<DataSource>? includeDataFrom = null, bool includeFiles = false, bool includeMediaInfo = false, bool includeAbsolutePaths = false, bool withXRefs = false, bool includeReleaseInfo = false)
```
Cazzar, line 122: I think we need to rethink this API model a bit better.
Something like

```cs
new Episode(_httpContext, _episode)
  .IncludeDataFrom(...)
  .WithFiles()
  .WithMediaInfo()
  .AbsolutePaths()
  .XRefs()
  .WithReleaseInfo()
```
might reduce some of the constructor bloat.

Shoko.Server/API/v3/Models/Shoko/File.cs
```diff
@@ -107,32 +110,30 @@ public partial class File
...
 
-    public File(SVR_VideoLocal_User userRecord, SVR_VideoLocal file, bool withXRefs = false, HashSet<DataSource> includeDataFrom = null, bool includeMediaInfo = false, bool includeAbsolutePaths = false)
+    public File(SVR_VideoLocal_User? userRecord, SVR_VideoLocal file, bool withXRefs = false, bool includeReleaseInfo = false, bool includeMediaInfo = false, bool includeAbsolutePaths = false)
     {
         var mediaInfo = file.MediaInfo as IMediaInfo;
         ID = file.VideoLocalID;
         Size = file.FileSize;
         IsVariation = file.IsVariation;
-        Hashes = new() { ED2K = file.Hash, MD5 = file.MD5, CRC32 = file.CRC32, SHA1 = file.SHA1 };
+        Hashes = new(file);
```
Cazzar, line 136: I would just leave this as an instance of the interface, not wrapping it with a POCO.

Shoko.Server/Databases/MySQL.cs
```diff
@@ -723,7 +723,7 @@ public class MySQL : BaseDatabase<MySqlConnection>
         new(109, 1, "UPDATE VideoLocal v INNER JOIN CrossRef_File_Episode CRFE on v.Hash = CRFE.Hash SET DateTimeImported = DateTimeCreated;"),
         new(110, 1, "CREATE TABLE `AniDB_FileUpdate` ( `AniDB_FileUpdateID` INT NOT NULL AUTO_INCREMENT, `FileSize` BIGINT NOT NULL, `Hash` varchar(50) NOT NULL, `HasResponse` BIT NOT NULL, `UpdatedAt` datetime NOT NULL, PRIMARY KEY (`AniDB_FileUpdateID`) );"),
         new(110, 2, "ALTER TABLE `AniDB_FileUpdate` ADD INDEX `IX_AniDB_FileUpdate` (`FileSize` ASC, `Hash` ASC) ;"),
-        new(110, 3, DatabaseFixes.MigrateAniDB_FileUpdates),
+        new(110, 3, DatabaseFixes.NoOperation),
```
Cazzar, line 726: What's the reason for killing an existing migration?

Shoko.Server/Models/Release/StoredReleaseInfo.cs
Cazzar: Additionally why don't we provide a FileHash table which is a combination of:
```
FileID   | int FK | PK
HashType | string | PK
Hash     | String |
```

Cazzar:
Overall design I like the idea, I haven't looked through it extensively as the large amount of changes does make things complex.

Shoko.Server/Databases/MySQL.cs
revam: It is no longer relevant to run since they're later removed, and the models the migration used is no longer available.

Shoko.Server/API/v3/Models/Shoko/File.cs
revam: The interface would need to be moved or duplicated to the API layer to have proper model names in swagger I think. Currently it's wrapping it in the hash dict model used in the APIv3, which I also reused in the release info APIv3 model.

revam:
> This is...a lot, so I'll need to look at it more later. One thing I see first off is the complication and kind of hacky handling of the scheduling and ProcessFile. I would split the jobs if possible, and make it so that it has a flow like so:
> 
> ```
> Discover
> Hash
> ProcessFile
>     Check the state of the data and what providers can/need to update, prolly via an interface/abstraction
>     Schedule the relevant jobs for each provider. AniDB will have one, which can be handled in scheduler via the exclusion types. Ashen might have one. Maybe an NFO one? It's extensible, after all.
> Get Provider File Info
>     The job mentioned before. It can do the job that Process File did and orchestrate other things. We can make helpers or a base "Get Provider File Info" in the abstractions for providers to extend.
> ...
> ```
> 
> The plugin abstractions might need to provide a hook to add Acquisition Filters, jobs, etc

The current service can be ran inside or outside the queue/job system, and in the current PoC then the release providers are processed in a _user-configurable_ order until a release is found. Only a single release can be assigned to the same video at any given time, so scheduling "relevant jobs for each provider" won't ever happen in parallel even if we changed how it works. In short, the logic to select a release happens strictly inside the service and the process-file job is now just asking the service to do it's thing on the backend. There are also other ways to interact with the service, be it from other plugins through the abstraction, or from RESTful clients through the new endpoints.

I do admit that the way I modified the AniDB banned acquisition filter is kind of hacky, but also correct, as it was modified to only block the process-file jobs if AniDB banned **ONLY IF** the AniDB release provider is enabled, as it needs to be able to use the AniDB UDP API. But flipping that would mean that as long as the AniDB provider isn't enabled then we don't need to block the process-file jobs at all, since it's not using the AniDB UDP API to find releases.

Shoko.Server/API/v3/Models/Shoko/Episode.cs
revam: It's not a bad idea to clean up the constructor bloat in the API layer, but can we work on this after this PR? I can look into it unless you or somebody else wants to after the PR, but I don't want to deal with converting it _just now_.

Shoko.Server/Models/Release/StoredReleaseInfo.cs
da3dsoul: Presumably, the point of abstracting the hashes is so that new hash types could be added, such as perceptual hashes. In those cases, the extra data would be timestamps. 
The extra data would be on the hash object. The hash object would replace the hashes everywhere a hash is relevant

da3dsoul:
That is a reasonable argument, but I would allow multiple so that you can cross reference them. AniDB is more likely correct than perceptual hashing, even though perceptual hashing is pretty accurate. We could even have a filename plugin with very low accuracy. Maybe add an enum for how trustworthy we expect a provider to be in those cases.

revam:
We kinda already have a user configurable "priority" to use. Can we add two modes, one mode to run in a sync. loop in the priority order until a release is found (AKA the current way), and one mode to run multiple providers in parallel, await all the responses, and then pick the highest priority release of the available candidates? I can add the new setting to the service and modify the description of the auto-finding method and endpoints to reflect the new ways it can be ran. The reason I would opt to keep both modes is to let the user choose how they want to do it. By default we will only have one included provider (…unless…), and my particular flow would require it to happen in sync. so it first checks the "nfo" (quotes intentional) file before asking any remote services (or a potential fallback local/offline provider), but some would maybe want to do it in a parallel fashion as you described with the phash and AniDB provider, so I'll say "let them pick their own poison."

Shoko.Server/Models/Release/StoredReleaseInfo.cs
revam: so if I'm understanding it right, then we add a new hash object (hash type, hash value, extra data) and convert all occurrences of the existing `IHashes` interface to instead expose a list of these new hash objects?

Rand-Random:
Why was this closed? Hope all the hard work wasn’t in vain.

PR #1235, opened 2025-03-29
Fixes for new Trakt collection limits

krbrs:
Add Setting for VipStatus, added some constants for later use, changed TraktTVHelper to not update collections if user doesn't have VIP status because of new ridiculously low collection limits introduced in January 25


krbrs:
Would have checked the VIP status from Trakt and set it automatically, if they provided a simple bool. But you would have to query the users limits and compare manually if limit is a certain number and then decide if it's VIP or not. 
So better to have the user set it manually if they are, so we don't have to check new limits all the time if they announce them on forums. WebUI needs the setting added, I couldn't work on that today.

PR #1254, opened 2025-07-15
change mysql and sqlserver tmdb keyword to text equivalent

lbux:
SQLite already has this column as their text equivalent, so it was left untouched. [Lord of Mysteries](https://www.themoviedb.org/tv/232230?language=en-US) has too many keywords that it would not make sense to keep expanding and expanding varchar when new release come out with even more keywords.

This is related to this specific error message:

```
11:45:39| AnimeSeriesRepository --- Saving Series Guimi Zhi Zhu
11:45:39| AnimeSeriesRepository --- Saving Series Guimi Zhi Zhu | Waiting for Database Lock
11:45:39| AnimeSeriesRepository --- Saving Series Guimi Zhi Zhu | Got Database Lock in 0.00001s
11:45:39| AnimeSeriesRepository --- Saving Series Guimi Zhi Zhu | Got Series from Database in 0.00144s
11:45:39| AnimeSeriesRepository --- Saving Series Guimi Zhi Zhu | Saving Series to Database
11:45:39| AnimeSeriesRepository --- Saving Series Guimi Zhi Zhu | Saved Series to Database in 0.00887s
11:45:39| AnimeSeriesRepository --- Saving Series Guimi Zhi Zhu | Finished Saving in 0.01051s
11:45:39| TmdbLinkingService --- Removing TMDB show link: AniDB anime (AnimeID=19357) → TMDB show (ID=232230)
11:45:39| TmdbLinkingService --- Removing 0 episodes cross-references for AniDB anime (AnimeID=19357) and TMDB show (ID=232230)
11:45:39| AnimeGroupService --- Starting Updating STATS for GROUP Lord of Mysteries from Top Level (recursively) - Watched Stats: False, Missing Episodes: False
11:45:39| AnimeGroupService --- Starting Updating STATS for GROUP Lord of Mysteries - Watched Stats: False, Missing Episodes: False
11:45:39| AnimeGroupService --- Finished Updating STATS for GROUP Lord of Mysteries in 0.8625ms
11:45:39| AnimeGroupService --- Finished Updating STATS for GROUP Lord of Mysteries from Top Level (recursively) in 0.9102ms
11:45:41| TmdbMetadataService --- Scheduled call: Searching shows for "诡秘之主" at year 2025
11:45:41| TmdbMetadataService --- Executing call: Searching shows for "诡秘之主" at year 2025 (Waited 0.0433ms)
[...]

PR #1255, opened 2025-07-16
Fix MSSQL migration problem introduced with #1254

krbrs:
Drop default constraint on TMDB_Show and TMDB_Movie Keywords column before modification of Keywords


revam (commented):
C# cide always run much later so even if you put it as 01 then it won't run first if you do it in C# and not raw SQL.

harshithmohan:
Let the record show that @revam was wrong

PR #1256, opened 2025-07-16
Increase SignalR timeout

hidden4003:
For cases mentioned by @harshithmohan "when the server gets too many API calls or it gets stuck processing a single big one, signalr disconnects"


revam (commented):
My only gripe is that `.AddNewtonsoftJsonProtocol` lost it indent. otherwise LGTM.

PR #1263, opened 2025-08-09
Rename TMDB_Show_Network in SQLite in 2 steps

krbrs:


krbrs:
SQLite handles this case insensitive, still the table name is inconsistent with others.

Can someone with SQLite test if the renaming back and forth works and then merge it?
Fixes failing DB-Migration for SQLite introduced in PR #1260 if that works

revam (approved):
Actually tested it this time.

PR #1273, opened 2025-09-06
feat: added SeriesVoted event for plugin developers

shalevc1098:
so i decided to add that event myself :)

### DIFF
--- a/Shoko.Plugin.Abstractions/IShokoEventHandler.cs
+++ b/Shoko.Plugin.Abstractions/IShokoEventHandler.cs
@@ -69,6 +69,11 @@ public interface IShokoEventHandler
     event EventHandler<MovieInfoUpdatedEventArgs> MovieUpdated;
 
     /// <summary>
+    /// Fired when a user submits a rating/vote for a series.
+    /// </summary>
+    event EventHandler<SeriesVotedEventArgs> SeriesVoted;
+
+    /// <summary>
     /// Fired when the core settings has been saved.
     /// </summary>
     event EventHandler<SettingsSavedEventArgs> SettingsSaved;
--- a/Shoko.Server/API/v3/Controllers/SeriesController.cs
+++ b/Shoko.Server/API/v3/Controllers/SeriesController.cs
@@ -2445,6 +2445,7 @@ public class SeriesController : BaseController
             _ => series.AniDB_Anime?.GetFinishedAiring() ?? false ? AniDBVoteType.Anime : AniDBVoteType.AnimeTemp,
         };
         await _seriesService.AddSeriesVote(series, voteType, vote.GetRating());
+        ShokoEventHandler.Instance.OnSeriesVoted(series, series.AniDB_Anime, vote.GetRating(), voteType, User);
 
         return NoContent();
     }
--- a/Shoko.Server/Server/ShokoEventHandler.cs
+++ b/Shoko.Server/Server/ShokoEventHandler.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.IO;
 using System.Linq;
+using Shoko.Models.Enums;
 using Shoko.Plugin.Abstractions;
 using Shoko.Plugin.Abstractions.DataModels;
 using Shoko.Plugin.Abstractions.Enums;
@@ -38,6 +39,8 @@ public class ShokoEventHandler : IShokoEventHandler
     public event EventHandler<EpisodeInfoUpdatedEventArgs>? EpisodeUpdated;
 
     public event EventHandler<MovieInfoUpdatedEventArgs>? MovieUpdated;
+
+    public event EventHandler<SeriesVotedEventArgs>? SeriesVoted;
 
     public event EventHandler<SettingsSavedEventArgs>? SettingsSaved;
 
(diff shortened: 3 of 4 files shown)


Shoko.Server/Server/ShokoEventHandler.cs
```diff
@@ -271,6 +274,15 @@ public void OnMovieUpdated(TMDB_Movie movie, UpdateReason reason)
         MovieUpdated?.Invoke(null, new(movie, reason));
     }
 
+    public void OnSeriesVoted(SVR_AnimeSeries series, SVR_AniDB_Anime anime, decimal voteValue, AniDBVoteType voteType, SVR_JMMUser user)
+    {
+        ArgumentNullException.ThrowIfNull(series, nameof(series));
+        ArgumentNullException.ThrowIfNull(anime, nameof(anime));
+        ArgumentNullException.ThrowIfNull(user, nameof(user));
+        var voteTypeString = voteType == AniDBVoteType.Anime ? "Permanent" : "Temporary";
```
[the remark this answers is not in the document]
shalevc1098: Sounds better, will make that an enum and update the pull

shalevc1098:
@revam i did the changes you and da3dsoul suggested. when you have time if you want you can take a look and merge :)

Shoko.Server/API/v3/Controllers/SeriesController.cs
```diff
@@ -2445,6 +2445,8 @@ public async Task<ActionResult> PostSeriesUserVote([FromRoute, Range(1, int.MaxV
             _ => series.AniDB_Anime?.GetFinishedAiring() ?? false ? AniDBVoteType.Anime : AniDBVoteType.AnimeTemp,
         };
         await _seriesService.AddSeriesVote(series, voteType, vote.GetRating());
+        var pluginVoteType = voteType == AniDBVoteType.Anime ? VoteType.Permanent : VoteType.Temporary;
```
revam, line 2448: This should either be moved into the series service, or the method on the series service should be moved to the user data service and  exposed so other plugins can use it (in addition to the core) to vote on series.

Shoko.Plugin.Abstractions/IShokoEventHandler.cs
```diff
@@ -68,6 +68,11 @@ public interface IShokoEventHandler
     /// </summary>
     event EventHandler<MovieInfoUpdatedEventArgs> MovieUpdated;
 
+    /// <summary>
+    /// Fired when a user submits a rating/vote for a series.
+    /// </summary>
+    event EventHandler<SeriesVotedEventArgs> SeriesVoted;
```
revam, line 74: You didn't remove it from the shoko event handler yet.

Shoko.Plugin.Abstractions/Services/IUserDataService.cs
```diff
@@ -72,6 +72,25 @@ public interface IUserDataService
...
+
+    /// <summary>
+    /// Triggers the SeriesVoted event.
+    /// </summary>
+    /// <param name="series">The series being voted on.</param>
+    /// <param name="anime">The AniDB anime information.</param>
+    /// <param name="voteValue">The vote value.</param>
+    /// <param name="voteType">The type of vote.</param>
+    /// <param name="user">The user who voted.</param>
+    void OnSeriesVoted(IShokoSeries series, ISeries anime, decimal voteValue, VoteType voteType, IShokoUser user);
```
revam, line 90: The method to invoke the event (without also submitting and/or saving the vote) shouldn't be exposed to the plugins.

revam (changes requested):
Had some time to actually take a look. +1 for the enum change, but I still have some things that needs to be resolved before I approve this. @da3dsoul feel free to provide your thoughts on the matter if they differ from (or are the same as) mine.

Shoko.Server/Services/AbstractUserDataService.cs
```diff
@@ -235,6 +240,80 @@ public async Task<bool> SetEpisodeWatchedStatus(IShokoUser user, IShokoEpisode e
...
+
+        ArgumentNullException.ThrowIfNull(user);
+
+        if (series is not SVR_AnimeSeries svrSeries)
+            throw new ArgumentException("Series must be a SVR_AnimeSeries", nameof(series));
+
+        if (svrSeries.AniDB_Anime is not { } anidbAnime)
+            throw new ArgumentException("AniDB anime is not available for the series! Aborting!");
+
+        if (voteValue < -1 || voteValue > 10)
```
revam, line 260: since `voteValue` is a decimal then we should probably check if it is exactly `-1` or between `0` and `10`, so it cannot be -0.00001 and make havoc on the logic below.

Shoko.Server/Services/AbstractUserDataService.cs
```diff
@@ -235,6 +240,80 @@ public async Task<bool> SetEpisodeWatchedStatus(IShokoUser user, IShokoEpisode e
...
+            throw new ArgumentOutOfRangeException(nameof(voteValue), "Vote value must be between -1 and 10");
+
+        var anidbVoteType = voteType == VoteType.Permanent ? AniDBVoteType.Anime : AniDBVoteType.AnimeTemp;
+
+        // Handle deletion case
+        if (voteValue == -1)
```
revam, line 266: or we can convert this check to handle everything below zero (`voteValue < 0`) if we don't want to do the changes to the range check above.

revam (changes requested):
other than the voteValue check(s) then it seems good.

PR #1276, opened 2025-09-11
Replace external ftp artifact upload with sftp

SilentBot1:
This PR replaces the existing FTP-based uploading of external artifacts to [files.shokoanime.com](https://files.shokoanime.com/) with SFTP. 

Due to this and the use of private keys for authentication, this PR requires the following new secrets to be populated:

- `FILES_SSH_KEY`: The private key for an associated public key which belongs to `FILES_USERNAME` on `FILES_SERVER`.
- `FILES_KNOWN_HOSTS`: The output of `ssh-keyscan -t rsa $FILES_SERVER`.
- `FILES_USERNAME`: The user who has the public key associated with `FILES_SSH_KEY` installed.
- `FILES_SERVER`: The server that the user and public key exist on.

Note: sftp.exe is pre-installed on the Windows-based image used to handle the uploads, so no pre-install step is required.

### DIFF
--- a/.github/workflows/build-daily.yml
+++ b/.github/workflows/build-daily.yml
@@ -431,7 +431,7 @@ jobs:
           path: ShokoServer
 
       - name: Install SSH key
-        uses: shimataro/ssh-key-action@v2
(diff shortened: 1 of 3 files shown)


revam (approved):
In silence we thrust. 📿🙏

PR #1298, opened 2026-01-28
Add action to get watch states from Trakt

harshithmohan:


harshithmohan:
@revam Can you check if I have done anything bad because I touched a lot of places? :pensive: 
(I'm using the same frequency for both the jobs)

PR #1312, opened 2026-04-06
Log renamer script whenever possible

Mik1ll:
Current renamer migration strategy makes it very likely user will lose configs (user upgraded Shoko and started before plugin updated). Try to log the lost config and script.

Renamer-defined GUID makes it hard to populate the new configs, maybe we can use the config type as a temp ID or something.


revam (approved):

revam:
> Renamer-defined GUID makes it hard to populate the new configs, maybe we can use the config type as a temp ID or something.

If you're willing, then you can submit a follow-up PR to add a form of temp. id we can later re-map or similar.

PR #1328, opened 2026-05-22
fix: remove obsolete recurring Quartz jobs after startup reconciliation

krbrs:
## Summary

This PR removes obsolete persisted recurring Quartz jobs during startup reconciliation.

Trakt support was removed, but old installations could still keep persisted recurring Trakt jobs in the Quartz database. Since Quartz state survives upgrades, simply removing the code and no longer scheduling the job was not enough to remove the old recurring entry.

Instead of hardcoding one Trakt-specific cleanup path, this change generalizes the startup reconciliation logic so Shoko removes persisted recurring jobs that are no longer part of the current supported recurring schedule set.

## Changes

Recurring startup reconciliation
- extend `QuartzStartup.ScheduleRecurringJobs` to reconcile persisted recurring Quartz jobs before registering current recurring jobs
- keep the existing startup scheduling flow for currently supported recurring jobs

Obsolete recurring job cleanup
- inspect persisted Quartz jobs across all groups
- detect recurring jobs by trigger type:
  - cron triggers
  - simple triggers with repeats
- remove recurring jobs that are not part of the current supported recurring job set
- keep cleanup scoped to recurring jobs only

Supported recurring job set
- define the currently supported recurring startup jobs explicitly from the current build
- include the existing `CheckNetworkAvailabilityJob` recurring schedule in that supported set
- make the cleanup extensible to future removed recurring jobs without adding per-job legacy cleanup code

Safety and behavior
- do not touch one-shot queued jobs stored in Quartz
- do not introduce noop placeholder jobs for removed features
- remove obsolete recurring jobs from persisted Quartz state on startup instead of allowing them to linger indefinitely

Implementation details
- add Quartz group matching to enumerate persisted jobs
- add recurring-trigger detection helper logic in `QuartzStartup`
- keep the existing typed recurring removal helper for cases where typed removal is still useful

## Result

Shoko now removes stale persisted recurring Quartz jobs that should no longer exist after upgrades.

This fixes the Trakt-removal case and provides a more maintainable startup cleanup path for future recurring jobs that may be removed later.


da3dsoul:
Closed in favor of a more flexible batch method

PR #1335, opened 2026-06-04
misc: update devcontainer to .NET 10 / Ubuntu 24.04 Noble [skip ci]

Foowy:
- Upgrade base image from dotnet:8.0-jammy to dotnet:10.0-noble
- Switch apt-packages feature from devcontainers-contrib to devcontainers-extra
- Remove stale .NET 8 runtime/aspnet version pins from dotnet feature
- Use apt-get and && chains in setup commands for reliability
- Drop sudo from dotnet dev-certs (--trust removed; cert gen doesn't need elevation)
- Add recommended VS Code extensions to customizations


da3dsoul:
Interesting. Looks good

PR #1336, opened 2026-06-04
add GHA Docker layer cache to all three build workflows

Foowy:
Add cache-from and cache-to with type=gha to the docker/build-push-action step in build-daily.yml, build-release.yml, and docker-manual.yml. All three use scope=docker-${{ matrix.arch }} to keep amd64 and arm64 caches separate while sharing the same cache pool across workflows — daily builds warm the cache for release and manual builds. Runtime stage layers (apt, GPG key setup, mediainfo install) will always be cache hits; the build stage will hit when no source files changed


da3dsoul:
Yes please

Issue #1239, opened 2025-04-24
Database script fails on case sensitive SQL Server


migoldfinger:
### Before submitting your issue, please verify the following.

- [x] You're using the latest **Stable** or **Daily** version of Shoko Server.
- [x] You're using the latest and correct (Stable:Stable, Daily:Daily) Web UI for the version of Shoko you're using.
- [x] You've searched this repository (top of the page) for the issue you are reporting.

### Shoko Version

5.1.0.0

### Install Type

Windows

### Description

I tried setting up Shoko on my MSSQL Server instance.
My Server Is setup to be case sensitive on DatabaseObjects.
The Script fails when trying to alter the table `jmmuser`  which is clear since the correct table name is `JMMUser`

### Steps to Reproduce

Setup a case sensitive SQL server.
Change the shoko config to use sql server
Start shoko.

### Log File

```
[...]

PR #1339, opened 2026-06-06
Normalizes Hash Casing and Updates SQL Server Queries - NEEDS REVIEW BEFORE COMMIT

Foowy:
Addresses inconsistencies in hash value casing across the database and improves SQL Server metadata queries.

*   **Standardizes hash storage and lookups:**
    *   Introduces a database migration (163) that converts all existing hash values and hash types in relevant tables to uppercase. This establishes a consistent format for stored hashes.
    *   Updates repository lookup methods to normalize input hash values and types to uppercase. This ensures that queries and cache lookups correctly match the standardized uppercase values, regardless of the input casing.
*   **Refines SQL Server catalog queries:**
    *   Replaces older `INFORMATION_SCHEMA` and `sysobjects` views with `sys.*` catalog views in SQL Server for improved compatibility and efficiency when checking database schema.
*   **Enhances test coverage:**
    *   Adds new unit tests specifically for hash normalization within `VideoLocal_HashDigestRepository`, verifying the robustness of case-insensitive lookups after the changes.

Ongoing changes around https://github.com/ShokoAnime/ShokoServer/issues/1239

### DIFF
--- a/Shoko.Server/Databases/SQLServer.cs
+++ b/Shoko.Server/Databases/SQLServer.cs
@@ -185,7 +185,7 @@ public class SQLServer(SystemService systemService) : BaseDatabase<SqlConnection
 
     public override bool HasVersionsTable()
     {
-        const string Sql = "SELECT Count(1) FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = 'Versions'";
+        const string Sql = "SELECT Count(1) FROM sys.tables WHERE name = 'Versions'";
         using var connection = new SqlConnection(GetConnectionString());
         var command = new SqlCommand(Sql, connection);
         connection.Open();
@@ -197,13 +197,13 @@ public class SQLServer(SystemService systemService) : BaseDatabase<SqlConnection
     {
         ConnectionWrapper(GetConnectionString(), myConn =>
         {
-            var create = ExecuteScalar(myConn, "Select count(*) from sysobjects where name = 'Versions'") is 0;
+            var create = ExecuteScalar(myConn, "SELECT count(*) FROM sys.objects WHERE name = 'Versions'") is 0;
             if (create)
             {
                 SystemService.StartupMessage = "Database - Creating Initial Schema...";
                 ExecuteWithException(myConn, _createVersionTable);
             }
-            var updateVersionTable = ExecuteScalar(myConn, "SELECT count(*) FROM INFORMATION_SCHEMA.COLUMNS WHERE [TABLE_NAME] = 'Versions' and [COLUMN_NAME]='VersionRevision'") is 0;
+            var updateVersionTable = ExecuteScalar(myConn, "SELECT count(*) FROM sys.columns WHERE object_id = OBJECT_ID('Versions') AND name = 'VersionRevision'") is 0;
             if (updateVersionTable)
             {
                 ExecuteWithException(myConn, _updateVersionTable);
--- a/Shoko.Server/Databases/SQLServer.cs
+++ b/Shoko.Server/Databases/SQLServer.cs
@@ -953,6 +953,13 @@ public class SQLServer(SystemService systemService) : BaseDatabase<SqlConnection
         new(162,  2, "ALTER TABLE VideoLocal_User ADD LastAudioStreamIndex int NULL;"),
         new(162,  3, "ALTER TABLE VideoLocal_User ADD LastSubtitleStreamIndex int NULL;"),
         new(162,  4, "ALTER TABLE VideoLocal_User ADD ClientData nvarchar(max) NULL;"),
+        new(163,  1, "UPDATE VideoLocal SET Hash = UPPER(Hash)"),
+        new(163,  2, "UPDATE VideoLocal_HashDigest SET Type = UPPER(Type)"),
+        new(163,  3, "UPDATE VideoLocal_HashDigest SET Value = UPPER(CAST(Value AS NVARCHAR(MAX))) WHERE Type IN ('ED2K', 'CRC32', 'MD5', 'SHA1', 'SHA256', 'SHA512')"),
+        new(163,  4, "UPDATE CrossRef_File_Episode SET Hash = UPPER(Hash) WHERE Hash IS NOT NULL"),
+        new(163,  5, "UPDATE FileNameHash SET Hash = UPPER(Hash)"),
+        new(163,  6, "UPDATE StoredReleaseInfo SET ED2K = UPPER(ED2K)"),
+        new(163,  7, "UPDATE StoredReleaseInfo_MatchAttempt SET ED2K = UPPER(ED2K)"),
     ];
 
     #endregion


da3dsoul:
Seems like a lot of redundancy here. If you write a migration to fix the casing, and you fix the write paths, then there's no reason to waste cycles in the read path

Foowy:
This will definitely need outside review as I am touching SQL calls outside of specifically SQL Server. This is for supporting case-sensitive SQL servers while maintaining existing support for case-insensitive SQL servers.

As to @[migoldfinger](https://github.com/migoldfinger) request:

"Personally, I would have loaded the Database Object names into a static class with properties and used this ever time to reference the objects. That way same name over all SQL Queries would be ensured"

I don't think i can achieve this as it would risk breaking existing installations/migrations due to need to modify all existing patches, including modifications to MySQL.cs and SQLite.cs. While a good idea, it is beyond my scope of expertise and would require all primary maintainers to sign off on due to impact.

da3dsoul:
We're moving to EF next version, so that's unnecessary. Our database system is enough of a mess to scrap it, but we need a version to start from...and we keep doing other things. Now we have AI, and I'm willing to spend $200 on fixing that.

da3dsoul:
Normalizing data is a prerequisite though, so yes, keep doing this in general

Foowy:
Testing removed as test invalidated after recent read-path changes.

Shoko.Server/Databases/DatabaseFixes.cs
```diff
@@ -2646,4 +2646,23 @@ private class DNF_UserAvatar
...
         public string AvatarImageMetadata { get; set; }
     }
+
+    public static void UppercaseHashValues()
+    {
+        using var session = ISystemService.StaticServices.GetRequiredService<DatabaseFactory>().SessionFactory.OpenStatelessSession();
+        using var trans = session.BeginTransaction();
+        session.CreateSQLQuery("UPDATE VideoLocal SET Hash = UPPER(Hash)").ExecuteUpdate();
+        session.CreateSQLQuery("UPDATE VideoLocal_HashDigest SET Type = UPPER(Type)").ExecuteUpdate();
+        session.CreateSQLQuery("UPDATE VideoLocal_HashDigest SET Value = UPPER(CAST(Value AS NVARCHAR(MAX)))").ExecuteUpdate();
```
revam: Not all hash values are case insensitive. E.g. if we had a base64 encoded or raw byte string hash, then this would kill it.

Shoko.Server/Databases/DatabaseFixes.cs
```diff
@@ -2646,4 +2646,23 @@ private class DNF_UserAvatar
         public byte[] AvatarImageBlob { get; set; }
         public string AvatarImageMetadata { get; set; }
     }
+
+    public static void UppercaseHashValues()
+    {
+        using var session = ISystemService.StaticServices.GetRequiredService<DatabaseFactory>().SessionFactory.OpenStatelessSession();
+        using var trans = session.BeginTransaction();
+        session.CreateSQLQuery("UPDATE VideoLocal SET Hash = UPPER(Hash)").ExecuteUpdate();
+        session.CreateSQLQuery("UPDATE VideoLocal_HashDigest SET Type = UPPER(Type)").ExecuteUpdate();
```
revam: Not all types are necessarily upper case in the current system. They're literally plugin defined.

Shoko.Server/Databases/DatabaseFixes.cs
```diff
@@ -2646,4 +2646,23 @@ private class DNF_UserAvatar
...
+        session.CreateSQLQuery("UPDATE VideoLocal_HashDigest SET Value = UPPER(CAST(Value AS NVARCHAR(MAX)))").ExecuteUpdate();
+        session.CreateSQLQuery("UPDATE CrossRef_File_Episode SET Hash = UPPER(Hash) WHERE Hash IS NOT NULL").ExecuteUpdate();
+        session.CreateSQLQuery("UPDATE FileNameHash SET Hash = UPPER(Hash)").ExecuteUpdate();
+        // AniDB_File has no NHibernate mapping and does not exist on fresh installs.
+        // CRC/MD5/SHA1 columns were dropped at patch 92.1 so only Hash remains.
+        session.CreateSQLQuery("IF OBJECT_ID('AniDB_File', 'U') IS NOT NULL UPDATE AniDB_File SET Hash = UPPER(Hash)").ExecuteUpdate();
```
revam: That table shouldn't exist so no need to update it. It is also dropped in DatabaseFixes above iirc.

Shoko.Server/Databases/DatabaseFixes.cs
```diff
@@ -2646,4 +2646,23 @@ private class DNF_UserAvatar
...
+        session.CreateSQLQuery("UPDATE StoredReleaseInfo SET ED2K = UPPER(ED2K)").ExecuteUpdate();
+        session.CreateSQLQuery("UPDATE StoredReleaseInfo_MatchAttempt SET ED2K = UPPER(ED2K)").ExecuteUpdate();
+        // AniDB_FileUpdate has no NHibernate mapping and does not exist on fresh installs.
+        session.CreateSQLQuery("IF OBJECT_ID('AniDB_FileUpdate', 'U') IS NOT NULL UPDATE AniDB_FileUpdate SET Hash = UPPER(Hash)").ExecuteUpdate();
```
revam: See my comment above about AniDB_File.

Shoko.Server/Databases/SQLServer.cs
```diff
@@ -953,6 +953,7 @@ public override void CreateAndUpdateSchema()
         new(162,  2, "ALTER TABLE VideoLocal_User ADD LastAudioStreamIndex int NULL;"),
         new(162,  3, "ALTER TABLE VideoLocal_User ADD LastSubtitleStreamIndex int NULL;"),
         new(162,  4, "ALTER TABLE VideoLocal_User ADD ClientData nvarchar(max) NULL;"),
+        new(163,  1, DatabaseFixes.UppercaseHashValues),
```
revam: Fixes ran against a single database shouldn't be added to the shared migration class, they should be appended to the database driver file itself.

revam (changes requested):
I see a few issues with this.

revam:
I'll also add that, if the migration can be done purely in SQL, then it doesn't need a C# helper method running said SQL.

Shoko.Server/Databases/DatabaseFixes.cs
Foowy: Artifact of testing file that has since been dropped. Testing was based off of initial build and was not accounting for playback of drops, this was user error on my side

Shoko.Server/Databases/DatabaseFixes.cs
Foowy: same as above

Shoko.Server/Databases/DatabaseFixes.cs
revam: If MS SQL Server is defining the column as case-insensitive then it needs to be ensured it is always case-sensitive.

Foowy:
> I'll also add that, if the migration can be done purely in SQL, then it doesn't need a C# helper method running said SQL.


Agreed—ideally, the end user would run it against their own box separately and keep Shoko Server out of the equation. Since some end users don’t seek out specific fixes, this helps reduce support requests.

Just a general question, as I am not (and honestly have a lot of other things on my plate) a part of internal discussions: Should I focus my free time on other issues? I’m not intending to sound ungrateful; I just realized an issue ticket I had touched was still open and was trying to see it through to some completion. I’m fine with dropping the PR, as it was focused on the case‑sensitive issue—and that on its own is rough, since I don’t normally use case‑sensitive setups and only ventured down that path to try supporting it. I agree with @[da3dsoul](https://github.com/da3dsoul) about normalizing data, but I feel this may be a wasted PR, as the target audience is exceptionally small and the overhead may be unnecessary.

Shoko.Server/Databases/DatabaseFixes.cs
da3dsoul: @revam by default all SQL Server and SQLite columns are case insensitive

revam:
> > I'll also add that, if the migration can be done purely in SQL, then it doesn't need a C# helper method running said SQL.
> 
> Agreed—ideally, the end user would run it against their own box separately and keep Shoko Server out of the equation. Since some end users don’t seek out specific fixes, this helps reduce support requests.
> 
> Just a general question, as I am not (and honestly have a lot of other things on my plate) a part of internal discussions: Should I focus my free time on other issues? I’m not intending to sound ungrateful; I just realized an issue ticket I had touched was still open and was trying to see it through to some completion. I’m fine with dropping the PR, as it was focused on the case‑sensitive issue—and that on its own is rough, since I don’t normally use case‑sensitive setups and only ventured down that path to try supporting it. I agree with @[da3dsoul](https://github.com/da3dsoul) about normalizing data, but I feel this may be a wasted PR, as the target audience is exceptionally small and the overhead may be unnecessary.

I meant that as, you can move the SQL you ran in the C# helper in DatabaseFixes directly in the driver's updates table, like how we write other raw SQL upgrades.

Telling the end-user to run SQL queries against their database outside the system is not really something we should do unless they themselves brought the issues they have into existence by manually fiddling with their SQL database outside the system.

da3dsoul:
@Foowy I would take the advice from this, upper case existing data to normalize it directly in SQL, and enforce case sensitivity on hash values, as revam made a good point for the future hash providers. If we don't uppercase it first, then the comparisons won't work in the future.

You can use SQL directly via `new(1,1, "UPDATE Table SET Something = UPPER(Something);")`
The database fixes thing is for when that would be hard or impossible.

Shoko.Server/Databases/DatabaseFixes.cs
revam: that's... troubling if they're running transformations on the inserted data and not just applying comparison rules... will take note, and in that case, we meed to enforce it on a lot more columns across all providers.

Foowy:
Removed artifacts, pulled databasefixes, isolated down to SQLServer.cs

Is this more inline to what you both were saying?

Shoko.Server/Databases/SQLServer.cs
```diff
@@ -953,6 +953,13 @@ public override void CreateAndUpdateSchema()
         new(162,  2, "ALTER TABLE VideoLocal_User ADD LastAudioStreamIndex int NULL;"),
         new(162,  3, "ALTER TABLE VideoLocal_User ADD LastSubtitleStreamIndex int NULL;"),
         new(162,  4, "ALTER TABLE VideoLocal_User ADD ClientData nvarchar(max) NULL;"),
+        new(163,  1, "UPDATE VideoLocal SET Hash = UPPER(Hash)"),
+        new(163,  2, "UPDATE VideoLocal_HashDigest SET Type = UPPER(Type)"),
+        new(163,  3, "UPDATE VideoLocal_HashDigest SET Value = UPPER(CAST(Value AS NVARCHAR(MAX)))"),
```
da3dsoul: @revam do we need to filter on type or something? Idk if we have any plugin providers that care yet

Foowy: Added filtering to just what Shoko Server handles, should help avoid plugin requirements. Plugins may play with data and ignore casing.

da3dsoul:
This part looks good. I'll do some research myself today, as the case sensitivity thing will affect multiple database types and a more columns.

Foowy:
Apologies, my commits were looking like a mess, flattened and only pulling recent changes

da3dsoul:
Aw look at you cleaning up the git history lol. We never have history with "typo fix" "missed a spot" "missed another spot" "damnit" commits

PR #1340, opened 2026-06-06
fix(hashing): normalize SHA512 to uppercase in `CalculateHashesRhash`

Foowy:
`SHA512` was the only hash type in the RHash path missing `.ToUpperInvariant()`.
All other types (ED2K, CRC32, MD5, SHA1, SHA256) already normalized their
output; this brings SHA512 in line.

### DIFF
--- a/Shoko.Server/Hashing/CoreHashProvider.cs
+++ b/Shoko.Server/Hashing/CoreHashProvider.cs
@@ -340,7 +340,7 @@ public class CoreHashProvider(ILogger<CoreHashProvider> logger, ConfigurationPro
         {
             Native.rhash_print(output, ctx, RHashIds.RHASH_SHA512, RhashPrintSumFlags.RHPR_DEFAULT);
             var sha512 = Marshal.PtrToStringAnsi(output)!;
-            hashes.Add(new HashDigest { Type = "SHA512", Value = sha512 });
+            hashes.Add(new HashDigest { Type = "SHA512", Value = sha512.ToUpperInvariant() });
         }
 
         Marshal.FreeHGlobal(output);


da3dsoul:
Why another PR? You could've pushed it to the other one

Foowy:
Since was single liner for existing consistency, didn't want it to get trapped in a possible no go PR.

Cazzar:
just a thought, instead of enforcing hashes to uppercase - should we look at making the hash matching case-insensitive?

to me that fixies the root of the problem, and makes it less likely to happen elsewhere.

Foowy:
C# is hashing to uppercase starting at line 248, but RHash SHA‑512 (changed in my PR) was missing this for consistency. My concern is that in the future, end users—looking at you, @migoldfinger—who use a case‑sensitive database could encounter duplicate entries or failed queries due to mismatched casing.

Cazzar:
> who use a case‑sensitive database

But the thing is, we control the indexes, we control the constraints, so we can do those checks in DB to ensure that those cases aren't the case. 

My comment stems down to the fact, that if in future we move away from RHash to something else, it has the problem that this bug can recur, so if in the root of things, the hash matching (wherever needed) is case insensitive, that's fine.

The main concern with case sensitive DB's is the relation names, and the column names need to be consistently designed in code, I ran across this with MySQL/MariaDB in Linux when we added support for linux.

Foowy:
If planning to drop case‑sensitivity support or leave it up to the end user, then I’m on board with handling it in CI. It keeps things simple and avoids extra overhead for a feature most people don’t really use

Building to both is painful.

Issue #1329, opened 2026-05-22
Windows tray icon tooltip does not dismiss


seth-rah:
### Before submitting your issue, please verify the following.

- [x] You're using the latest **Stable** or **Daily** version of Shoko Server.
- [x] You're using the latest and correct (Stable:Stable, Daily:Daily) Web UI for the version of Shoko you're using.
- [x] You've searched this repository (top of the page) for the issue you are reporting.

### Shoko Version

5.3.3.0 (8924611)

### Install Type

Windows 11

### Description

The Shoko Server windows tray icon uses a non-standard tooltip implementation 

This implementation encounters issues where it will show the tooltip for shoko server when shoko server was never engaged with. At times a fresh boot of windows can have the shoko server tooltip spawning at the top left of the display, and other times it will remain persistent because the cursor flew past it while trying to access something else in the system tray.

Suggestion is to make use of the native windows tooltip capability, as the implementation seems to be custom. It's the only tooltip I have that behaves this way, and also does not adhere to system styling preferences.

### Steps to Reproduce

- Open the windows system tray
- Flick your cursor past the shoko server icon
- should render tooltip after a short delay that does not dismiss unless reinvoked

### Log File

N/A

### Screenshots

https://github.com/user-attachments/assets/a1aa4841-7b17-4b23-8d4e-c472601191b1

revam:
The implementation in 5.3.3.0 is the native tooltip impl. in .NET 8.0 Desktop, not the "custom" impl. from our use of the cross platform Avalonia framework on master.

revam:
Stable; https://github.com/ShokoAnime/ShokoServer/blob/releases/v5.x.y/Shoko.TrayService/App.xaml.cs#L76

Daily; https://github.com/ShokoAnime/ShokoServer/blob/master/Shoko.TrayService/App.axaml.cs#L70

revam:
Not really much we can do with it on stable,  considering it'susing the bog standard native taskbar icon, and we've also replaced the impl. on daily, which may or may not fare better compared to the older 'native' menu impl..

seth-rah:
That's kind of strange

It's the only taskbar icon I have that uses a white tooltip instead of a dark one like the rest of my system.

It's also the only tooltip that has this persistence issue where it fails to dismiss correctly.

PR #1343, opened 2026-06-06
fix: tray tooltip persistence and dark mode color on Windows 11 - Untested on Windows - Needs Review

Foowy:
Hardcodet.NotifyIcon.Wpf creates a WPF ToolTip popup instead of using the native Win32 tooltip. When the cursor flicks past the icon quickly, NIN_POPUPCLOSE arrives before the popup is open (no-op), then NIN_POPUPOPEN fires and shows the popup with no matching close event, leaving it stuck on screen indefinitely.

Fix 1: subscribe to TrayToolTipOpen/Close and run a 5-second DispatcherTimer that closes TrayToolTipResolved if Hardcodet never receives a matching NIN_POPUPCLOSE.

Fix 2: detect system dark mode via AppsUseLightTheme registry key and register a matching implicit ToolTip style so the popup colour matches the rest of the taskbar instead of using WPF's default light theme.

Related to https://github.com/ShokoAnime/ShokoServer/issues/1329

### DIFF
--- a/Shoko.TrayService/App.xaml.cs
+++ b/Shoko.TrayService/App.xaml.cs
@@ -7,7 +7,10 @@ using System.Threading;
 using System.Threading.Tasks;
 using System.Windows;
 using System.Windows.Controls;
+using System.Windows.Media;
+using System.Windows.Threading;
 using Hardcodet.Wpf.TaskbarNotification;
+using Microsoft.Win32;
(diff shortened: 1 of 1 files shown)


Foowy:
appeasing to sonarcloud god... dont know why but he was angry with my last commit

da3dsoul:
Pulled it down and tested. works fine

PR #1345, opened 2026-06-07
fix(db): make queue migrations run on SQL Server and MySQL

Foowy:
EF Core filters migrations by exact DbContext type via [DbContext] attribute. All migration Designer files are tagged with [DbContext(typeof(SqliteQueueDbContext))], so SqlServerQueueDbContext and MySqlQueueDbContext found zero applicable migrations — MigrateAsync silently no-oped, Jobs table was never created, and WorkerPoolManager crashed on LoadAllAsync.

Secondary issue: migration files hard-coded SQLite affinity type strings (TEXT, INTEGER). SQL Server rejects TEXT as a PRIMARY KEY type, so even if migrations were discovered they would have failed at CREATE TABLE.

- Add QueueMigrationsAssembly: extends MigrationsAssembly, overrides Migrations to find SqliteQueueDbContext-tagged migrations regardless of the current provider context
- Register via ReplaceService in SqlServerQueueDbContext and MySqlQueueDbContext
- Remove explicit type strings from InitialCreate and AddJobChainContext migrations; EF Core now infers provider-appropriate DDL (uniqueidentifier/nvarchar/bigint on SQL Server, char(36)/varchar/bigint on MySQL, TEXT/INTEGER on SQLite unchanged)

### DIFF
--- a/Shoko.QueueProcessor.Tests/MultiProviderMigrationTests.cs
+++ b/Shoko.QueueProcessor.Tests/MultiProviderMigrationTests.cs
@@ -1,0 +1,35 @@
+using System.Linq;
+using Microsoft.EntityFrameworkCore;
+using Shoko.QueueProcessor.Storage;
+using Xunit;
+
+namespace Shoko.QueueProcessor.Tests;
+
+/// <summary>
+/// Verifies that non-SQLite provider contexts expose the same migration set as
+/// <see cref="SqliteQueueDbContext"/> via <see cref="QueueMigrationsAssembly"/>.
+/// <para>
+/// The root cause this guards against: EF Core's default <c>IMigrationsAssembly</c>
+/// filters by exact <c>[DbContext]</c> type. All migration Designer files are tagged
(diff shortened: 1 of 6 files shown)


Foowy:
Discovered this issue when configuring Shoko to point at fresh MS SQL database.

Example of the case:

    ShokoServer Docker container spun up and configured using the default SQLite instance

    Stop container

    Configure the database and queue to point to the MS SQL Server

    Start container

    A crash loop occurred

When both Database and Queue pointed at the same ShokoServer SQL Server database, both EF Core contexts share the same __EFMigrationsHistory table. The main database context ran its migrations first and populated that table. When the queue context started up, it saw the history table already existed, assumed its own migrations were already applied, and skipped creating the Jobs table — then immediately tried to query it and crashed.

This is likely user error on my part by not splitting into 2 separate DB's , but in the process of investigating I came across other gotchas.

Even maintaining split, we would see issues with queue processor being migrated to a DB that wasn't SQLite.

da3dsoul:
TIL I can push to PRs lol

PR #1346, opened 2026-06-07
fix(db): use provider-appropriate column types in queue migrations

Foowy:
The QueueMigrationsAssembly.ModelSnapshot returned the SQLite snapshot, whose HasColumnType("TEXT") annotations were passed to the EF Core SQL generator as model context. GetColumnType() finds properties via FindProperty(model, ...) before falling back to the CLR type mapper, so it used TEXT for every column — SQL Server rejects TEXT as a primary-key type, causing InitialCreate to fail with a CREATE TABLE error, leaving the Jobs table nonexistent and crashing WorkerPoolManager.StartAsync.

Three-layer fix:

- QueueMigrationsAssembly.ModelSnapshot now returns null. This class is only registered on non-SQLite contexts; returning null ensures EF Core always falls through to the provider's CLR type mapper (uniqueidentifier, nvarchar, bigint, bit, datetimeoffset on SQL Server; char(36), varchar, bigint, tinyint(1), datetime(6) on MySQL).

- InitialCreate and AddJobChainContext migrations now set explicit column types via ActiveProvider so GetColumnType() short-circuits on operation.ColumnType before the model lookup, making them immune to model-annotation bleed.

- InitialCreate used bigint for QueuedAt/ScheduledAt on all providers. SQLite is correct (DateTimeOffset stored as unix milliseconds via ValueConverter), but SQL Server and MySQL have no such converter so EF Core hits error 206 (operand type clash) at insert time. InitialCreate now uses datetimeoffset(7)/datetime(6) for those providers. FixQueueTimestampColumns migration idempotently alters existing deployments.

- SQL Server Down(): DROP COLUMN fails when a named DEFAULT constraint is still attached. Conditional DROP CONSTRAINT for DF_Jobs_QueuedAt added before DropColumn so rollback works correctly.

- MySQL Up(): MODIFY COLUMN from bigint to datetime(6) fails in strict mode (default since MySQL 5.7) when rows contain Unix-millisecond integers. Jobs table is cleared before the type change — Jobs is an ephemeral crash-recovery store; surviving rows are re-queued on next startup, so clearing is safe.


da3dsoul:
One of my concerns here is that migrations are typically auto generated via `dotnet ef migration add`. I'm not sure if modifying the migrations so heavily will cause inconsistencies. Need to look into it

Foowy:
understand that. I have zero issues with review. Going through fixes as i test the process of swapping to MSSql. More or less  bug/error correction through testing process

Foowy:
Closing to resubmit with consolidated and optimised commit.

PR #1348, opened 2026-06-08
fix(tmdb): add jitter to local rate-limit retry delay

Foowy:
`UseClient` retries on `RateLimitRejectedException` by waiting `RetryAfter` then immediately re-attempting. With up to 6 concurrent callers (bulkhead limit), all competing callers wake up at nearly the same instant when `RetryAfter` is milliseconds-scale — only one wins the token, the rest get rejected and retry again. Observed result: a single call accumulating 60+ attempts before finally getting a slot.

Adds a random 0–50 ms jitter on top of `RetryAfter` so concurrent callers spread their wakeup times and stop competing for the same token simultaneously.

### DIFF
--- a/Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
+++ b/Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
@@ -320,9 +320,11 @@ public class TmdbMetadataService : ITmdbMetadataService
                 switch (ex)
                 {
                     // If we got a _local_ rate limit exception, wait and try again.
+                    // Jitter is added so concurrent callers don't all wake up simultaneously
+                    // and compete for the same token (thundering herd).
                     case RateLimitRejectedException rlrEx:
                     {
-                        var retryAfter = rlrEx.RetryAfter;
+                        var retryAfter = rlrEx.RetryAfter + TimeSpan.FromMilliseconds(Random.Shared.Next(0, 50));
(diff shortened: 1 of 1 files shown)


da3dsoul:
This feels like a band aid. I'm not sure how TMDB's limit works, but a class that actually keeps track of calls and limits effectively as we do with AniDB seems better. I've briefly discussed this with @revam. I'm fine with merging this, as it'll make things better until then, but I don't want the illusion that it's fixed if it's not

Foowy:
Agreed, i will explore tracking TMDB queries. It does feel like a bandaid, but trying to limit how much outside data i am touching as we seem to be working on the same area of code.

Looks like TMDB will trip a 429 error if you go above their limit, They do not have a disclosed maximum other than around ~40/s. Might be able to build it like how AniDB is handled. So far i can't find a reset time frame. But at least it is something to go on.

revam:
> This feels like a band aid. I'm not sure how TMDB's limit works, but a class that actually keeps track of calls and limits effectively as we do with AniDB seems better. I've briefly discussed this with @revam. I'm fine with merging this, as it'll make things better until then, but I don't want the illusion that it's fixed if it's not

It seems like a band aid on top of the existing three band aids. It might improve the experience until we get proper tracking, or it might not. I'll leave it to you if you want to merge this before the refactor or not. @da3dsoul

PR #1349, opened 2026-06-08
refactor(tmdb): replace polly rate limit with sliding-window TmdbRateLimiter

Foowy:
Replaced `Policy.RateLimitAsync` (token bucket, 45 req/10s) with a dedicated `TmdbRateLimiter` singleton that tracks real request timestamps in a 1-second sliding window.

- Added `TmdbRateLimiter` with `EnsureRateAsync<T>`, `NotifyRateLimitExceeded`, and `CallsInWindow`/`RemainingInWindow` stats
- Added `TmdbRateLimitSettings` (default: 30 req/s window) nested under `TMDBSettings.RateLimit`, configurable at runtime
- Fixed the `RequestLimitExceededException` handler which previously did not wait at all before retrying on a 429 response; it now calls `NotifyRateLimitExceeded` and waits `retryAfter + jitter`
- Removed `RateLimitRejectedException` from the retry policy (no longer thrown since the Polly rate limit policy is gone)
- Jitter on all wait paths prevents thundering herd when concurrent callers hit the window limit or a server backoff simultaneously
- Added `TmdbRateLimiterTests` covering sliding-window throttling, backoff propagation, concurrent caller behaviour, and stats properties

### DIFF
--- a/Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
+++ b/Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
@@ -317,14 +317,17 @@ public class TmdbMetadataService : ITmdbMetadataService
                 // Retry on rate limit exceptions, throw on everything else.
                 switch (ex)
                 {
-                    // If we got a _remote_ rate limit exception, notify the rate limiter and wait.
+                    // If we got a _remote_ rate limit exception, notify the rate limiter.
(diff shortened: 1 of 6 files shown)


Foowy:
Yay, not a band-aid

Foowy:
Closing to squash commits and resubmit with improved PR description.

PR #1352, opened 2026-06-09
refactor(tmdb): replace Polly rate limiter with BCL SlidingWindowRateLimiter

Foowy:
## Summary

Replaces the Polly `AsyncRateLimitPolicy` + retry-loop with a purpose-built `TmdbRateLimiter` backed by the BCL `SlidingWindowRateLimiter`, fixing several stability and correctness issues in the process. Also parallelizes season fetching and `RepairMissingPeople` for faster show updates.

## Motivation

- **Thundering herd on local rate limit**: all concurrent callers woke simultaneously after a `RateLimitRejectedException`, causing a single call to accumulate 60–70 retry attempts before getting a slot.
- **Silent exception swallowing**: `ProcessWithConcurrencyAsync` wrapped `Task.WhenAny` in a `try/catch` — `Task.WhenAny` never throws, so all worker failures disappeared silently.
- **Unbounded 429 retries**: `WaitAndRetryAsync(int.MaxValue, …)` had no cap for `RequestLimitExceededException`; a suspended API key would spin forever.
- **Sequential season fetching**: each `GetTvSeasonAsync` call blocked the next; a 10-season show made 10 serial API round-trips.

### `TmdbRateLimiter` (new)
- `SlidingWindowRateLimiter` (BCL, no new NuGet reference on .NET 10) replaces `AsyncRateLimitPolicy`.
- Server 429 backoff via `NotifyRateLimitExceeded`; all callers observe it through `WaitForBackoffAsync` with per-caller jitter. Backoff is re-checked after acquiring a slot to close the race window where a concurrent 429 could arrive between the two calls.
- `_disposeCts` token passed to `WaitForBackoffAsync` and `AcquireAsync` — shutdown produces `OperationCanceledException` instead of `ObjectDisposedException`.
- `OnSettingsSaved` defers disposal of the old limiter by 15 s so in-flight `AcquireAsync` calls drain without leaking the internal replenishment timer.
- `sealed` — satisfies IDisposable S3881; correct since it is a DI singleton.
- `TmdbRateLimitSettings` added to `TMDBSettings` (debug/advanced): `MaxRequestsPerWindow` (default 30) and `WindowDurationMs` (default 1000).

### `TmdbMetadataService`
- `ProcessWithConcurrencyAsync`: fixed `try/catch` placement (was around `WhenAny`, now around `await task`), `List<Task>` → `HashSet<Task>` (O(1) remove), `OperationCanceledException` handling moved to outer loop with token-matched `when` filter, off-by-one fixed (`> maxConcurrent` → `>= maxConcurrent`).
- `_retryPolicy`: retains `Or<GeneralHttpException>()` in the handle chain (required so `OnTmdbRetryAsync` fires for logging before re-throw); removed `RateLimitRejectedException` (now handled by `TmdbRateLimiter`). Remote 429 retries capped at 10.
- `OnTmdbRetryAsync` extracted from constructor lambda.
- Both `UpdatePerson` and `PurgePerson` fan-outs in `UpdateMovieCastAndCrew` / `UpdateShowCastAndCrew` catch `AggregateException` and log a warning — partial cast/crew failures are non-fatal and must not abort the job.
- `PurgeUnlinkedPeople` converted from sequential `foreach` to `ProcessWithConcurrencyAsync`.
- `UpdateShowSeasonsAndEpisodes`: season loop converted from sequential `foreach` to `ProcessWithConcurrencyAsync`; `GetTvSeasonAsync` calls now run concurrently up to `_maxConcurrency`. Uses `ConcurrentBag` for `seasonsToSkip`/`seasonsToSave` and `Interlocked` for all shared counters. Also fixes a pre-existing `episodesToAdd++` race inside the concurrent episode lambda.
- `RepairMissingPeople`: sequential `UpdatePerson` loop replaced with `ProcessWithConcurrencyAsync`.

### `ImportJob`
- Catches `AggregateException` from `RunImport_PurgeUnlinkedTmdbPeople` — a partial purge must not abort the rest of the import pipeline.

### Tests
- 7 unit tests covering window throughput, slot waiting, window expiry, backoff delay, concurrent backoff, and counter properties.

## Test plan

- [x] All 7 `TmdbRateLimiterTests` pass
- [x] Full unit test suite passes (`dotnet test Shoko.Tests/Shoko.Tests.csproj`)
- [x] Integration tests pass (`dotnet test Shoko.IntegrationTests/Shoko.IntegrationTests.csproj`)
- [x] CI green (SQLite / MySQL / SQL Server)
- [x] SonarCloud green (no new issues)


Foowy:
@da3dsoul I believe this might seal the wound, please let me know if this is more in line to your view of what it should be

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -306,53 +340,13 @@ TMDB_Show_NetworkRepository xrefTmdbShowNetwork
...
         _xrefTmdbShowNetwork = xrefTmdbShowNetwork;
         _entityLock = new(logger);
+        _rateLimiter = rateLimiter;
         _instance ??= this;
         _bulkheadPolicy = Policy.BulkheadAsync(_maxConcurrency, int.MaxValue);
-        _rateLimitPolicy = Policy.RateLimitAsync(45, TimeSpan.FromSeconds(10), 45);
         _retryPolicy = Policy
-            .Handle<RateLimitRejectedException>()
-            .Or<HttpRequestException>()
-            .Or<GeneralHttpException>()
```
revam, line 315: Add back the missing `GeneralHttpException` handling.

revam (changes requested):
Other than the GHE handling getting lost somewhere, then it seems fine. Good work. 👍

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
Foowy: added back, apologies i was thinking it was an accidental double call, did not catch that it was allowing polly more logging/exception handling  outside of just the retry.

revam (approved):

Foowy:
Completed out last items i forgot about for concurrency. Still same amount of calls just faster. Like caffeine, still do stupid things... just faster.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -1182,27 +1191,29 @@ public async Task<bool> UpdateShow(TmdbShowUpdateOptions options)
         var episodeEventsToEmit = new ConcurrentDictionary<TMDB_Episode, UpdateReason>();
         var allPeopleToAddOrKeep = new ConcurrentBag<int>();
         var allPeopleToPotentiallyRemove = new ConcurrentBag<int>();
-        foreach (var reducedSeason in show.Seasons!)
+        await ProcessWithConcurrencyAsync(_maxConcurrency, show.Seasons!, async reducedSeason =>
```
revam, line 1194: Seasons were intentionally running in a synchronous loop for better performance and less parallel waiting calls when both running one and when running multiple tmdb series/show updates.

We were running the episodes in parallel for each season so we know they are done before processingthe next season, to try and maintain some order of processing when refreshing/processing a series.

Also don't cram more and more changes unprompted into the PR after a review please.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -1280,17 +1291,14 @@ await ProcessWithConcurrencyAsync(_maxConcurrency, season.Episodes!, async (redu
                 tmdbSeason.HiddenEpisodeCount = hiddenEpisodeCount;
                 seasonUpdated = true;
             }
-            if (seasonUpdated)
+            if (seasonUpdated && !seasonAlreadySaved)
```
revam: This was an intentional so we would update the last updated time if the season data changed, regardless of if it was already in the list to save. Revert please.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -1348,20 +1356,34 @@ await ProcessWithConcurrencyAsync(_maxConcurrency, season.Episodes!, async (redu
         var peopleAdded = 0;
         var peopleUpdated = 0;
         var peoplePurged = 0;
-        var peopleToCheck = allPeopleToAddOrKeep.ToArray().Distinct().ToList();
-        var peopleToPurge = allPeopleToPotentiallyRemove.ToArray().Distinct().Except(peopleToCheck).ToList();
-        foreach (var personId in peopleToCheck)
+        var peopleToCheck = allPeopleToAddOrKeep.Distinct().ToList();
```
revam, line 1360: The first .ToArray() is to collect the bag, converting the ConcurrentBag<int> to an int[], before processing it.

revam (changes requested):
As stated in my first review comment, please don't cram more changes unprompted into a PR after a review. If the PR is not ready, then it should be marked as a draft. Also, please respond to the comments.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
Foowy: Two different checks — keep and purge. The encapsulation in the foreach statement compares which items need to be removed. For example, with Black Clover, we don’t need to download the same person 170 times (once per episode); we only need to download or insert them once. Instead of making 170 calls, we only need one. 

Edit: Clarity

Foowy: had to ask claude for exact technical::

ConcurrentBag<int> implements IEnumerable<T> directly. Calling .ToArray() before .Distinct() materializes the bag into an intermediate array before deduplication — an unnecessary allocation. Distinct() works on any IEnumerable<T> and the bag is only read after all concurrent work is complete, so there's no thread-safety reason for the snapshot either.

The logic itself:
- allPeopleToAddOrKeep — person IDs collected across all episodes that have active cast/crew entries; these need to be upserted.
- allPeopleToPotentiallyRemove — person IDs seen in the existing cast/crew records; candidates for purge if no episode still references them.
- peopleToPurge = potentiallyRemove.Except(toCheck) — only purge people who appear in no episode's current cast/crew. A person in both bags (e.g., they're on one episode but being removed from another) stays.

The dedup via .Distinct() is necessary because both bags accumulate one entry per episode, not per person — Black Clover with 170 episodes could produce 170 duplicate entries for the same person.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
Foowy: reverting this would need to refactor the whole loop. I can guard it with the following to preserve timestamp for comparison. So if newer data exists it will grab.

            if (seasonUpdated && !seasonAlreadySaved)
            if (seasonUpdated)
            {
                tmdbSeason.LastUpdatedAt = DateTime.Now;
                seasonsToSave.Add(tmdbSeason);
                if (!seasonAlreadySaved)
                    seasonsToSave.Add(tmdbSeason);

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
Foowy: Before TMDB Update:  Each assembly line worker did one task at a time — fetch season 1, wait, fetch season 2, wait, fetch season 3...

After TMDB Update: Each assembly line worker now does all the season fetches at the same time, then waits for all of them to finish before moving on.

Foowy: unlike anidb with the 1 api call per 2 seconds, TMDB accepts 1 call per 25ms. With coded setting of 30 per second, we can make calls to tmdb every 33ms. increasing grab rate over current ( 45 requests per 10 seconds,). Import to DB is still serialized, so ultimately still funnels to single input at a time. maintaining endpoint performance, and less waits for data.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
Foowy: I have staged the commit for guard, but not pushed, Let me know if you want guard process or refactor

revam: Make it so we set the last updated timestamp again if the season is updated.

revam (approved):

PR #1355, opened 2026-06-10
fix(tmdb): correct episodesToSave inversion, purge method fixes, and lower default rate limit

Foowy:
Both commits in this branch have been through a code review pass before submission.

## Commit 1 — lower default rate limit

`fix(tmdb): lower default rate limit from 30 to 10 requests per second`

`TmdbRateLimitSettings.MaxRequestsPerWindow` default lowered from 30 to 10. Conservative default to avoid overwhelming end-user hardware with API requests and the data processing each response triggers. The setting remains user-adjustable up to 40 (TMDB's enforced cap).

## Commit 2 — episodesToSave inversion + purge fixes

`fix(tmdb): correct episodesToSave inversion and fix purge method issues`

**`episodesToSave.IsEmpty` inversion** (`TmdbMetadataService.cs`): the check was inverted — `changed = true` was returned when *nothing* changed, causing spurious DB saves and `OnSeriesUpdated` events on every quick-refresh pass.

**`PurgeAllUnusedMovies` / `PurgeAllUnusedShows`**: replaced chained `Concat` iterators with `HashSet.UnionWith`. The old code called `GetAll()` on the AniDB xref table twice (once per Concat call) and produced an `IEnumerable` that re-iterated the source each time it was consumed. The new form builds the set once.

**`PurgeUnlinkedShowNetworks`**: precompute a `HashSet<int>` of linked network IDs once before the loop (previously one DB query per network). Added a re-check after acquiring the per-network lock to prevent purging a network that was re-linked by a concurrent show update between the initial set-build and the purge call. Also fixed a copy-paste log message ("staff members" → "networks").

**`PurgeUnlinkedPeople`**: extracted `GetLinkedPersonIds()` as the single source of truth across the four cast/crew tables. The previous loop issued up to 4 DB queries per person to check linkage; the new form builds the full linked-person set once before the loop.

## Commit 3 — log placeholder casing fix

`chore(tmdb): fix log placeholder casing in PurgeUnlinkedShowNetworks`

Corrected `{count}` to `{Count}` in the orphaned-networks log message in `PurgeUnlinkedShowNetworks`. Structured logging conventions (MEL/Serilog) require PascalCase placeholders; lowercase silently works but produces analyzer warnings.

### DIFF
--- a/Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
+++ b/Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
@@ -2015,7 +2015,7 @@ public class TmdbMetadataService : ITmdbMetadataService
     {
         var linkedNetworkIds = _xrefTmdbShowNetwork.GetAll().Select(x => x.TmdbNetworkID).ToHashSet();
         var networks = _tmdbNetwork.GetAll().Where(p => !linkedNetworkIds.Contains(p.TmdbNetworkID)).ToList();
-        _logger.LogDebug("Checking {count} orphaned networks if they should be purged.", networks.Count);
+        _logger.LogDebug("Checking {Count} orphaned networks if they should be purged.", networks.Count);
         foreach (var network in networks)
             await PurgeShowNetwork(network.TmdbNetworkID);
     }
--- a/Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
+++ b/Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
@@ -701,7 +701,7 @@ public class TmdbMetadataService : ITmdbMetadataService
             peopleAdded,
             peopleUpdated,
             peoplePurged,
-            peopleToPurge.Count + peopleToPurge.Count - peopleAdded - peopleUpdated - peoplePurged,
+            peopleToKeep.Count + peopleToPurge.Count - peopleAdded - peopleUpdated - peoplePurged,
             tmdbMovie.EnglishTitle,
             tmdbMovie.Id
         );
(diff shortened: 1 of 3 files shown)


Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -1220,7 +1215,7 @@ await ProcessWithConcurrencyAsync(_maxConcurrency, show.Seasons!, async reducedS
 
             var episodeBag = new ConcurrentBag<TMDB_Episode>();
             var hiddenEpisodeBag = new ConcurrentBag<TMDB_Episode>();
-            await ProcessWithConcurrencyAsync(_maxConcurrency, season.Episodes!, async (reducedEpisode) =>
+            await ProcessWithConcurrencyAsync(1, season.Episodes!, async (reducedEpisode) =>
```
da3dsoul, line 1218: What's the point of process with concurrency 1

Foowy: Initial, forgot to set back from testing. It is a good thing to keep though. Inner loop reprieve on the rate limiter so calls don't stack up too fast.

Concurrency inside Concurrency issue. _maxConcurrency seasons + _maxConcurrency episode just presents to rate limiter as _maxConcurrency^2. So it sees a large burst, backs off to avoid ceiling limit. So the queue moves fast, pauses, and repeats cycle (think of stop and go traffic on freeway)

Leaving episode as 1 is less bursty and queue moves smoother without limiter getting twitchy. No, your not going freeway speed, but you aren't having to stop either.

revam: What was the point of making it into a parallel execution if we're just going to keep it at 1? I'd rather we revert back to the for loop then keep it as process with concurrency at 1.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -2519,13 +2509,13 @@ public async Task<bool> PurgePerson(int personId, bool force = false)
             {
                 if (!person.LastOrphanedAt.HasValue)
                 {
-                    person.LastOrphanedAt = DateTime.Now;
+                    person.LastOrphanedAt = DateTime.UtcNow;
```
revam, line 2512: We store local time in the database, not UTC. also, comparing against Now or UtcNow should produce the same result, but if it doesn't, then Now is the better of the two as it matches the database. Please revert, and take note for the future about any date handling internally. There are some places where we use UTC, but not involving the database unless it is by mistake.

Shoko.Server/Settings/TmdbRateLimitSettings.cs
```diff
@@ -11,13 +11,14 @@ public class TmdbRateLimitSettings
...
-    /// </summary>
+    /// TMDB enforces approximately 40 requests per second; this defaults to 10 to avoid overwhelming
+    /// end-user hardware with API requests and the data processing that follows each one.
+    ///</summary>
     [Badge("Debug", Theme = DisplayColorTheme.Warning)]
     [Visibility(Size = DisplayElementSize.Small, Advanced = true)]
     [Display(Name = "Max Requests Per Window")]
     [Range(1, 40)]
-    public int MaxRequestsPerWindow { get; set; } = 30;
+    public int MaxRequestsPerWindow { get; set; } = 10;
```
revam, line 21: Without a setting migration then the users which upgraded to the server when it was 30 will be locked at 30 now, even if their hardware can only do 10 in practice.

Shoko.Server/Providers/TMDB/TmdbRateLimiter.cs
```diff
@@ -76,7 +76,7 @@ private static SlidingWindowRateLimiter CreateLimiter(int maxRequests, int windo
         {
             PermitLimit = maxRequests,
             Window = TimeSpan.FromMilliseconds(windowMs),
-            SegmentsPerWindow = 1,
+            SegmentsPerWindow = 10,
```
revam, line 79: Should this be surfaced to our own options so we can control it at runtime to further tune it in the future, be it for all users or per setup?

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -1928,23 +1923,19 @@ private List<TitleLanguage> GetLanguages(TitleLanguage? mainLanguage = null) =>
...
-            .Concat(_tmdbAlternateOrderingEpisodes.GetAll().Select(episode => episode.TmdbShowID))
-            .ToHashSet();
-        var toKeep = _xrefAnidbTmdbShows.GetAll()
-            .Select(xref => xref.TmdbShowID)
-            .ToHashSet();
-        var toBePurged = allShows
-            .Except(toKeep)
-            .ToHashSet();
+        var toKeep = _xrefAnidbTmdbShows.GetAll().Select(xref => xref.TmdbShowID).ToHashSet();
+        // Seeded with toKeep so AllCount in the log includes xref-linked IDs that have no TMDB_Show row yet.
```
revam, line 1927: This comment feels overly verbose when reading the code.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -830,21 +830,16 @@ private async Task DownloadMovieImages(int movieId, TitleLanguage? mainLanguage
...
-            .Concat(_xrefTmdbCollectionMovies.GetAll().Select(collectionMovie => collectionMovie.TmdbMovieID))
-            .ToHashSet();
-        var toKeep = _xrefAnidbTmdbMovies.GetAll()
-            .Select(xref => xref.TmdbMovieID)
-            .ToHashSet();
-        var toBePurged = allMovies
-            .Except(toKeep)
-            .ToHashSet();
+        var toKeep = _xrefAnidbTmdbMovies.GetAll().Select(xref => xref.TmdbMovieID).ToHashSet();
+        // Seeded with toKeep so AllCount in the log includes xref-linked IDs that have no TMDB_Movie row yet.
```
revam, line 834: Same as above review comment.

revam (commented):
@da3dsoul Next time can you please wait for my review if it involves TMDB? 🥺

@Foowy Lots of good changes here, but I do have a review for you even if this was merged already.

PR #1356, opened 2026-06-10
chore(tmdb): fix `contentRanting` typo in local variable names

Foowy:
Fixes a typo in two local variable names: `contentRantingLanguages` → `contentRatingLanguages`, in `UpdateMovie` and `UpdateShow`. No behaviour change.


Foowy:
This is also to correct CI check in with my last PR,. My Skip CI call on final commit of PR blocked docker from updating, so change are in repo, but not in the wild.

PR #1357, opened 2026-06-10
chore(tmdb): add inline comments, fix cast/crew orphan purge, skip unchanged seasons/episodes

Foowy:
## Summary

This PR started as inline comment improvements for code from #1352 and #1355, and grew to include several correctness fixes and one feature discovered during review.

### Commits

1. **`chore(tmdb): add inline comments to rate limiter and purge methods`**
   Original commit. Adds doc-comments and inline `// why` comments to `TmdbRateLimiter.cs` and the cast/crew purge paths in `TmdbMetadataService.cs`.

2. **`fix(tmdb): remove orphaned cast/crew entries when person fetch fails`**
   Cast/crew rows are written before person records are fetched. When the fan-out to `UpdatePerson` fails mid-way (transient error, rate limit), person IDs that were never written leave dangling cast/crew references — the API cannot resolve those entries and returns HTTP 500. After catching the exception, the fix deletes cast/crew rows whose person ID has no corresponding `TMDB_Person` row.

3. **`refactor(tmdb): replace episode concurrency with sequential foreach and ActionBlock`**
   Replaces the `SemaphoreSlim`-based episode loop with `ActionBlock<T>` from TPL Dataflow. Individual worker exceptions are collected so all items complete before any failure is re-thrown as `AggregateException`, rather than faulting the block early and abandoning queued items.

4. **`feat(tmdb): expose TMDB_API_Limits as configurable rate limiter setting`**
   Promotes `SegmentsPerWindow` (previously hardcoded to `10`) to a `TmdbRateLimitSettings` property so it can be tuned alongside `MaxRequestsPerWindow` and `WindowDurationMs`.

5. **`feat(tmdb): skip unchanged seasons/episodes using TMDB Changes API`**
   After the freshness gate, calls `GetShowChangedItemsAsync` to fetch which seasons/episodes changed since `LastUpdatedAt`. Skips `GetTvSeasonAsync` / `GetTvEpisodeAsync` for entries not in the change set. Falls back to full refresh when the entity is newly added, `LastUpdatedAt` is older than 14 days, `forceRefresh` is set, or the Changes API returns `null`. Fixes a bug where a `"season"` key change (which adds to `SeasonNumbers` but not `Episodes`) caused all existing episodes in that season to be wrongly skipped.

6. **`chore(tmdb): add unit tests for ParseShowChanges`**
   15 unit tests covering normal, edge, and error cases for the `ParseShowChanges` internal method introduced in commit 5.

## Notes

- PR remains in **draft** — open to feedback on scope before requesting review.

### DIFF
--- a/Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
+++ b/Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Threading.Tasks.Dataflow;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Polly;
@@ -1255,12 +1256,9 @@ public class TmdbMetadataService : ITmdbMetadataService
 
             seasonsToSkip.Add(tmdbSeason.Id);
 
-            var episodeBag = new ConcurrentBag<TMDB_Episode>();
-            var hiddenEpisodeBag = new ConcurrentBag<TMDB_Episode>();
-            // Concurrency 1: episode tasks run inside the outer season loop which already runs at
-            // _maxConcurrency. Allowing episode parallelism too would produce _maxConcurrency² API
-            // calls in flight simultaneously, overwhelming the rate limiter.
-            await ProcessWithConcurrencyAsync(1, season.Episodes!, async (reducedEpisode) =>
+            var episodeBag = new List<TMDB_Episode>();
+            var hiddenEpisodeBag = new List<TMDB_Episode>();
+            foreach (var reducedEpisode in season.Episodes!)
             {
                 _logger.LogDebug("Checking episode {EpisodeNumber} in season {SeasonNumber} for show {ShowTitle} (Show={ShowId})", reducedEpisode.EpisodeNumber, reducedSeason.SeasonNumber, show.Name, show.Id);
                 if (!existingEpisodes.TryGetValue(reducedEpisode.Id, out var tmdbEpisode))
@@ -1317,7 +1315,7 @@ public class TmdbMetadataService : ITmdbMetadataService
                     hiddenEpisodeBag.Add(tmdbEpisode);
                 else
                     episodeBag.Add(tmdbEpisode);
-            });
+            }
 
             var episodeCount = episodeBag.Count;
             var hiddenEpisodeCount = hiddenEpisodeBag.Count;
@@ -1410,7 +1408,7 @@ public class TmdbMetadataService : ITmdbMetadataService
                     Interlocked.Increment(ref peopleUpdated);
             });
         }
-        catch (AggregateException ex)
+        catch (Exception ex)
         {
             // Partial cast/crew failures are non-fatal: a missing person record does not invalidate
             // the show. Log and continue so a transient API error doesn't abort the entire show update.
@@ -1447,7 +1445,7 @@ public class TmdbMetadataService : ITmdbMetadataService
                     Interlocked.Increment(ref peoplePurged);
             });
         }
-        catch (AggregateException ex)
+        catch (Exception ex)
         {
             _logger.LogWarning(ex, "TMDB: Failed to purge one or more people during show cast/crew update (Show={ShowId})", show.Id);
         }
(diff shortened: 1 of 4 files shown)


Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -682,6 +684,8 @@ await ProcessWithConcurrencyAsync(_maxConcurrency, peopleToKeep, async personId
         }
         catch (AggregateException ex)
         {
+            // Partial cast/crew failures are non-fatal: a missing person record does not invalidate the
```
revam: Missing a person entry will lead to 500 response status code when getting the cast from the API, as we have the cast but not the person. And I'm not a fan of hiding the cast entry in the API layer if the person doesn't exist in our local database, nor a fan of making the person on the cast entry in the API layer nullable and having to teach **all** clients how to deal with that, so we need better handing internally of these cases.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -832,6 +837,7 @@ public async Task PurgeAllUnusedMovies()
     {
         var toKeep = _xrefAnidbTmdbMovies.GetAll().Select(xref => xref.TmdbMovieID).ToHashSet();
         // Seeded with toKeep so AllCount in the log includes xref-linked IDs that have no TMDB_Movie row yet.
+        // UnionWith calls each GetAll() exactly once; a Concat chain would re-enumerate the same sources.
```
revam: Again, same as my review on the other PR; I feel this comment overly verbose here. But then again I'm not an LLM so i dunno how they would read this code without comments.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -1215,6 +1221,9 @@ await ProcessWithConcurrencyAsync(_maxConcurrency, show.Seasons!, async reducedS
 
             var episodeBag = new ConcurrentBag<TMDB_Episode>();
             var hiddenEpisodeBag = new ConcurrentBag<TMDB_Episode>();
+            // Concurrency 1: episode tasks run inside the outer season loop which already runs at
```
revam: Again, same as my review on the other PR; if we're to hard code the concurrency at 1 then I would rather we revert back to the synchronous for loop and simplify the code again, instead of having to deal with extra concurrency logic for no added benefit.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -1367,6 +1376,8 @@ await ProcessWithConcurrencyAsync(_maxConcurrency, peopleToCheck, async personId
         }
         catch (AggregateException ex)
         {
+            // Partial cast/crew failures are non-fatal: a missing person record does not invalidate
```
revam: See above review comment.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -1925,6 +1937,7 @@ public async Task PurgeAllUnusedShows()
     {
         var toKeep = _xrefAnidbTmdbShows.GetAll().Select(xref => xref.TmdbShowID).ToHashSet();
         // Seeded with toKeep so AllCount in the log includes xref-linked IDs that have no TMDB_Show row yet.
+        // UnionWith calls each GetAll() exactly once; a Concat chain would re-enumerate the same sources.
```
revam: See above review comment.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -2013,6 +2026,8 @@ private async Task PurgeShowNetworks(int showId)
 
     public async Task PurgeUnlinkedShowNetworks()
     {
+        // Build the linked-network set once before iterating. Checking inside the loop would
```
revam: No need to explain this, as the code speaks for itself.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -2646,6 +2667,9 @@ Func<T, Task> processAsync
         var tasks = enumerable
             .Select(item => Task.Run(async () =>
             {
+                // The semaphore is acquired inside the Task.Run lambda so all tasks are started
```
revam: We should be changing to an ActionBlock and BufferBlock instead of commenting why creating N tasks which wait on a semaphore is bad. This is something I did not know at the time that code was originally written, and which I have put off fixing as it wasn't a major issue, but the task handing here is sub-optimal for what we want it to do.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -2656,6 +2680,7 @@ Func<T, Task> processAsync
                     semaphore.Release();
                 }
             }))
+            // HashSet gives O(1) removal after Task.WhenAny; a List would scan the whole set each time.
```
revam: No need to explain why we would use a hash set over a list. Any programmer (or LLM) should familiarize themselves with those concepts beforehand if necessary, as part of their learning journey (or training data).

Shoko.Server/Providers/TMDB/TmdbRateLimiter.cs
```diff
@@ -76,6 +76,8 @@ private static SlidingWindowRateLimiter CreateLimiter(int maxRequests, int windo
         {
             PermitLimit = maxRequests,
             Window = TimeSpan.FromMilliseconds(windowMs),
+            // 10 segments divide the window into sub-intervals, smoothing request distribution and
```
revam: Again, as I said in my other review; can we make this into a setting?

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
Foowy: agreed, but not all who peruse changes and look over code is necessarily a programmer, some are just interested in seeing the inner parts(non-mechanic watching mechanic metaphor). That and every group has a specific want/need of comments even on the basic items. 

I will only cover more complex or why i did it this way code changes for comments to avoid over-explaining something

Shoko.Server/Providers/TMDB/TmdbRateLimiter.cs
Foowy: yes, how does SegmentsPerWindow, or something simpler? TMDB_APi_Limits? Leaning toward TMDB_APi_Limits to avoid rbroad name

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
Foowy: This should be able to be done, are the actionblocks already used in filewatcher and videoservice good references to look at?

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
Foowy: explained in another comment, I will only cover more complex or why i did it this way code changes for comments to avoid over-explaining something

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
Foowy: See above

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
Foowy: will redesign it back to original.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
Foowy: See above

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
Foowy: Will toy around with a few things, will commit to the PR for review, placed PR in draft for now

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
Foowy: that and i can add later if anyone feels something needs to be explained better.

Shoko.Server/Providers/TMDB/TmdbRateLimiter.cs
revam: The property name would be scoped to the tmdb rate limit settings, so a relative name to that would be fine. I'll let you decide which, but no namespacing on the property value; so no `settings.TMBD.RateLimit.TMDB_API_Limits`.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
revam: I think they would be fine.

PR #1358, opened 2026-06-12
fix(tmdb): retry person fetch on transient failure, fix 404 routing

Foowy:
### Transient-error resilience and circuit breaker

- **fix**: \`UpdatePerson\` now routes \`NotFoundException\` (HTTP 404) to \`PurgePersonInternal\` — full cross-show/movie cleanup + forced refresh of linked entities — instead of silently returning.
- **fix**: Transient failures (\`HttpRequestException\`, \`RequestLimitExceededException\`) during the person fetch fan-out enqueue an \`UpdateTmdbPersonJob\` retry per person; cast/crew rows for transiently-failed people are preserved (excluded from orphan cleanup) so the retry can link them once the person row is fetched.
- **fix**: \`HasMovieChangedSinceAsync\` and \`GetShowChangedItemsAsync\` now catch transient TMDB errors and fall back to a full refresh rather than aborting the entire job.
- **fix**: \`UpdateTmdbPersonJob\` gains optional \`TmdbShowID\`/\`TmdbMovieID\` fields so \`PurgePersonInternal\` can exclude the triggering entity from re-scheduling, preventing a 404 retry loop.
- **refactor**: Episode update loop replaced with sequential \`foreach\` + \`ActionBlock\` with a cancellable circuit-breaker. \`ProcessWithConcurrencyAsync\` gains an \`onDropped: Action<T>?\` callback so items silently discarded when the circuit breaker fires are handed back to the caller for retry enqueueing.

### Orphan cleanup

- **fix**: Orphaned cast/crew cleanup was only running inside the \`catch\` block, so on the all-transient path (every failure absorbed by the inner handler) cleanup was silently skipped. Moved outside the try/catch so it runs on every fan-out completion.
- **fix**: \`TrySkipUnchangedEpisode\` and \`TryHandleUnchangedSeason\` did not add skipped episodes' existing cast/crew people to \`AllPeopleToAddOrKeep\`. A person appearing only in unchanged episodes would have their \`TMDB_Person\` row purged, leaving dangling cast/crew references. Fixed by querying existing cast/crew rows for skipped episodes/seasons and adding each person to \`AllPeopleToAddOrKeep\` when \`DownloadCrewAndCast\` is true.

### Selective refresh via TMDB Changes API

- **feat**: Unchanged seasons and episodes are now skipped using the TMDB Changes API — avoids redundant \`GetTvSeasonAsync\` / \`GetTvEpisodeAsync\` calls for unmodified content. Falls back to a full refresh when the entity is newly added, \`LastUpdatedAt\` is older than 14 days, or the Changes API returns \`null\`.
- **fix**: \`HasMovieChangedSinceAsync\` and \`GetShowChangedItemsAsync\` passed \`LastUpdatedAt\` (local \`DateTime.Now\`) to TMDbLib, which appends a hardcoded \`" UTC"\` suffix regardless of \`DateTime.Kind\` — wrong on any non-UTC host. Fixed with \`since.ToUniversalTime()\`.

### Missing person stubs

- **fix**: When \`GetTmdbPerson()\` returns null at API serialisation time, \`Role.InitStub\` enqueues a background \`UpdateTmdbPersonJob\` and sets \`IsStub = true\`. All TMDB cast/crew API responses filter stubs via \`.Where(r => !r.IsStub)\` so entries with empty names never reach callers.
- **refactor**: Extracted \`CreateStaffFromTmdbPerson\` helper in \`Role\` to eliminate 8× duplicated person-init blocks across TMDB cast/crew constructors. \`ILogger?\` threaded through all 8 constructors so enqueue failures are logged via the caller's logger.

### Configurable rate limiter

- **feat**: The rate limit is now configurable via \`MaxRequestsPerWindow\` (serialised as \`TMDB_API_Limits\`, range 1–40, default 10).

### Misc

- **fix**: Skipped-count log formula (\`total - added - updated - purged\`) assumed mutual exclusivity of the added/updated counters. Preserved \`else if (updated)\` so a newly-added person increments only \`peopleAdded\`, keeping the formula correct.
- **refactor**: \`ShowSeasonUpdateState\` skip sets changed from \`ConcurrentBag<int>\` to \`ConcurrentDictionary<int,byte>\` for correct deduplication; episode cast/crew pre-loaded as a single \`ILookup<episodeId, personId>\` before the season fan-out instead of issuing per-entity DB queries inside the loop.
- **refactor**: \`TrySkipUnchangedEpisode\` marked \`static\` (no instance access). \`ShowSeasonUpdateState\` constructor reduced from 8 to 6 parameters by grouping \`downloadCrewAndCast\`, \`quickRefresh\`, and \`shouldFireEvents\` into a single \`options\` tuple.

## New Setting: \`TMDB_API_Limits\`

| Field | Value |
|---|---|
| C# property | \`MaxRequestsPerWindow\` |
| JSON key | \`TMDB_API_Limits\` |
| UI label | \`"API Rate Limit"\` |
| Location | TMDB › Rate Limit settings |
| Range | 1–40 requests per window |
| Default | 10 |

### DIFF
--- a/Shoko.Server/API/v3/Controllers/TmdbController.cs
+++ b/Shoko.Server/API/v3/Controllers/TmdbController.cs
@@ -300,8 +300,8 @@ public partial class TmdbController : BaseController
             return NotFound(MovieNotFound);
 
         return movie.Cast
-            .Select(cast => new Role(cast))
-            .Where(r => !r.IsStub)
+            .Select(Role.FromTmdb)
+            .OfType<Role>()
             .ToList();
     }
 
@@ -317,8 +317,8 @@ public partial class TmdbController : BaseController
             return NotFound(MovieNotFound);
 
         return movie.Crew
-            .Select(cast => new Role(cast))
-            .Where(r => !r.IsStub)
+            .Select(Role.FromTmdb)
+            .OfType<Role>()
             .ToList();
     }
 
@@ -1270,8 +1270,8 @@ public partial class TmdbController : BaseController
                     return ValidationProblem("Invalid alternateOrderingID for show.", "alternateOrderingID");
 
                 return alternateOrdering.Cast
-                    .Select(cast => new Role(cast))
-                    .Where(r => !r.IsStub)
+                    .Select(Role.FromTmdb)
+                    .OfType<Role>()
                     .ToList();
             }
 
@@ -1280,8 +1280,8 @@ public partial class TmdbController : BaseController
         }
 
         return show.Cast
-            .Select(cast => new Role(cast))
-            .Where(r => !r.IsStub)
+            .Select(Role.FromTmdb)
+            .OfType<Role>()
             .ToList();
     }
 
@@ -1312,8 +1312,8 @@ public partial class TmdbController : BaseController
                     return ValidationProblem("Invalid alternateOrderingID for show.", "alternateOrderingID");
 
                 return alternateOrdering.Crew
-                    .Select(cast => new Role(cast))
-                    .Where(r => !r.IsStub)
+                    .Select(Role.FromTmdb)
+                    .OfType<Role>()
                     .ToList();
             }
 
@@ -1322,8 +1322,8 @@ public partial class TmdbController : BaseController
         }
 
         return show.Crew
-            .Select(cast => new Role(cast))
-            .Where(r => !r.IsStub)
+            .Select(Role.FromTmdb)
+            .OfType<Role>()
             .ToList();
     }
 
@@ -2076,8 +2076,8 @@ public partial class TmdbController : BaseController
                 return NotFound(SeasonNotFound);
 
             return altOrderSeason.Cast
-                .Select(cast => new Role(cast))
-                .Where(r => !r.IsStub)
+                .Select(Role.FromTmdb)
+                .OfType<Role>()
                 .ToList();
         }
 
@@ -2089,8 +2089,8 @@ public partial class TmdbController : BaseController
             return NotFound(SeasonNotFound);
 
         return season.Cast
-            .Select(cast => new Role(cast))
-            .Where(r => !r.IsStub)
+            .Select(Role.FromTmdb)
+            .OfType<Role>()
             .ToList();
     }
 
@@ -2108,8 +2108,8 @@ public partial class TmdbController : BaseController
                 return NotFound(SeasonNotFound);
 
             return altOrderSeason.Crew
-                .Select(crew => new Role(crew))
-                .Where(r => !r.IsStub)
+                .Select(Role.FromTmdb)
+                .OfType<Role>()
                 .ToList();
         }
 
@@ -2121,8 +2121,8 @@ public partial class TmdbController : BaseController
             return NotFound(SeasonNotFound);
 
         return season.Crew
-            .Select(crew => new Role(crew))
-            .Where(r => !r.IsStub)
+            .Select(Role.FromTmdb)
+            .OfType<Role>()
             .ToList();
     }
 
@@ -2617,8 +2617,8 @@ public partial class TmdbController : BaseController
             return NotFound(EpisodeNotFound);
 
         return episode.Cast
-            .Select(cast => new Role(cast))
-            .Where(r => !r.IsStub)
+            .Select(Role.FromTmdb)
+            .OfType<Role>()
             .ToList();
     }
 
@@ -2634,8 +2634,8 @@ public partial class TmdbController : BaseController
             return NotFound(EpisodeNotFound);
 
         return episode.Crew
-            .Select(cast => new Role(cast))
-            .Where(r => !r.IsStub)
+            .Select(Role.FromTmdb)
+            .OfType<Role>()
             .ToList();
     }
 
--- a/Shoko.Server/API/v3/Controllers/TmdbController.cs
+++ b/Shoko.Server/API/v3/Controllers/TmdbController.cs
@@ -301,7 +301,7 @@ public partial class TmdbController : BaseController
 
         return movie.Cast
             .Select(Role.FromTmdb)
-            .OfType<Role>()
+            .WhereNotNull()
             .ToList();
     }
 
@@ -318,7 +318,7 @@ public partial class TmdbController : BaseController
 
         return movie.Crew
             .Select(Role.FromTmdb)
-            .OfType<Role>()
+            .WhereNotNull()
             .ToList();
     }
 
@@ -1271,7 +1271,7 @@ public partial class TmdbController : BaseController
 
                 return alternateOrdering.Cast
                     .Select(Role.FromTmdb)
-                    .OfType<Role>()
+                    .WhereNotNull()
                     .ToList();
             }
 
@@ -1281,7 +1281,7 @@ public partial class TmdbController : BaseController
 
         return show.Cast
             .Select(Role.FromTmdb)
-            .OfType<Role>()
+            .WhereNotNull()
             .ToList();
     }
 
@@ -1313,7 +1313,7 @@ public partial class TmdbController : BaseController
 
                 return alternateOrdering.Crew
                     .Select(Role.FromTmdb)
-                    .OfType<Role>()
+                    .WhereNotNull()
                     .ToList();
             }
 
@@ -1323,7 +1323,7 @@ public partial class TmdbController : BaseController
 
         return show.Crew
             .Select(Role.FromTmdb)
-            .OfType<Role>()
+            .WhereNotNull()
             .ToList();
     }
 
@@ -2077,7 +2077,7 @@ public partial class TmdbController : BaseController
 
             return altOrderSeason.Cast
                 .Select(Role.FromTmdb)
-                .OfType<Role>()
+                .WhereNotNull()
                 .ToList();
         }
 
@@ -2090,7 +2090,7 @@ public partial class TmdbController : BaseController
 
         return season.Cast
             .Select(Role.FromTmdb)
-            .OfType<Role>()
+            .WhereNotNull()
             .ToList();
     }
 
@@ -2109,7 +2109,7 @@ public partial class TmdbController : BaseController
 
             return altOrderSeason.Crew
                 .Select(Role.FromTmdb)
-                .OfType<Role>()
+                .WhereNotNull()
                 .ToList();
         }
 
@@ -2122,7 +2122,7 @@ public partial class TmdbController : BaseController
 
         return season.Crew
             .Select(Role.FromTmdb)
-            .OfType<Role>()
+            .WhereNotNull()
             .ToList();
     }
 
@@ -2618,7 +2618,7 @@ public partial class TmdbController : BaseController
 
         return episode.Cast
             .Select(Role.FromTmdb)
-            .OfType<Role>()
+            .WhereNotNull()
             .ToList();
     }
 
@@ -2635,7 +2635,7 @@ public partial class TmdbController : BaseController
 
         return episode.Crew
             .Select(Role.FromTmdb)
-            .OfType<Role>()
+            .WhereNotNull()
             .ToList();
     }
 
--- a/Shoko.Server/API/v3/Controllers/TmdbController.cs
+++ b/Shoko.Server/API/v3/Controllers/TmdbController.cs
@@ -301,6 +301,7 @@ public partial class TmdbController : BaseController
 
         return movie.Cast
             .Select(cast => new Role(cast))
+            .Where(r => !r.IsStub)
             .ToList();
     }
 
@@ -317,6 +318,7 @@ public partial class TmdbController : BaseController
 
         return movie.Crew
             .Select(cast => new Role(cast))
+            .Where(r => !r.IsStub)
             .ToList();
     }
 
@@ -1269,6 +1271,7 @@ public partial class TmdbController : BaseController
 
                 return alternateOrdering.Cast
                     .Select(cast => new Role(cast))
+                    .Where(r => !r.IsStub)
                     .ToList();
             }
 
@@ -1278,6 +1281,7 @@ public partial class TmdbController : BaseController
 
         return show.Cast
             .Select(cast => new Role(cast))
+            .Where(r => !r.IsStub)
             .ToList();
     }
 
@@ -1309,6 +1313,7 @@ public partial class TmdbController : BaseController
 
                 return alternateOrdering.Crew
                     .Select(cast => new Role(cast))
+                    .Where(r => !r.IsStub)
                     .ToList();
             }
 
@@ -1318,6 +1323,7 @@ public partial class TmdbController : BaseController
 
         return show.Crew
             .Select(cast => new Role(cast))
+            .Where(r => !r.IsStub)
             .ToList();
     }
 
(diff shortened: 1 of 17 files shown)


Foowy:
TMDB_API_Limits, I would like input into what is preferred by everyone. I have not referenced the webui code to see how this would be presented. Just exposed a point. If anything about the setting needs to be changed, i am open arms. Including nuking it from orbit.

Shoko.Server/API/v3/Models/Common/Role.cs
```diff
@@ -1,312 +1,262 @@
...
-using Microsoft.Extensions.DependencyInjection;
-using Newtonsoft.Json;
-using Shoko.Abstractions.Core.Services;
-using Shoko.Abstractions.Metadata;
-using Shoko.Server.API.v3.Helpers;
-using Shoko.Server.Models.AniDB;
-using Shoko.Server.Models.TMDB;
-using Shoko.Server.Providers.TMDB;
-using Shoko.Server.Server;
+using System;
```
revam: Check which line endings the file is using, and convert them back to their original line endings.

Shoko.Server/API/v3/Models/Common/Role.cs
```diff
@@ -1,312 +1,262 @@
...
+            Type = staff.Type.ToString(),
+        };
+        RoleName = xref.RoleType;
+        RoleDetails = xref.Role;
+    }
+
+    public Role(TMDB_Movie_Cast cast, ILogger? logger = null)
+    {
+        var person = cast.GetTmdbPerson();
+        if (person is null) { InitStub(cast.TmdbPersonID, CreatorRoleType.Actor, CharacterRole, cast.CharacterName, logger); return; }
```
revam: …If you're going to do multiple statements within the `if` scope… then break the statements into multiple lines and do not inline them like this.

Shoko.Server/API/v3/Models/Common/Role.cs
```diff
@@ -1,312 +1,262 @@
...
+        };
+    }
+
+    [MemberNotNull(nameof(Staff))]
+    private void InitStub(int tmdbPersonId, CreatorRoleType roleType, string roleDetails, string? characterName = null, ILogger? logger = null)
+    {
+        try
+        {
+            var scheduler = ISystemService.StaticServices.GetRequiredService<IQueueScheduler>();
+            _ = scheduler.Enqueue<UpdateTmdbPersonJob>(j => j.TmdbPersonID = tmdbPersonId);
```
revam: Enqueue won't run it then and there, so we might as well not run it in the first place. It's already too late for the client which asked for the data, and said client probably won't ask again for the data unless a forced client-side cache invalidation is ran, which is not often. The old code ran it then and there, and we do have the ability to run jobs outside the queue, so either replaced the enqueue with running it now, outside the queue, or we remove it all and just return the stub, which, as i said in my other comment, there isn't any reason to set the details for since we're just going to ignore the stub anyways.

Shoko.Server/API/v3/Models/TMDB/TmdbShow.cs
```diff
@@ -255,10 +255,12 @@ public TmdbShow(TMDB_Show show, TMDB_AlternateOrdering? alternateOrdering, Inclu
         if (include.HasFlag(IncludeDetails.Cast))
             Cast = (alternateOrdering is null ? show.Cast : alternateOrdering.Cast)
                 .Select(cast => new Role(cast))
+                .Where(r => !r.IsStub)
```
revam: What's the point of setting ANY details on the stub if we're just going to ignore it?

Shoko.Server/Settings/TmdbRateLimitSettings.cs
```diff
@@ -11,13 +12,14 @@ public class TmdbRateLimitSettings
...
     /// <summary>
     /// Maximum number of requests allowed within the rate limit window.
-    /// TMDB enforces approximately 40 requests per second; this defaults to 10 to avoid overwhelming
+    /// TMDB enforces a maximum of 40 requests per second; this defaults to 10 to avoid overwhelming
     /// end-user hardware with API requests and the data processing that follows each one.
     ///</summary>
     [Badge("Debug", Theme = DisplayColorTheme.Warning)]
     [Visibility(Size = DisplayElementSize.Small, Advanced = true)]
-    [Display(Name = "Max Requests Per Window")]
+    [Display(Name = "API Rate Limit")]
```
revam: Max Requests Per Window sounds better. No need to change it to this.

If we compare "TMDB → Rate Limit → Max Requests Per Window" to "TMDB → Rate Limit → API Rate Limit", then I think the former sounds more descriptive from it's name from what it does, and it also doesn't just repeat 'rate limit' twice.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -2634,56 +2865,36 @@ private bool IsPersonLinkedToOtherEntities(int tmdbPersonId)
...
     {
         if (maxConcurrent < 1)
             throw new ArgumentOutOfRangeException(nameof(maxConcurrent), "Concurrency level must be at least 1.");
 
-        var semaphore = new SemaphoreSlim(maxConcurrent);
-        var exceptions = new List<Exception>();
-        var cancellationTokenSource = new CancellationTokenSource();
-        var tasks = enumerable
-            .Select(item => Task.Run(async () =>
+        var exceptions = new ConcurrentBag<Exception>();
```
revam, line 2857: Good work on the action block.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -2771,5 +2982,99 @@ private bool IsEntityLocked(DataEntityType entityType, int id, string metadataKe
     private Task<IDisposable> GetLockForEntity(DataEntityType entityType, int id, string metadataKey, string reason)
         => _entityLock.GetLockForEntityAsync(entityType, id, metadataKey, reason);
 
+    /// <summary>
+    /// Returns <see langword="true"/> if TMDB has recorded any changes for the given movie since
+    /// <paramref name="since"/>, or if the TMDB Changes API 14-day window has been exceeded (in
+    /// which case a full refresh is required because the history is no longer available).
+    /// </summary>
+    private async Task<bool> HasMovieChangedSinceAsync(int movieId, DateTime since)
```
revam, line 2975: These methods do not belong to 'Locking'. Put them in their own region _below_ locking.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -669,21 +701,50 @@ private async Task<bool> UpdateMovieCastAndCrew(TMDB_Movie tmdbMovie, MovieCredi
...
+                .Where(c => missingPersonIds.Contains(c.TmdbPersonID)).ToList();
+            if (orphanedCast.Count > 0 || orphanedCrew.Count > 0)
+            {
+                _logger.LogWarning("TMDB: Removed {CastCount} cast and {CrewCount} crew entries for {PersonCount} people that failed to fetch. (Movie={MovieId})",
+                    orphanedCast.Count, orphanedCrew.Count, missingPersonIds.Count, tmdbMovie.Id);
+                _tmdbMovieCast.Delete(orphanedCast);
+                _tmdbMovieCrew.Delete(orphanedCrew);
+            }
+        });
+        await Task.WhenAll(transientlyFailedMovieSet.Select(personId =>
```
revam: Why are we removing their cast/crew roles and then afterwards enqueuing a job to update the local person data for the TMDB people AFTER we removed their cast/crew roles? What am I missing here?

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -1091,6 +1152,22 @@ public async Task<bool> UpdateShow(TmdbShowUpdateOptions options)
             if (show is null)
                 return false;
 
+            // Null result (14-day window exceeded or API failure) triggers a full refresh in UpdateShowSeasonsAndEpisodes.
+            // Not applied on forced refreshes or newly-added shows — those always do a full fetch.
+            (HashSet<int> SeasonNumbers, HashSet<(int Season, int Episode)> Episodes)? changedItems = null;
+            if (!newlyAdded && !forceRefresh)
```
revam: Possibly also need to disable this for quick-refresh; as that's intended as a full-but-lightweight fetch of all the essential information needed for episode linking in the Web UI. IDR which checks we had for quick-refresh already, but please check if it's needed here.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -1162,242 +1239,378 @@ public async Task<bool> UpdateShow(TmdbShowUpdateOptions options)
         }
     }
 
-    private async Task<(bool episodesOrSeasonsUpdated, Dictionary<TMDB_Season, UpdateReason> updatedSeasons, Dictionary<TMDB_Episode, UpdateReason> updatedEpisodes, int episodeCount, int hiddenEpisodeCount)> UpdateShowSeasonsAndEpisodes(TvShow show, bool downloadCrewAndCast = false, bool forceRefresh = false, bool downloadImages = false, bool quickRefresh = false, bool shouldFireEvents = false)
+    private sealed class ShowSeasonUpdateState
```
revam: I must say, splitting this into multiple methods with a shared state made it, at least to me, harder to keep track of what goes where and figuring out the intent behind the code compared to if everything was in-line. I'm not saying it was a bad move, but it did make it harder for me personally to get a mental model of what it does.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -1162,242 +1239,378 @@ public async Task<bool> UpdateShow(TmdbShowUpdateOptions options)
...
+        public readonly ConcurrentBag<TMDB_Season> SeasonsToSave = new();
+        public readonly ConcurrentDictionary<TMDB_Season, UpdateReason> SeasonEventsToEmit = new();
+        public readonly ConcurrentDictionary<int, byte> EpisodesToSkip = new();
+        public readonly ConcurrentBag<TMDB_Episode> EpisodesToSave = new();
+        public readonly ConcurrentDictionary<TMDB_Episode, UpdateReason> EpisodeEventsToEmit = new();
+        public readonly ConcurrentDictionary<int, byte> AllPeopleToAddOrKeep = new();
+        public readonly ConcurrentDictionary<int, byte> AllPeopleToPotentiallyRemove = new();
+
+        public ShowSeasonUpdateState(
+            (bool DownloadCrewAndCast, bool QuickRefresh, bool ShouldFireEvents) options,
```
revam: Please don't resort to tuples to "reduce" the number of arguments you take. Also, instead of doing a constructor, you can mark the fields above as `required` to have to pass them into the object initiator when creating a new object. E.g. `new() { QuickRefresh = quickRefresh }`, etc..

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -1162,242 +1239,378 @@ public async Task<bool> UpdateShow(TmdbShowUpdateOptions options)
...
-            Interlocked.Add(ref totalEpisodeCount, episodeCount);
-            Interlocked.Add(ref totalHiddenEpisodeCount, hiddenEpisodeCount);
-        });
+        var episodePeople = downloadCrewAndCast
+            ? _tmdbEpisodeCast.GetByTmdbShowID(show.Id).Select(c => (c.TmdbEpisodeID, c.TmdbPersonID))
+                .Concat(_tmdbEpisodeCrew.GetByTmdbShowID(show.Id).Select(c => (c.TmdbEpisodeID, c.TmdbPersonID)))
+                .ToLookup(x => x.TmdbEpisodeID, x => x.TmdbPersonID)
+            : Enumerable.Empty<(int, int)>().ToLookup(x => x.Item1, x => x.Item2);
+        var state = new ShowSeasonUpdateState((downloadCrewAndCast, quickRefresh, shouldFireEvents), changedItems, (preferredTitleLanguages, preferredOverviewLanguages), existingSeasons, existingEpisodes, episodePeople);
+        await ProcessWithConcurrencyAsync(_maxConcurrency, show.Seasons!, reducedSeason =>
```
revam: Didn't we agree (on discord) to remove the concurrency and async-ness on the season loop? Why is it back to `_maxConcurreny` times `_maxConcurrency` (one for seasons, one for episodes) again?

revam (changes requested):
A few things.

Foowy:
Apparently i have an inherent ability to break line endings

da3dsoul:
> Apparently i have an inherent ability to break line endings

It happens

Foowy:
Another thing to add to the list of things i am good at, angering SonarCloud.

Sucks at Math = True
Line Endings = False
Sonar Cloud Gods Appeased = False
Grammar and Spelling = Criminal, the police have been notified

Also remember, 2 try statements in a row just means try really hard

Shoko.Server/API/v3/Controllers/TmdbController.cs
```diff
@@ -300,7 +300,8 @@ [FromRoute] int movieID
             return NotFound(MovieNotFound);
 
         return movie.Cast
-            .Select(cast => new Role(cast))
+            .Select(Role.FromTmdb)
+            .OfType<Role>()
```
revam: Replace `.OfType<Role>()` → `.WhereNotNull()` for all occurences. Import `Shoko.Abstractions.Exensions` if necessary.

Shoko.Server/API/v3/Models/Common/Role.cs
```diff
@@ -1,312 +1,181 @@
...
-using Microsoft.Extensions.DependencyInjection;
-using Newtonsoft.Json;
-using Shoko.Abstractions.Core.Services;
-using Shoko.Abstractions.Metadata;
-using Shoko.Server.API.v3.Helpers;
-using Shoko.Server.Models.AniDB;
-using Shoko.Server.Models.TMDB;
-using Shoko.Server.Providers.TMDB;
-using Shoko.Server.Server;
+using System.ComponentModel.DataAnnotations;
```
revam: There are still line ending issues in this file.

Shoko.Server/API/v3/Models/Common/Role.cs
```diff
@@ -1,312 +1,181 @@
...
+    private static Person CreateStaffFromTmdbPerson(TMDB_Person person)
+    {
+        var images = person.GetImages();
+        return new()
+        {
+            ID = person.Id,
+            Name = person.EnglishName,
+            AlternateName = person.Aliases.Count == 0 ? person.EnglishName : person.Aliases[0].Split("/").Last().Trim(),
+            Description = person.EnglishBiography,
+            Image = images.Count > 0 ? new Image(images[0]) : null,
```
revam: Apply the below, and remove the `var images` above.
```suggestion
            Image = (person as ICreator).PrimaryImage is { } staffImage ? new Image(staffImage) : null,
```

revam (changes requested):
Logic seems fine, and will get some testing in the wild later. No need to change it further until merged. There are still a few nits to fix though before said merge.

Foowy:
Going to try figuring out why role.cs is the only file that keeps getting mangled LE's

revam (approved):
Time for wider testing. As for Role.cs, probably had mixed endings. Can be fixed in a separate PR/commit.

PR #1362, opened 2026-06-20
fix(tmdb): require exact title match for subtitle-stripped show searches

Foowy:
## Problem

`AutoSearchForShowUsingTitle` uses up to 6 TMDB search attempts, progressively stripping sequel suffixes and subtitles from the query to find the root show. However, attempts #2, #3, #5, and #6 accepted **any** animation result even when the top-ranked TMDB hit was a spinoff that happened to share a year and title prefix with the intended series.

Concrete example: searching for `転生したらスライムだった件` with year 2021 caused TMDB to return **The Slime Diaries** as the top result — it aired in 2021 and its full Japanese title contains the Tensura title as a prefix — even though the correct match is the root Tensura show.

### 1. Title match guard (attempts #2, #3, #5, #6)

These four attempts strip sequel suffixes or subtitles specifically to search for the **root** show. A guard (`IsTitleMatch`) now requires that the TMDB result's `OriginalName` or `Name` exactly matches the stripped query before it is accepted.

Attempts #1 (full title + year) and #4 (full title, no year) are intentional broad fallbacks and are left unguarded.

The Slime Diaries is unaffected: its own full Japanese title (`転生したらスライムだった件 転スラ日記`) is looked up via attempt #1, which has no guard.

### 2. Punctuation normalization

AniDB and TMDB sometimes encode the same title with different Unicode punctuation (`!` vs `！`, ` ` vs `〜`). `NormalizeTitle` applies NFKC to convert full-width ASCII to half-width, then folds wave dash `〜` (U+301C — not covered by NFKC) to a space. Both the query and the candidate title are normalized before comparison.

### 3. Mixed-script title support

Some AniDB titles combine Latin and CJK in one string (e.g. `ONE PUNCH MAN ワンパンマン`). After stripping a sequel suffix the combined string matches neither TMDB's Japanese `OriginalName` nor its English `Name`. `IsTitleMatch` now extracts CJK and Latin word groups independently and tries each against both TMDB fields, so `ワンパンマン` can match a Japanese `OriginalName` and `ONE PUNCH MAN` can match an English `Name`.

## Testing

Verified on a local dev build:
- `転生したらスライムだった件 (2021)` → correctly resolves to Tensura (not The Slime Diaries)
- `転スラ日記` (The Slime Diaries itself) → still resolves correctly via attempt #1
- Existing xrefs (shows already linked to TMDB) bypass the search entirely — no regressions expected for those

### DIFF
--- a/Shoko.Server/Providers/TMDB/TmdbSearchService.cs
+++ b/Shoko.Server/Providers/TMDB/TmdbSearchService.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
+using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
@@ -461,12 +462,15 @@ public partial class TmdbSearchService : ITmdbSearchService
         }
 
         // Brute force attempt #2: Same as above, but after stripping the title of common "sequel endings"
+        // Require an exact title match for the same reason as attempts #3/#5/#6: we stripped a suffix to
+        // derive the root-series name, so a TMDB result whose own title contains further content (e.g. a
+        // spinoff that has the root title as a prefix) is the wrong show.
         var strippedTitle = SequelSuffixRemovalRegex().Match(originalTitle) is { Success: true } regexResult
             ? originalTitle[..^regexResult.Length].TrimEnd() : null;
         if (!string.IsNullOrEmpty(strippedTitle))
         {
             (results, totalFound) = await SearchShowsRaw(strippedTitle, includeRestricted: restricted, year: airDate.Year).ConfigureAwait(false);
-            firstViableResult = results.FirstOrDefault(result => IsAnimation(result.GetGenres()));
+            firstViableResult = results.FirstOrDefault(result => IsAnimation(result.GetGenres()) && IsTitleMatch(strippedTitle, result.OriginalName, result.Name));
             if (firstViableResult is not null)
             {
                 _logger.LogTrace("Found {Count} show results for search on {Query}, best match; {ShowName} ({ID})", totalFound, strippedTitle, firstViableResult.OriginalName, firstViableResult.Id);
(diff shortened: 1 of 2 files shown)


da3dsoul:
I would use the new fuzzy search service as it's more forgiving

Foowy:
> I would use the new fuzzy search service as it's more forgiving

Will do

Foowy:
I think there may be a semantic mismatch worth clarifying first.

FuzzySearchService.FuzzyMatchesAnyName uses substring matching as its primary check. For the guarded attempts (#2/#3/#5/#6), that would reintroduce the original bug: the spinoff title `転生したらスライムだった件 転スラ日記` contains the root query `転生したらスライムだった件` as a substring, so the spinoff would be accepted again.

The guard specifically needs **exact normalized equality** — the point is that a result whose own title has content beyond the stripped query is a spinoff, not the root show.

One clean path forward: add an `ExactMatchesAnyName(string query, IReadOnlySet<string> names)` method to `IFuzzySearchService` that uses `NormalizeForIndex` for normalization but does equality comparison instead of substring/fuzzy. That would consolidate normalization into the service and remove the custom `NormalizeTitle` from `TmdbSearchService`. `NormalizeForIndex` would also need a `〜` (U+301C wave dash) → space step, since NFKC does not convert it (unlike U+FF5E fullwidth tilde, which NFKC maps to `~`).

```
// on IFuzzySearchService / FuzzySearchService
public bool ExactMatchesAnyName(string query, IReadOnlySet<string> names)
    => names.Any(name => string.Equals(
        NormalizeForIndex(query),   // updated to handle 〜
        NormalizeForIndex(name),
        StringComparison.OrdinalIgnoreCase));
```

Is that the direction you had in mind, or were you thinking of a different approach?

revam:
also check one punch man, and owarimonogatari, as those two tend to match to totally unrelated shows. also check the shows they match to that they're matching fine.

Foowy:
I can go forward with attempting to cutover to fuzzy search and see about fixing individual language character gochas in edge, or treat it as a refactor later? 

Current changes on this PR correct for One Punch Man and Owarimonogatari.

On the fence with go forward, while this corrects the now issue, using Search Service is more uniform across codebase

da3dsoul:
Fair point. I would make a fuzzy version that is more configurable on the acceptable distance/mistake threshold, and make it skip the substring optimizations.

da3dsoul:
You can also tell the search to try to use the shortest matching title, ignoring short names

revam:
we don't control the search, we only act on the search results returned by TMDb's API.

also  if we match the X number of first matches on an episode level and pick the one with best episode matches then we don't need to workaround their search API shenanigans as we'll be checking which matches best. I pinged @Foowy with this idea on discord the other day if they wanted to implement that, but no response so dunno if they've seen it yet or not.

da3dsoul:
I've never suggested that because it involves pulling a lot of data. It might be worth it for auto-matching, as that happens in the background. It definitely shouldn't happen in front of the user, though, whether for sorting, suggesting, etc

revam:
> I've never suggested that because it involves pulling a lot of data. It might be worth it for auto-matching, as that happens in the background. It definitely shouldn't happen in front of the user, though, whether for sorting, suggesting, etc

this whole search is for the auto matching though

Issue #1285, opened 2025-11-03
One Punch Man (2025) (Season 3) gets matched with "random" TMDB Entry


Rand-Random:
### Before submitting your issue, please verify the following.

- [x] You're using the latest **Stable** or **Daily** version of Shoko Server.
- [x] You're using the latest and correct (Stable:Stable, Daily:Daily) Web UI for the version of Shoko you're using.
- [x] You've searched this repository (top of the page) for the issue you are reporting.

### Shoko Version

5.1.0.142

### Shoko Version - Daily Commit

3a1e5b5

### Install Type

Windows

### Description

One Punch Man (2025) (Season 3) is linking to two animes:

<img width="719" height="420" alt="Image" src="https://github.com/user-attachments/assets/f4d92916-4fb5-494d-a2d6-a6b63c6da1ec" />

https://www.themoviedb.org/tv/271605
https://www.themoviedb.org/tv/63926

and 271605 seems rather random pick
https://www.themoviedb.org/tv/271605

<img width="968" height="669" alt="Image" src="https://github.com/user-attachments/assets/92c9c229-8154-46b2-aba4-1fdfdd25bea4" />

I cant remove the entry for 271605 because when I do that it loses the episode links to the correct tmdb entry.

### Steps to Reproduce

Have One Punch Man (2025) (Season 3) in your library and use "Auto-Search TMDB Match" function.

### Log File

```
[2025-11-03 01:21:58:124] Info|Shoko.Server.Providers.TMDB.TmdbLinkingService > Removing TMDB show link: AniDB anime (AnimeID=17576) → TMDB show (ID=63926)
[2025-11-03 01:21:58:124] Info|Shoko.Server.Providers.TMDB.TmdbLinkingService > Removing 0 episodes cross-references for AniDB anime (AnimeID=17576) and TMDB show (ID=63926)
[2025-11-03 01:21:58:124] Info|Shoko.Server.Services.AnimeGroupService > Starting Updating STATS for GROUP One Punch Man from Top Level (recursively) - Watched Stats: False, Missing Episodes: False
[2025-11-03 01:21:58:132] Info|Shoko.Server.Services.AnimeGroupService > Starting Updating STATS for GROUP One Punch Man - Watched Stats: False, Missing Episodes: False
[2025-11-03 01:22:14:489] Info|SearchTmdbJob_AnimeID:17576 > Processing SearchTmdbJob for One Punch Man (2025): AniDB ID 17576
[...]

PR #1363, opened 2026-06-20
feat(tmdb,search): collect top-N candidates and score by title, date, and season

Foowy:
## Summary

Replaces the first-match strategy in \`AutoSearchForShowUsingTitle\` and \`AutoSearchMovieUsingTitle\` with a top-N candidate pool scored by title quality and date match before a winner is chosen. \`TMDBSettings.AutoSearchShowCandidateCount\` and \`TMDBSettings.AutoSearchMovieCandidateCount\` control pool size independently (default 5, range 1–10).

**Fixes:**
- **Spinoff/season mismatches** (e.g. TenSura S2 linking to The Slime Diaries): year-free searches always run so the root show is always a candidate even when its \`first_air_date\` predates the season year.
- **#1285** (One Punch Man S3 co-matching an unrelated show): scoring prefers the entry that reaches \`TitleMatches\` over one that only scores \`FirstAvailable\`.
- **Subtitle-stripped parent show mismatches** (e.g. \"Fairy Tail: 100 Years Quest\" matching base Fairy Tail): \`titleWithoutSubTitle\` (colon-split) is fuzzy-only and can score at most \`TitleKindaMatches\`, so the specific entry that exact-matches the full title wins.
- **Prequel-traversal double match** (e.g. 100 Years Quest returning both s248947 and s46261): when the prequel chain is followed to the root, the current anime's own original title is tried first, then the main title; if either scores better than the root-title result, the root series' stored xrefs are suppressed via a \`prequelFollowed\` flag.
- **Local xref path losing \`MatchRating\`**: existing cross-references now preserve their rating.

**Other changes:**
- \`SequelSuffixRemovalRegex\` extended to strip \`第X期\` alongside \`第X季\`, \`2nd Season\`, \`(YYYY)\`.
- Wave-dash U+301C \`〜\` added to \`SeriesSearch.NormalizeForIndex\` punctuation-to-space map.
- 15 unit tests covering \`isNotExact\` gate, \`NormalizeForIndex\` edge cases, and subtitle colon-split behaviour.

## Scoring

```
exactTitle  + dateMatch  → DateAndTitleMatches
exactTitle  only         → TitleMatches
fuzzyTitle  + dateMatch  → DateAndTitleKindaMatches
date only               → DateMatches
fuzzyTitle  only         → TitleKindaMatches
none                     → FirstAvailable
```

\`originalTitle\` and \`strippedTitle\` are exact-eligible; \`titleWithoutSubTitle\` is fuzzy-only. Shows use season-level air year + episode-count diff as tiebreaker. Movies fetch all regional release dates (\`MovieMethods.ReleaseDates\`) and match on any region's year rather than only the primary \`ReleaseDate\`.

Priority order (highest to lowest): \`DateAndTitleMatches\` → \`TitleMatches\` → \`DateAndTitleKindaMatches\` → \`DateMatches\` → \`TitleKindaMatches\` → \`FirstAvailable\`. \`MatchRating\` has a numeric ordering inversion so an explicit \`ShowMatchPriority\` helper is used instead of sorting by raw enum value. The private \`TitleMatchKind\` enum has been removed; \`ScoreQueryVariants\` now returns \`MatchRating\` directly (\`None\` / \`TitleKindaMatches\` / \`TitleMatches\`).

Both scoring loops break early when a \`DateAndTitleMatches\` result is found — no remaining candidate can outscore it, so further show and season fetches are skipped.

## Episode-level date check (show scoring)

When the year-level date check fails for a show candidate — i.e. neither the best season's air year nor the show's \`first_air_date\` year matches the AniDB air date — the scorer fetches the best-matching season's first episode from TMDB and compares its air date against AniDB episode 1's specific air date (±3 days). A close match promotes \`dateMatch\` to \`true\` before the rating switch.

This catches premieres that straddle a year boundary or split-cour shows where the season-level year metadata would miss the correct entry. The check is skipped entirely when AniDB has no episode 1 date. \`GetTvSeasonAsync\` is called with no extra methods; episode air dates are included in the base season response. Episodes are ordered by \`EpisodeNumber\` before taking the first so TMDB data ordering differences cannot produce the wrong episode.

## Known limitations

- **Exclamation-mark title collision** (\`K-On!\` vs \`K-On!!\`, \`Working!!\` vs \`Working!!!\`): \`!\` is in the punctuation-to-space map, so both forms normalize identically and score the same \`TitleMatches\` rating. Disambiguation falls through to the tiebreaker — episode count diff for shows, search-result order for movies. In practice the correct entry has the right episode count or ranks higher in TMDB's relevance order, so this resolves correctly but is not guaranteed for edge cases.
- **Single-character title queries** (e.g. \`C\` from \`[C] - The Money of Soul and Possibility Control\`): \`GetMaxErrors(1) = 0\` so the edit-distance path never fires for 1-character strings. The scorer falls through to prefix/substring matching only; the correct TMDB entry is still found via \`PrefixMatchesAnyName\` if the short title appears as a prefix, but any edit-distance fuzzy match is unavailable.
- **Year-free pool cap**: the year-free candidate pool is capped at \`2 × candidateCount\`. At the minimum setting of 1, this allows at most 2 year-free candidates. Sufficient for most titles; increase \`AutoSearchShowCandidateCount\` / \`AutoSearchMovieCandidateCount\` if edge cases arise.

### DIFF
--- a/Shoko.Server/Providers/TMDB/TmdbSearchService.cs
+++ b/Shoko.Server/Providers/TMDB/TmdbSearchService.cs
@@ -301,6 +301,7 @@ public partial class TmdbSearchService : ITmdbSearchService
                 (true, _, true)  => MatchRating.DateAndTitleMatches,
                 (true, _, false) => MatchRating.TitleMatches,
                 (_, true, true)  => MatchRating.DateAndTitleKindaMatches,
+                (_, _, true)     => MatchRating.DateMatches,
                 (_, true, false) => MatchRating.TitleKindaMatches,
                 _                => MatchRating.FirstAvailable,
             };
@@ -310,6 +311,10 @@ public partial class TmdbSearchService : ITmdbSearchService
                 "Candidate movie {MovieName} ({ID}): rating={Rating}, releaseYear={ReleaseYear}",
                 full.Title, full.Id, rating, full.ReleaseDate?.Year
             );
+
+            // Maximum confidence — no other candidate can outscore this, so skip remaining fetches.
+            if (rating is MatchRating.DateAndTitleMatches)
+                break;
         }
 
         // If all GetMovieAsync calls returned null (transient outage), fall back to the first candidate
@@ -677,6 +682,7 @@ public partial class TmdbSearchService : ITmdbSearchService
                 (true, _, true)  => MatchRating.DateAndTitleMatches,
                 (true, _, false) => MatchRating.TitleMatches,
                 (_, true, true)  => MatchRating.DateAndTitleKindaMatches,
+                (_, _, true)     => MatchRating.DateMatches,
                 (_, true, false) => MatchRating.TitleKindaMatches,
                 _                => MatchRating.FirstAvailable,
             };
@@ -686,6 +692,10 @@ public partial class TmdbSearchService : ITmdbSearchService
                 "Candidate show {ShowName} ({ID}): rating={Rating}, bestSeasonEpisodeDiff={EpisodeDiff}, seasonYearMatch={SeasonYearMatch}",
                 full.Name, full.Id, rating, bestSeasonEpisodeDiff, seasonYearMatch
             );
+
+            // Maximum confidence — no other candidate can outscore this, so skip remaining fetches.
+            if (rating is MatchRating.DateAndTitleMatches)
+                break;
         }
 
         // If all GetTvShowAsync calls returned null (transient outage), fall back to the first candidate
--- a/Shoko.Server/Providers/TMDB/TmdbSearchService.cs
+++ b/Shoko.Server/Providers/TMDB/TmdbSearchService.cs
@@ -6,14 +6,18 @@ using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Shoko.Abstractions.Extensions;
+using Shoko.Abstractions.Filtering.Services;
 using Shoko.Abstractions.Metadata;
 using Shoko.Abstractions.Metadata.Anidb;
 using Shoko.Abstractions.Metadata.Enums;
 using Shoko.Abstractions.Metadata.Tmdb;
 using Shoko.Abstractions.Metadata.Tmdb.Services;
 using Shoko.Server.Models.AniDB;
+using Shoko.Server.Settings;
+using Shoko.Server.Utilities;
 using TMDbLib.Objects.General;
 using TMDbLib.Objects.Search;
+using TMDbLib.Objects.TvShows;
 
 #nullable enable
 namespace Shoko.Server.Providers.TMDB;
@@ -24,22 +28,28 @@ public partial class TmdbSearchService : ITmdbSearchService
     /// <summary>
     /// This regex might save the day if the local database doesn't contain any prequel metadata, but the title itself contains a suffix that indicates it's a sequel of sorts.
     /// </summary>
-    [GeneratedRegex(@"\(\d{4}\)$|\bs(?:eason)? (?:\d+|(?=[MDCLXVI])M*(?:C[MD]|D?C{0,3})(X[CL]|L?X{0,3})(I[XV]|V?I{0,3}))$|\bs\d+$|第(零〇一二三四五六七八九十百千萬億兆京垓點)+季$|\b(?:second|2nd|third|3rd|fourth|4th|fifth|5th|sixth|6th) season$", RegexOptions.IgnoreCase | RegexOptions.Compiled | RegexOptions.Multiline)]
+    [GeneratedRegex(@"\(\d{4}\)$|\bs(?:eason)? (?:\d+|(?=[MDCLXVI])M*(?:C[MD]|D?C{0,3})(X[CL]|L?X{0,3})(I[XV]|V?I{0,3}))$|\bs\d+$|第(零〇一二三四五六七八九十百千萬億兆京垓點)+[季期]$|\b(?:second|2nd|third|3rd|fourth|4th|fifth|5th|sixth|6th) season$", RegexOptions.IgnoreCase | RegexOptions.Compiled | RegexOptions.Multiline)]
     private partial Regex SequelSuffixRemovalRegex();
 
     private readonly ILogger<TmdbSearchService> _logger;
 
     private readonly TmdbMetadataService _tmdbService;
 
+    private readonly IFuzzySearchService _fuzzySearch;
+
+    private readonly ISettingsProvider _settingsProvider;
+
     /// <summary>
     /// Max days into the future to search for matches against.
     /// </summary>
     private readonly TimeSpan _maxDaysIntoTheFuture = TimeSpan.FromDays(15);
 
-    public TmdbSearchService(ILogger<TmdbSearchService> logger, TmdbMetadataService tmdbService)
+    public TmdbSearchService(ILogger<TmdbSearchService> logger, TmdbMetadataService tmdbService, IFuzzySearchService fuzzySearch, ISettingsProvider settingsProvider)
     {
         _logger = logger;
         _tmdbService = tmdbService;
+        _fuzzySearch = fuzzySearch;
+        _settingsProvider = settingsProvider;
     }
 
     async Task<IReadOnlyList<ITmdbAutoSearchResult>> ITmdbSearchService.SearchForAutoMatch(IAnidbAnime anime)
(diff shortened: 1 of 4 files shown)


Foowy:
Brought everything back into TmdbSearchService.cs to avoid polluting fuzzysearch and ifuzzysearch.

Shoko.Server/Providers/TMDB/TmdbSearchService.cs
```diff
@@ -24,22 +28,30 @@ public partial class TmdbSearchService : ITmdbSearchService
     /// <summary>
     /// This regex might save the day if the local database doesn't contain any prequel metadata, but the title itself contains a suffix that indicates it's a sequel of sorts.
     /// </summary>
-    [GeneratedRegex(@"\(\d{4}\)$|\bs(?:eason)? (?:\d+|(?=[MDCLXVI])M*(?:C[MD]|D?C{0,3})(X[CL]|L?X{0,3})(I[XV]|V?I{0,3}))$|\bs\d+$|第(零〇一二三四五六七八九十百千萬億兆京垓點)+季$|\b(?:second|2nd|third|3rd|fourth|4th|fifth|5th|sixth|6th) season$", RegexOptions.IgnoreCase | RegexOptions.Compiled | RegexOptions.Multiline)]
+    [GeneratedRegex(@"\(\d{4}\)$|\bs(?:eason)? (?:\d+|(?=[MDCLXVI])M*(?:C[MD]|D?C{0,3})(X[CL]|L?X{0,3})(I[XV]|V?I{0,3}))$|\bs\d+$|第(零〇一二三四五六七八九十百千萬億兆京垓點)+[季期]$|\b(?:second|2nd|third|3rd|fourth|4th|fifth|5th|sixth|6th) season$", RegexOptions.IgnoreCase | RegexOptions.Compiled | RegexOptions.Multiline)]
     private partial Regex SequelSuffixRemovalRegex();
 
+    private enum TitleMatchKind { None, KindaMatch, ExactMatch }
```
revam: Any reason why we can't re-use the match rating enum?

Shoko.Server/Providers/TMDB/TmdbSearchService.cs
```diff
@@ -225,50 +237,95 @@ private async Task<bool> AutoSearchForMovie(List<TmdbAutoSearchResult> list, Ani
 
     private async Task<TmdbAutoSearchResult?> AutoSearchMovieUsingTitle(AniDB_Anime anime, AniDB_Episode episode, string query, bool includeRestricted = false, int year = 0)
     {
-        // Brute force attempt #1: With the original title and earliest known aired year.
-        var (results, totalCount) = await SearchMoviesRaw(query, includeRestricted: includeRestricted, year: year).ConfigureAwait(false);
-        var firstViableResult = results.FirstOrDefault(result => IsAnimation(result.GetGenres()));
-        if (firstViableResult is not null)
-        {
-            _logger.LogTrace("Found {Count} movie results for search on {Query}, best match; {MovieName} ({ID})", totalCount, query, firstViableResult.OriginalTitle, firstViableResult.Id);
+        var candidateCount = _settingsProvider.GetSettings().TMDB.AutoSearchCandidateCount;
```
revam: I feel like we would need different counts for movies vs. series.

Shoko.Server/Providers/TMDB/TmdbSearchService.cs
```diff
@@ -225,50 +237,95 @@ private async Task<bool> AutoSearchForMovie(List<TmdbAutoSearchResult> list, Ani
...
-                _logger.LogTrace("Found {Count} movie results for search on {Query}, best match; {MovieName} ({ID})", totalCount, strippedTitle, firstViableResult.OriginalTitle, firstViableResult.Id);
+            var full = await _tmdbService.UseClient(
+                c => c.GetMovieAsync(candidate.Id, "en-US", null, TMDbLib.Objects.Movies.MovieMethods.Translations),
+                $"Fetch candidate movie {candidate.Id} \"{candidate.OriginalTitle}\" for auto-match scoring"
+            ).ConfigureAwait(false);
+            if (full is null) continue;
+
+            var allTitles = new HashSet<string>(
+                new[] { full.Title, full.OriginalTitle }
+                    .Concat(full.Translations?.Translations?.Select(t => t.Data?.Name).OfType<string>() ?? [])
```
revam: use .WhereNoNull() instead of .OfType<T>().

Shoko.Server/Providers/TMDB/TmdbSearchService.cs
```diff
@@ -225,50 +237,95 @@ private async Task<bool> AutoSearchForMovie(List<TmdbAutoSearchResult> list, Ani
...
+                    .OfType<string>()
+                    .Where(s => !string.IsNullOrWhiteSpace(s))
```
revam, line 288: Will the title ever be an empty string or a string of only whitespaces?

Shoko.Server/Providers/TMDB/TmdbSearchService.cs
```diff
@@ -225,50 +237,95 @@ private async Task<bool> AutoSearchForMovie(List<TmdbAutoSearchResult> list, Ani
...
 
-                return new(anime, episode, firstViableResult) { IsRemote = true };
-            }
+            var rating = (exactTitle, fuzzyTitle, dateMatch) switch
+            {
+                (true, _, true)  => MatchRating.DateAndTitleMatches,
+                (true, _, false) => MatchRating.TitleMatches,
+                (_, true, true)  => MatchRating.DateAndTitleKindaMatches,
+                (_, true, false) => MatchRating.TitleKindaMatches,
+                _                => MatchRating.FirstAvailable,
```
revam, line 303: i don't see a check for the date matches. also, tmdb has the release dates for each region which can be checked against the anidb air date for 'date matches' instead of only the 'year matches' check above. since we're pulling in the raw movie object w/translations anyways then i don't see why we cannot also pull in the release dates to check.

Shoko.Server/Providers/TMDB/TmdbSearchService.cs
```diff
@@ -225,50 +237,95 @@ private async Task<bool> AutoSearchForMovie(List<TmdbAutoSearchResult> list, Ani
...
 
-        return null;
+        // If all GetMovieAsync calls returned null (transient outage), fall back to the first candidate
+        // at lowest confidence rather than returning null when TMDB did return search results.
+        if (scored.Count == 0)
+            return new(anime, episode, candidates[0]) { IsRemote = true };
+
+        var best = scored
+            .OrderBy(x => ShowMatchPriority(x.rating))
+            .ThenByDescending(x => x.voteCount)
```
revam: why is user controlled (on tmdb's side) vote count arbitrarily affecting our ordering?

Shoko.Server/Providers/TMDB/TmdbSearchService.cs
```diff
@@ -225,50 +237,95 @@ private async Task<bool> AutoSearchForMovie(List<TmdbAutoSearchResult> list, Ani
...
+            .First();
+
+        _logger.LogTrace(
+            "Best match for \"{Query}\": {MovieName} ({ID}) rating={Rating}",
```
revam, line 327: Vote count is also not logged, even though it affected which we picked.

revam: Also, we should log all candidates at trace level. best pick can be debug level.

Shoko.Server/Providers/TMDB/TmdbSearchService.cs
```diff
@@ -450,94 +525,220 @@ private async Task<IReadOnlyList<TmdbAutoSearchResult>> AutoSearchForShow(AniDB_
...
 
     #region Helpers
 
+    private static int ShowMatchPriority(MatchRating r) => r switch
+    {
+        MatchRating.UserVerified             => 0,
+        MatchRating.DateAndTitleMatches      => 1,
+        MatchRating.DateMatches              => 2,
+        MatchRating.TitleMatches             => 3,
+        MatchRating.DateAndTitleKindaMatches => 4,
```
revam: I feel like date and title kinda matches should have a higher priority then date matches.

Shoko.Server/Providers/TMDB/TmdbSearchService.cs
```diff
@@ -402,35 +459,53 @@ private async Task<IReadOnlyList<TmdbAutoSearchResult>> AutoSearchForShow(AniDB_
...
+        // When the prequel chain was followed, the searches above used the ROOT series title.
+        // If the current anime is a separate TMDB entry (e.g. "Fairy Tail: 100 Years Quest"
+        // is its own TMDB show, not a season of base Fairy Tail), the root-title search will
+        // return the wrong show. Try the current anime's own main title and prefer the result
+        // if it scores better. When the own title wins, treat it as if no prequel was followed
+        // so the root series' stored xrefs are not added as additional matches below.
+        var prequelFollowed = series.ID != anime.AnimeID;
+        if (prequelFollowed && !string.Equals(mainTitle.Value, originalTitle, StringComparison.Ordinal))
+        {
+            var ownTitleMatch = await AutoSearchForShowUsingTitle(anime, mainTitle.Value, airDate.Value, anime.IsRestricted, false).ConfigureAwait(false);
```
revam: so we're not going to try the original title before the main title here?

revam (changes requested):
a few questions and suggested changes. also, think the series search could still benefit from an episode level link check so we don't have to 100% rely on the title checks, but that can happen later, if at all. keep up the good work.

Shoko.Server/Providers/TMDB/TmdbSearchService.cs
Foowy: If TMDB does not have a translation for title, this is more of a guard against that. For TMDB the Name field is often "" when no translated title has been submitted yet.

Shoko.Server/Providers/TMDB/TmdbSearchService.cs
Foowy: Fixed in commit, github not flagging

Shoko.Server/Providers/TMDB/TmdbSearchService.cs
Foowy: per discussion in Discord, all at trace, picked is at informational level for feedback

Foowy:
Working on comparing to episode dates as well for comparison. Using AniDB as the point of truth for these date values

Foowy:
Rebased to current [https://github.com/ShokoAnime/ShokoServer/commit/58d1593f5fa84406dc4a8118410df57bb3b3a0b1](url), ready for full review whenever

revam (approved):
time for wild testing.

PR #1364, opened 2026-06-20
repo(docker): unify Dockerfile, modernize GHA manifest creation, add compose examples

Foowy:
## Summary

Three related changes to the Docker build infrastructure:

- **Unified Dockerfile** — removes `Dockerfile.aarch64`; a single Dockerfile now handles both `amd64` and `arm64` via `--platform=$BUILDPLATFORM` and a `TARGETARCH`→.NET RID mapping (`amd64→linux-x64`, `arm64→linux-arm64`). Cross-compilation runs natively on the amd64 build host via `dotnet publish -r`, not QEMU emulation.
- **GHA manifest creation** — replaces the deprecated `docker manifest create/push` with `docker buildx imagetools create` across all three workflows (`build-daily`, `build-release`, `docker-manual`). Requires `docker/setup-buildx-action@v3` in manifest jobs (added).
- **Compose examples** — adds `docker-compose.yml` (SQLite default), `docker-compose.mssql.yml` (MSSQL override), and `docker-compose.mysql.yml` (MySQL/MariaDB override) for users standing up a local stack.

## ⚠️ Needs maintainer review — consequences I cannot fully test

I do not have access to the upstream CI secrets (`DOCKER_REPO`, `DOCKER_USERNAME`, `DOCKER_PASSWORD`, `SENTRY_DSN`, `TMDB_API`, `AVD3_URL`) or the upstream runner environment, so I cannot end-to-end verify the GHA changes produce correctly tagged and pushed images. The logic changes are sound from a reading of the workflow, but the manifest push path in particular needs someone who can observe a real CI run to confirm.

**Separately: I have nagging concerns about the current build process that this PR does not fully solve.** The existing workflows spend significant time on steps that feel redundant or poorly cached — NuGet restore re-runs across jobs, the WebUI download is sequential where it could be parallelised, and the multi-job structure for what is essentially `build → push` adds handoff overhead. I've addressed the most obvious inefficiency (NuGet layer caching, GHA build cache), but I suspect there's more to gain from a broader look at the job graph. Raising this now so it's on the radar, not as a blocker for this PR.

## Breaking / notable changes

**`Dockerfile.aarch64` is deleted.** Any external script or fork referencing it directly will need updating. The GHA matrix drops the `dockerfile` column and now uses the unified `Dockerfile` unconditionally.

**`docker buildx imagetools create` requires Buildx ≥ 0.10.** GitHub-hosted `ubuntu-latest` satisfies this. Self-hosted runners on Docker Engine < 23 or without the Buildx plugin will fail at the manifest step.

**Release manifest job consolidation.** The previous approach ran one GHA job per tag in parallel (a 4-way matrix over `latest`, semver, major, minor). The new approach is a single job with a shell loop. Upside: simpler job graph. Tradeoff: a registry error on one tag now aborts the remaining tags rather than letting them complete independently. Empty-tag values are skipped via `[[ -z "$t" ]] && continue`.

**Key replacement steps are now conditional.** `Replace Sentry DSN and other keys` (single always-runs step) is split into three separate steps, each guarded by `if: ${{ secrets.SECRET != '' }}`. Steps are skipped gracefully in forks or environments where the secrets are not configured.

## Other changes

- NuGet restore scoped to the 4 CLI-dependency `.csproj` files before `COPY . .` to maximise layer cache reuse.
- GHA build cache added (`type=gha`, per-workflow per-arch scope) across all three workflows.
- `.dockerignore` expanded to exclude test, bench, and tray projects; `*.sln`, `*.md`, `docs/`, and `.github/` from the build context (none are needed by the CLI build).
- `Dockerfile.aarch64` removed from `Shoko.Server.sln` solution items.


Foowy:
This is purely a attempt at making docker process cleaner, by no means is this a end all be all. If we add new projects to shoko it will need to be revisited, but current state just feels "Wrong" to me.

.github/workflows/build-daily.yml
```diff
@@ -631,11 +628,22 @@ jobs:
           name: offline-importer
           path: Shoko.Server/plugins
 
-      - name: Replace Sentry DSN and other keys
+      - name: Replace Sentry DSN
```
revam, line 631: If we're going to update these, then change it to 'Add' instead of 'Replace'.

.github/workflows/build-daily.yml
```diff
@@ -660,14 +668,13 @@ jobs:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
 
-      # Disabled provenance for now, until it works with docker manifest create.
-      # The manifest list produced by the new feature is incompatible with the
-      # expected format used in the docker manifest create command.
+      # provenance disabled: docker buildx imagetools create (used in the
```
revam, line 671: I see no reason to edit this comment.

.github/workflows/build-daily.yml
```diff
@@ -680,6 +687,8 @@ jobs:
             date=${{ needs.current_info.outputs.date }}
             tag=${{ needs.current_info.outputs.tag }}
           provenance: false
+          cache-from: type=gha,scope=docker-daily-${{ matrix.arch }}
```
revam, line 690: Pretty sure this will give us the 5m cache upload penalty again, but we can try if you _really_ think this is necessary. Personally I don't see why we even need the caching when it builds in 1.5m without it.

.github/workflows/build-daily.yml
```diff
@@ -690,6 +699,9 @@ jobs:
       - docker-build
 
     steps:
```
revam, line 701: Don't pin it. See other comment.

.github/workflows/build-daily.yml
```diff
@@ -703,15 +715,16 @@ jobs:
...
-          docker manifest create ${{ secrets.DOCKER_REPO }}:daily --amend ${{ secrets.DOCKER_REPO }}:daily-amd64 --amend ${{ secrets.DOCKER_REPO }}:daily-arm64
-
-      - name: Create & push manifest (Docker Hub)
+      - name: Create and push multi-arch manifest
+        env:
+          REPO: ${{ secrets.DOCKER_REPO }}
         run: |
-          docker manifest push ghcr.io/${{ secrets.DOCKER_REPO }}:daily
-          docker manifest push ${{ secrets.DOCKER_REPO }}:daily
+          docker buildx imagetools create -t ghcr.io/${REPO}:daily \
```
revam, line 722: Please revert this. They were separate steps for a reason. You also shifted the container image to not be in sync with the GH repo name.

revam: Instead make the non-GHCR not run if the repo name is not set.

.github/workflows/build-release.yml
```diff
@@ -511,7 +519,7 @@ jobs:
           platforms: arm64
         if: ${{ matrix.arch == 'arm64' }}
 
-      - uses: docker/setup-buildx-action@v3
+      - uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f  # v3
```
revam, line 522: Don't pin it. Please revert. If we were going to pin, the actions then we would pin ALL of them. But we don't do pinning, so once again, please revert to the tag.

.github/workflows/build-release.yml
```diff
@@ -527,14 +535,13 @@ jobs:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
 
-      # Disabled provenance for now, until it works with docker manifest create.
-      # The manifest list produced by the new feature is incompatible with the
-      # expected format used in the docker manifest create command.
+      # provenance disabled: docker buildx imagetools create (used in the
```
revam, line 538: Again, same as the other review comment.

docker-compose.mssql.yml
```diff
@@ -0,0 +1,44 @@
+# Shoko Server — Microsoft SQL Server override
```
revam, line 1: MS SQL and MySQL/MariaDB are not _officially_ supported, so we don't need docker compose examples for them. Maybe after the EF Core refactor they will become _officially_ supported. Also, there are ways to set the db type through env vars so you don't need to do the "setup" below, but I won't go into details here as to exactly how, because again, it's not _officially_ supported.

docker-compose.mysql.yml
```diff
@@ -0,0 +1,46 @@
+# Shoko Server — MySQL / MariaDB override
```
revam, line 1: See other comment.

docker-compose.yml
```diff
@@ -0,0 +1,24 @@
+# Shoko Server — SQLite (default)
```
revam, line 1: While we can add this, I would rather we stick to the dynamically generated compose file in the docs, so we don't have 2 places we need to edit if we ever change our recommended compose file setup.

revam (changes requested):
Some comments.

.github/workflows/build-daily.yml
revam: We can switch to buildx, but don't combine them.

PR #1365, opened 2026-06-22
Modernise Docker build pipeline

Foowy:
## Summary

This PR removes the two previous attempts at caching (both reverted) by fixing the structural root cause, then layers a proper caching strategy on top.

**Why the old GHA cache never worked:**
Both Dockerfiles opened the build stage with `COPY . /usr/src/app/source`, copying the entire source tree before NuGet restore. Any commit invalidated every downstream layer. The cache uploaded after a full build; the next run hit the same invalidation point and rebuilt from scratch anyway — explaining the 4–5 minute upload overhead with zero benefit. The GHA cache pool (10 GB shared, LRU eviction) is also unsuitable for .NET SDK layers (~800 MB+).

---

### Dockerfile
- Rewritten as a pure packaging step — no build stage, no SDK image.
- `Dockerfile.aarch64` deleted; a single `Dockerfile` handles both platforms via `ARG TARGETARCH` and `COPY ./publish/${TARGETARCH} .`.
- Hygiene: `apt-utils` removed, `--no-install-recommends` added to both `apt-get install` calls, `rm -rf /var/lib/apt/lists/*` added, deprecated `MAINTAINER` removed.
- Inline comment block documents the outside-Docker rationale, the inside-Docker tradeoff, and local build instructions.

### `build-daily.yml` / `build-release.yml` / `docker-manual.yml`
- **Compilation moved outside Docker.** Each workflow now has a `cli` job that runs `dotnet publish` for `linux-x64` and `linux-arm64`, uploads workflow artifacts, and the docker job downloads them into `publish/amd64/` and `publish/arm64/` before calling `docker build`. `TARGETARCH` selects the correct directory at build time.
- **Native multi-platform build.** `docker-build` / `docker-release-build` / `docker-manual-build` are no longer matrices — each is a single job using `build-push-action@v7` with `platforms: linux/amd64,linux/arm64`. BuildKit produces a proper OCI index manifest directly.
- **Manifest assembly jobs deleted.** `docker-push-manifest`, `docker-release-push_manifest`, and `docker-manual-push_manifest` are gone. The per-arch intermediate tags (`:daily-amd64`, `:daily-arm64`) are also gone.
- **Registry layer cache.** `cache-from`/`cache-to` added to all three docker jobs using `type=registry,ref=ghcr.io/${{ secrets.DOCKER_REPO }}:cache-buildkit,mode=max`. With no build stage, only the runtime apt/mediainfo layers are cached — these change rarely (only on base image or package list changes), so hit rate will be high. Cache is shared across daily/release/manual builds via the same tag.
- **NuGet runner cache.** `actions/cache@v5` on `~/.nuget/packages` added to `cli`, `cli-release`, and `tray-service-framework-cross` jobs. Keyed on `matrix.rid` + hash of `*.csproj`/`*.sln` files. Runner-level cache is faster than layer cache and shared within a workflow run.
- **`Framework linux-arm64` added to CLI matrix** in both daily and release workflows. Previously this artifact was produced by building inside Docker from source; it now comes from `dotnet publish` like the x64 artifact.
- **`provenance: true`** — the old `provenance: false` workaround (required because `docker manifest create` was incompatible with provenance attestation format) is removed. Native multi-platform builds via `build-push-action` produce a proper OCI index that includes provenance.
- **Pack folded into publish step** for `plugin-nuget` and `queue-nuget` in both `build-daily.yml` and `build-release.yml`. `dotnet pack` now runs inside the `if: NUGET_API_KEY != ''` step, so no artifact is produced in forks or environments without publish access.

### `brandedoutcast/publish-nuget@v2.5.2` — reliability fix

> **Reviewers: this is a reliability fix, not a cosmetic action version bump.**

`brandedoutcast/publish-nuget@v2.5.2` was last updated in 2020 and uses the deprecated `::set-output::` command syntax that GitHub made a hard error in 2024. It currently produces warnings but can silently fail to set outputs. The failure mode is serious: `plugin-nuget` and `queue-nuget` are listed in `needs:` for `cli`, `tray-service-*`, and `docker-build` — a hard failure there stops the entire build.

Replaced in both `build-daily.yml` and `build-release.yml` with `dotnet pack` + `dotnet nuget push --skip-duplicate` directly. No third-party action, no version-pinning risk.

### Action version updates

All remaining GitHub Actions dependencies updated to their latest versions. The majority of updated actions have moved their runtime to Node.js 24, dropping Node.js 20. See also: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

| Action | Old | New |
|---|---|---|
| `actions/checkout` | v4 / @master | v7 |
| `actions/cache` | v4 | v5 |
| `actions/upload-artifact` | v4 | v7 |
| `actions/download-artifact` | v4 | v8 |
| `actions/github-script` | v7 | v9 |
| `actions/setup-dotnet` | v4 | v5 |
| `actions/setup-node` | v4 | v6 |
| `docker/build-push-action` | v6 | v7 |
| `docker/login-action` | v3 | v4 |
| `docker/setup-buildx-action` | v3 | v4 |
| `docker/setup-qemu-action` | v3 | v4 |

### Minor fixes
- `Dockerfile.aarch64` removed from `Shoko.Server.sln` Solution Items.

---

## Expected impact

| | Before | After |
|---|---|---|
| SDK image pull per docker job | ~30–60s | Eliminated |
| NuGet restore (code-only change) | ~60s (inside Docker, always) | ~0s (runner cache hit) |
| Runtime layer (apt + mediainfo) | Rebuilt every docker job | Registry cache hit (rarely invalidates) |
| Cache upload overhead | 4–5 min (GHA, always a miss) | ~10–20s incremental (registry, runtime only) |
| Manifest assembly | `docker manifest create` in a separate job | Eliminated |
| Per-arch intermediate tags | `:daily-amd64`, `:daily-arm64` in registries | Eliminated |


Foowy:
**SonarCloud S7637 — action SHA pinning**

SonarCloud flagged the new `docker/*` action references in this PR for using mutable tags (`@v3`, `@v6`) instead of full commit SHAs. This is a supply-chain security concern.

The latest commit pins all `actions/*` and `docker/*` references across the three affected workflow files to their current full SHAs, with the version tag preserved as a comment for readability.

**Note on pre-existing state:** `master` already contained unpinned `docker/*` and `actions/*` references in the existing docker-build jobs. SonarCloud's PR analysis only reports issues in the diff, so those were invisible to Sonar until this PR introduced new docker jobs. The pinning commit covers both the new code and the pre-existing unpinned references in these three files, making the files consistent. The one remaining `actions/checkout@master` reference (in the `sentry-upload` job, outside docker scope) is pre-existing and left for a separate cleanup.

.github/workflows/tray-standalone-manual.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           rm latest.zip
 
       - name: Upload Web UI Artifact
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
```
revam, line 45: this should be rebased onto master again, since the other PR touching the workflows have been merged.

.github/workflows/build-daily.yml
```diff
@@ -228,32 +228,35 @@ jobs:
...
           dotnet-version: ${{ matrix.dotnet }}
 
-      - name: Run dotnet build
+      - name: Build Shoko.Abstractions
         run: dotnet build -c Release Shoko.Abstractions
 
-      - name: Publish Shoko.Abstractions Nuget
-        uses: brandedoutcast/publish-nuget@v2.5.2
+      - name: Pack Shoko.Abstractions
+        run: dotnet pack -c Release --no-build Shoko.Abstractions/Shoko.Abstractions.csproj
```
revam: this can be folded into the publish step, which will ensure we're only packing the release if a valid NuGet API key is set.

.github/workflows/build-daily.yml
```diff
@@ -262,25 +265,28 @@ jobs:
...
           dotnet-version: ${{ matrix.dotnet }}
 
-      - name: Run dotnet build
+      - name: Build Shoko.QueueProcessor
         run: dotnet build -c Release Shoko.QueueProcessor
 
-      - name: Publish Shoko.QueueProcessor Nuget
-        uses: brandedoutcast/publish-nuget@v2.5.2
+      - name: Pack Shoko.QueueProcessor
+        run: dotnet pack -c Release --no-build Shoko.QueueProcessor/Shoko.QueueProcessor.csproj
```
revam: See above comment.

revam (changes requested):
for now, fix these three things, then we'll take it from there.

.github/workflows/build-daily.yml
```diff
@@ -638,80 +649,46 @@ jobs:
           ./.github/workflows/ReplaceTmdbApiKey.ps1 -apiKey ${{ secrets.TMDB_API }}
           ./.github/workflows/ReplaceAVD3URL.ps1 -url ${{ secrets.AVD3_URL }}
 
-      - uses: docker/setup-qemu-action@v4
-        name: Set up QEMU
+      - name: Set up QEMU
```
revam: Flip the `name` and `uses` lines for these steps.

.github/workflows/build-release.yml
```diff
@@ -505,96 +531,52 @@ jobs:
           ./.github/workflows/ReplaceTmdbApiKey.ps1 -apiKey ${{ secrets.TMDB_API }}
           ./.github/workflows/ReplaceAVD3URL.ps1 -url ${{ secrets.AVD3_URL }}
 
-      - uses: docker/setup-qemu-action@v4
-        name: Set up QEMU
+      - name: Set up QEMU
```
revam: Same as above.

.github/workflows/build-release.yml
```diff
@@ -244,6 +266,14 @@ jobs:
           tag: ${{ needs.current_info.outputs.tag }}
           file_glob: true
 
+      - name: Upload Artifacts
+        if: matrix.build_type == 'Framework'
+        uses: actions/upload-artifact@v7
+        with:
+          name: Shoko.CLI_${{ matrix.build_type }}_${{ matrix.rid }}
+          path: Shoko.Server/bin/Release/net10.0/${{ matrix.rid }}/publish
+          retention-days: 5
```
revam: Since the artifacts are only used for the docker builds, then we can tune down the retention to 1 day. Alternatively, we can fetch it from the release instead of using the artifact storage in the release workflow.

.github/workflows/docker-manual.yml
```diff
@@ -52,113 +52,147 @@ jobs:
...
+      - name: Retrieve Commit Date
+        id: version
+        run: echo "COMMIT_DATE=$(git log -1 --format='%aI')" >> $GITHUB_OUTPUT
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v4
+        with:
+          platforms: arm64
+
+      - name: Set up Docker Buildx
```
revam: See above comment about flipping. I also think the diff view will change for this file once the props are swapped.

Dockerfile
```diff
@@ -1,42 +1,62 @@
...
     PGID=100 \
     LANG=C.UTF-8 \
     LC_CTYPE=C.UTF-8 \
     LC_ALL=C.UTF-8
 
-RUN apt-get update && apt-get install -y gnupg curl \
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends gnupg curl \
     && curl --retry 3 -O https://mediaarea.net/repo/deb/debian/pubkey.gpg \
-    # This converts the old format gpg key to the new format. The old format cannot be used with [signed-by] in the apt sources.list file.
```
revam, line 26: Any reason this comment was removed?

Dockerfile
```diff
@@ -1,42 +1,62 @@
...
     && gpg --no-default-keyring --keyring ./temp-keyring.gpg --import pubkey.gpg \
     && gpg --no-default-keyring --keyring ./temp-keyring.gpg --export --output /usr/share/keyrings/mediainfo.gpg \
     && rm pubkey.gpg \
-    && echo "deb [signed-by=/usr/share/keyrings/mediainfo.gpg] https://mediaarea.net/repo/deb/debian/ bookworm main" | tee -a /etc/apt/sources.list \
-    && apt-get update && apt-get install -y apt-utils gosu mediainfo librhash-dev
+    && echo "deb [signed-by=/usr/share/keyrings/mediainfo.gpg] https://mediaarea.net/repo/deb/debian/ bookworm main" \
```
revam: Line break at 180c is fine, no need to convert it to line breaks at 120c.

Dockerfile
```diff
@@ -1,42 +1,62 @@
...
+# TARGETARCH is provided automatically by BuildKit for multi-platform builds (values: amd64, arm64).
+# CI stages pre-built artifacts under publish/amd64/ and publish/arm64/ before docker build.
+ARG TARGETARCH
+COPY ./publish/${TARGETARCH} .
 COPY ./dockerentry.sh /dockerentry.sh
 
 VOLUME /home/shoko/.shoko/
 
-HEALTHCHECK --start-period=5m CMD curl -s -H "Content-Type: application/json" -H 'Accept: application/json' 'http://localhost:8111/api/v3/Init/Status' || exit 1
+HEALTHCHECK --start-period=5m \
```
revam: See above comment.

revam (changes requested):
Last round of questions and change requests (I hope).

Dockerfile
Foowy: Bad copy pasta, my oops

revam (approved):

PR #1369, opened 2026-06-25
fix(api): eliminate N+1 query in Filter/Preview/Group endpoint

Baine:
## Problem

The endpoint \/api/v3/Filter/Preview/Group\ took **over 2 hours** for a single page of 50 results.

## Root Cause

The \GetFilteredGroups\ method in \FilterController.cs\ had a severe N+1 query pattern:

- **Pre-pagination scan:** \	.group.AllSeries.Any(s => s.VideoLocals.Count > 0)\ executed for **every group** (thousands) before pagination. \AllSeries\ recursively walks all child groups (NHibernate queries per child), and \VideoLocals.Count\ triggers additional repository lookups per series.
- **Double construction:** The \Group\ constructor then loaded \AllSeries\, \AllAnimeEpisodes\, \VideoLocals\, and \GetImages\ for every group **again** — all before \ToListResult(page, pageSize)\ applied pagination.

## Fix

Pre-compute everything from in-memory caches **before** pagination:

1. One pass over \CrossRef_File_Episode.Cache.Values\ (O(total xrefs)) to build the set of animeIDs that have files
2. For each group, recursive check via cached \AnimeSeriesRepository.GetByGroupID\ (in-memory PocoIndex) + \CrossRef_File_Episode\ cache membership
3. Filter via dictionary lookup → paginate → **then** construct \Group\ objects (only for 50 paginated groups)

## Result

**2 hours → ~4 seconds** (~1800x improvement)

## Changes

- \Shoko.Server/API/v3/Controllers/FilterController.cs\: Rewrote \GetFilteredGroups()\ to defer expensive operations until after pagination. Added \CrossRef_File_EpisodeRepository\ dependency injection.

### DIFF
--- a/Shoko.Server/API/v3/Controllers/FilterController.cs
+++ b/Shoko.Server/API/v3/Controllers/FilterController.cs
@@ -33,7 +33,8 @@ public class FilterController(
     IMetadataFilteringService filteringService,
     AnimeGroupRepository _animeGroups,
     AnimeSeriesRepository _animeSeries,
-    FilterPresetRepository _filterPresets
+    FilterPresetRepository _filterPresets,
+    CrossRef_File_EpisodeRepository _crossRefFileEpisode
 ) : BaseController(settingsProvider)
 {
     internal const string FilterNotFound = "No Filter entry for the given filterID";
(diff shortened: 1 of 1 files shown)


revam:
This won't fix the "2h" issue since `includeEmpty` defaults to `true` if not set in the query. And our filtering requests don't set it in the query. the bottleneck is elsewhere.

PR #1370, opened 2026-06-26
feat(tmdb): add 5XX circuit breaker with escalating backoff and force-run UX

Foowy:
## Summary

Adds a 5XX circuit breaker to `TmdbRateLimiter` that detects sustained TMDB server distress and pauses all TMDB API job dispatch until the server recovers, preventing pointless retry floods. Jobs queued during a pause are held and dispatched automatically when the pause lifts — no work is lost.

**Circuit breaker behaviour**
- A ring buffer tracks the last 3 distress-error timestamps. When all 3 fall within a 10-second window the breaker trips.
- Pause duration escalates with each successive trip: 1 → 3 → 5 → 15 → 60 min (capped at level 5).
- On trip, the ring buffer is cleared so errors from requests already in flight when the breaker tripped don't immediately re-trip and escalate the level further.
- A `Task.Delay`-based timer auto-clears the pause when the duration elapses, so the queue resumes even if no TMDB request is made to call `NotifySuccess` (which would otherwise deadlock behind the acquisition filter).
- The pause also clears on the first successful request after the backoff window passes (`NotifySuccess`), which resets the ramp level and ring buffer. A cheap pre-check skips the lock entirely once the breaker is idle, since every successful TMDB call (up to 10 concurrent) would otherwise funnel through it for nothing.
- 429 backoff and 5XX pause share a single `_backoffUntilTicks` deadline. Both interact safely under `_breakerLock`.
- `GetPauseSnapshot()` reads the paused flag and remaining-time deadline together under the lock, since both are cleared as a pair elsewhere and reading them independently could observe a mid-flip state (paused but no remaining time, or vice versa).

**Job gating**
- New `[TmdbApiRateLimited]` acquisition attribute (extends `[NetworkRequired]` at TMDB worker priority).
- `TmdbApiRateLimitedAcquisitionFilter` blocks all `[TmdbApiRateLimited]` jobs from dispatch while the circuit breaker is tripped, and fires `StateChanged` when the pause toggles so the queue wakes immediately. Implements `IDisposable` via the standard dispose pattern.
- `DownloadTmdbShowImagesJob` and `DownloadTmdbMovieImagesJob` gained `[TmdbApiRateLimited]` (previously missing despite making API calls).
- `PurgeTmdbMovieJob` and `PurgeTmdbShowJob` reverted to `[NetworkRequired]` (they only do local DB deletes, no TMDB API calls).

**Force-run UX**
- When a user force-triggers a refresh or image download while the circuit breaker is active, the job is queued at elevated priority and the endpoint returns **503** with a `Retry-After` header instead of 202. The log emits the approximate wait in seconds.
- The quick-refresh early-return (already updating, episodes present) is answered entirely from local state, so it's checked before the pause gate — it never touches TMDB and shouldn't be refused just because TMDB itself is unavailable.
- `QuickRefresh` is only honored on the immediate execution path (`body.Immediate && body.QuickRefresh`); scheduled/background show refreshes always run the full job with `ForceRefresh` as requested, applied consistently across all force paths.
- Job deduplication in `IQueueScheduler.Enqueue` prevents users from stacking duplicate jobs while paused.

**API surface**
- `TmdbRateLimiter.GetPauseSnapshot()` returns `(IsPaused, Remaining)` as an atomic, lock-protected read so callers never observe a TOCTOU split between the two fields.
- `TmdbMetadataService` exposes `Is5xxPaused`, `RemainingPauseTime`, and `GetPauseSnapshot()` as pass-throughs so the controller doesn't need to inject `TmdbRateLimiter` directly.

**Code quality / concurrency**
- `_5xxPauseLevel` is `volatile` so the lock-free fast-path check in `NotifySuccess` is safe.
- Removed a `fireEvent` variable that was always `true` at its read site (early returns inside the lock exit the lambda entirely, so the log/event only fire on the path where the pause was actually cleared).

## Rebase history
- 2026-07-04 20:05 UTC — rebased onto `master` @ `6dd36ed5e` (2 commits), now at `a6a36b39c`

### DIFF
--- a/Shoko.Server/API/v3/Controllers/TmdbController.cs
+++ b/Shoko.Server/API/v3/Controllers/TmdbController.cs
@@ -95,6 +95,32 @@ public partial class TmdbController : BaseController
         _tmdbShows = tmdbShows;
     }
 
+    // When paused and the caller can wait (not immediate), queue the job (prioritized so it jumps
+    // the backlog on resume) and return 503 with Retry-After. Dedup in Enqueue prevents the same job
+    // from stacking while paused. When the caller wanted an immediate result, queuing for later is
+    // useless — the UI is waiting synchronously — so just refuse with 503 instead.
+    //
+    // This only gives a fast, friendly 503+Retry-After response — it is not the safety net. Any TMDB
+    // job type dispatched here must also carry [TmdbApiRateLimited] (see TmdbApiRateLimitedAcquisitionFilter)
+    // so dispatch is still blocked at the queue layer even if a call site here forgets to check the pause.
+    private async Task<ActionResult?> TryQueueWhenPaused<T>(Action<T> configure, string jobDescription, bool immediate) where T : class, IQueueJob
+    {
+        var status = _tmdbMetadataService.GetPauseStatus();
+        if (!status.IsPaused) return null;
+        var seconds = (int)(status.RemainingPauseTime?.TotalSeconds ?? 0);
+        if (immediate)
+        {
+            _logger.LogInformation("TMDB is currently paused. {Job} was requested immediately and has been refused; retry in approximately {Seconds} second(s).", jobDescription, seconds);
+        }
+        else
+        {
+            _logger.LogInformation("TMDB is currently paused. {Job} has been queued and will start in approximately {Seconds} second(s).", jobDescription, seconds);
+            await _scheduler.StartJob(configure, prioritize: true);
+        }
+        Response.Headers.RetryAfter = seconds.ToString();
+        return StatusCode(503);
+    }
+
     #region Movies
 
     #region Constants
@@ -604,27 +630,24 @@ public partial class TmdbController : BaseController
                 return Ok();
         }
 
-        if (body.Immediate)
-        {
-            await _jobFactory.Execute<UpdateTmdbMovieJob>(j =>
-            {
-                j.TmdbMovieID = movieID;
-                j.ForceRefresh = body.Force;
-                j.DownloadImages = body.DownloadImages;
-                j.DownloadCrewAndCast = body.DownloadCrewAndCast;
-                j.DownloadCollections = body.DownloadCollections;
-            });
-            return Ok();
-        }
-
-        await _scheduler.StartJob<UpdateTmdbMovieJob>(j =>
+        Action<UpdateTmdbMovieJob> configure = j =>
         {
             j.TmdbMovieID = movieID;
             j.ForceRefresh = body.Force;
             j.DownloadImages = body.DownloadImages;
             j.DownloadCrewAndCast = body.DownloadCrewAndCast;
             j.DownloadCollections = body.DownloadCollections;
-        });
+        };
+        if (await TryQueueWhenPaused(configure, "Movie refresh", body.Immediate) is { } pausedMovie)
+            return pausedMovie;
+
+        if (body.Immediate)
+        {
+            await _jobFactory.Execute(configure);
+            return Ok();
+        }
+
+        await _scheduler.StartJob(configure);
         return NoContent();
     }
 
@@ -648,21 +671,21 @@ public partial class TmdbController : BaseController
         if (movie is null)
             return NotFound(MovieNotFound);
 
-        if (body.Immediate)
-        {
-            await _jobFactory.Execute<DownloadTmdbMovieImagesJob>(j =>
-            {
-                j.TmdbMovieID = movieID;
-                j.ForceDownload = body.Force;
-            });
-            return Ok();
-        }
-
-        await _scheduler.StartJob<DownloadTmdbMovieImagesJob>(j =>
+        Action<DownloadTmdbMovieImagesJob> configure = j =>
         {
             j.TmdbMovieID = movieID;
             j.ForceDownload = body.Force;
-        });
+        };
+        if (await TryQueueWhenPaused(configure, "Movie image download", body.Immediate) is { } pausedMovieImages)
+            return pausedMovieImages;
+
+        if (body.Immediate)
+        {
+            await _jobFactory.Execute(configure);
+            return Ok();
+        }
+
+        await _scheduler.StartJob(configure);
         return NoContent();
     }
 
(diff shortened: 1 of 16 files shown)


Shoko.Server/Providers/TMDB/TmdbRateLimiter.cs
```diff
@@ -146,4 +163,87 @@ private async Task WaitForBackoffAsync(CancellationToken cancellationToken = def
...
+                _5xxPauseLevel = 0;
+                // Clear ring buffer so a single 5xx after recovery doesn't immediately re-trip.
+                Array.Clear(_errorTimestamps, 0, _errorTimestamps.Length);
+                _errorSlot = 0;
+            }
+        }
+    }
+
+    /// <summary>
+    /// Maps a ramp level (1–5) to a pause duration. Level 0 (30s) is defined but unreachable —
```
revam: why is level 0 defined if it's never used?

Shoko.Server/Providers/TMDB/TmdbRateLimiter.cs
```diff
@@ -27,6 +27,23 @@ public sealed class TmdbRateLimiter : IDisposable
...
+
+    private int _errorSlot;
+
+    private readonly long _errorWindowTicks;
+
+    // Guards _5xxPauseLevel and all writes to _backoffUntilTicks so that the level/deadline
+    // pair is always updated atomically. Reads of _backoffUntilTicks from WaitForBackoffAsync
+    // and EnsureRateAsync happen outside this lock via Interlocked.Read — that is safe because
+    // lock exit provides a release fence and Interlocked.Read provides an acquire fence.
+    private readonly object _breakerLock = new();
```
revam: use `Lock` for (new) locks, not `object`.

Shoko.Server/Providers/TMDB/TmdbRateLimiter.cs
```diff
@@ -146,4 +163,87 @@ private async Task WaitForBackoffAsync(CancellationToken cancellationToken = def
...
+    /// Called by <see cref="Notify5xxError"/> and exposed internally for tests.
+    /// </summary>
+    internal static TimeSpan Get5xxPauseDuration(int level) => level switch
+    {
+        0 => TimeSpan.FromSeconds(30),
+        1 => TimeSpan.FromSeconds(60),
+        2 => TimeSpan.FromSeconds(90),
+        3 => TimeSpan.FromSeconds(120),
+        4 => TimeSpan.FromSeconds(180),
+        _ => TimeSpan.FromSeconds(240),
```
revam: We can add a 10m pause if we reach level 5, instead of a 4m pause. if we failed 4 times, chances are, it's going to take a while before we can continue.

revam (changes requested):
Also, we should ideally add a guard to block new TMDB API jobs in the queue from running (images are fine) if we're currently in a pause because of 5xx errors. Ask @da3dsoul for help on how to implement it if you can't figure it out looking at the existing queue filters.

Shoko.Server/Providers/TMDB/TmdbRateLimiter.cs
revam: We can probably increase it even more too.

Foowy:
I extended out catch to all calls for force-refresh and soft-refresh to queue if in a 500 blocking state. Force refresh will queue with priority.

Shoko.Server/API/v3/Controllers/TmdbController.cs
```diff
@@ -604,27 +617,24 @@ public async Task<ActionResult> RefreshTmdbMovieByMovieID(
...
+        Action<UpdateTmdbMovieJob> configure = j =>
         {
             j.TmdbMovieID = movieID;
             j.ForceRefresh = body.Force;
             j.DownloadImages = body.DownloadImages;
             j.DownloadCrewAndCast = body.DownloadCrewAndCast;
             j.DownloadCollections = body.DownloadCollections;
-        });
+        };
+        if (await TryQueueWhenPaused(configure, "Movie refresh") is { } pausedMovie)
```
revam: if we want it immediately, then it's usually because we need the data for the UI. Scheduling it for later won't work, and we should instead throw if it's not possible to run it immediately because of 5xx errors instead of scheduling it to be ran in the queue at "some later time" — which is _too late_ for the UI waiting for the data.

Shoko.Server/API/v3/Controllers/TmdbController.cs
```diff
@@ -1757,34 +1767,30 @@ public async Task<ActionResult> RefreshTmdbShowByShowID(
...
+        {
+            j.TmdbShowID = showID;
+            j.ForceRefresh = !body.QuickRefresh && body.Force;
+            j.QuickRefresh = body.QuickRefresh;
+            j.DownloadImages = body.DownloadImages;
+            j.DownloadCrewAndCast = body.DownloadCrewAndCast;
+            j.DownloadAlternateOrdering = body.DownloadAlternateOrdering;
+            j.DownloadNetworks = body.DownloadNetworks;
+        };
+        if (await TryQueueWhenPaused(configure, "Show refresh") is { } pausedShow)
```
revam: See above comment.

Shoko.Server/API/v3/Controllers/TmdbController.cs
```diff
@@ -1757,34 +1767,30 @@ public async Task<ActionResult> RefreshTmdbShowByShowID(
...
         {
             // If we want quick results, we're already running an update, and we already have episodes to use, then just return early.
             if (body.QuickRefresh && _tmdbMetadataService.IsShowUpdating(showID) && _tmdbEpisodes.GetByTmdbShowID(showID).Count > 0)
                 return Ok();
 
-            await _jobFactory.Execute<UpdateTmdbShowJob>(j =>
-            {
-                j.TmdbShowID = showID;
-                j.ForceRefresh = !body.QuickRefresh && body.Force;
-                j.QuickRefresh = body.QuickRefresh;
```
revam, line 1770: Notice how QuickRefresh is only forwarded if the job is to be ran immediately? Well, that's gone during this refactor.

Shoko.Server/API/v3/Controllers/TmdbController.cs
```diff
@@ -648,21 +658,21 @@ public async Task<ActionResult> DownloadImagesForTmdbMovieByMovieID(
         if (movie is null)
             return NotFound(MovieNotFound);
 
+        Action<DownloadTmdbMovieImagesJob> configure = j =>
+        {
+            j.TmdbMovieID = movieID;
+            j.ForceDownload = body.Force;
+        };
+        if (await TryQueueWhenPaused(configure, "Movie image download") is { } pausedMovieImages)
```
revam: See above comment.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -198,6 +198,12 @@ public static string ImageServerUrl
 
     private readonly TmdbRateLimiter _rateLimiter;
 
+    public bool Is5xxPaused => _rateLimiter.Is5xxPaused;
```
revam: Maybe expose these in the abstraction so plugins can also drag use of them, and not just us internally?

revam: Just add a model for all fields, so the 'is paused due to upstream error' (previously 'is 5xx paused'), and the 'remaining pause time.' Also, why are we exposing them separately and in tuple form?

Shoko.Server/Providers/TMDB/TmdbRateLimiter.cs
```diff
@@ -146,4 +165,161 @@ private async Task WaitForBackoffAsync(CancellationToken cancellationToken = def
...
+    /// Maps a ramp level (1–5) to a pause duration.
+    /// Called by <see cref="Notify5xxError"/> and exposed internally for tests.
+    /// </summary>
+    internal static TimeSpan Get5xxPauseDuration(int level) => level switch
+    {
+        1 => TimeSpan.FromMinutes(1),
+        2 => TimeSpan.FromMinutes(3),
+        3 => TimeSpan.FromMinutes(5),
+        4 => TimeSpan.FromMinutes(15),
+        _ => TimeSpan.FromMinutes(30),
```
revam: we can do 1h for the last level. they've already had problems for 24m (1+3+5+15), so we might as well wait a full hour between checks.

revam (changes requested):

revam:
Also, good job adding the job guards.

Foowy:
Question - Realized when looking up another item. How committed are we to using polly library? Not looking to replace, just wondering if i should rewrite using polly since we are using it else where in code. In essence redoing my circuit breaker code to use polly built-in circuit breaker process? The advanced circuit breaker (supports same structure idea I am using - faults occur within a time period). Since this code isn't in daily yet, can avoid a refactor later.

Foowy:
rebased as of 13:25 (utc) 02/07/2026

revam (approved):

PR #1371, opened 2026-06-26
feat(tmdb,scheduling): add daily job to purge orphaned TMDB data

Foowy:
## Summary

Adds a background job that automatically purges TMDB shows and movies from the local database when they have no AniDB cross-references and have not been updated within a configurable threshold period.

- **New setting** `AutoPurgeUnlinkedAfterDays` (default: 14, range: 0–365, env: `TMDB_AUTO_PURGE_UNLINKED_AFTER_DAYS`) — set to `0` to disable automatic purging entirely
- **`PurgeAllUnusedShows` / `PurgeAllUnusedMovies`** on `ITmdbMetadataService` and `TmdbMetadataService` gain an optional `DateTime? olderThan` parameter; the age filter is applied after the full cross-reference check (all 9 xref types) so orphan detection is not duplicated
- **`PurgeOrphanedTmdbDataJob`** delegates to those service methods, reads the threshold from settings, computes the cutoff date, and returns early when purging is disabled
- Registered as a 24-hour recurring job alongside the other daily maintenance jobs in `SystemService`

## Notes

The `olderThan` filter uses an `is not { LastUpdatedAt: var t }` pattern — a null return from the repository (orphan ID with no local row) is treated as always eligible for purging, which is the correct fallback.

### DIFF
--- a/Shoko.Server/Services/SystemService.cs
+++ b/Shoko.Server/Services/SystemService.cs
@@ -336,7 +336,7 @@ public class SystemService : ISystemService
             if (!FileSystem.AlternateDataStreamExists(dllFile, "Zone.Identifier"))
                 continue;
 
-            _logger.LogError("Found blocked DLL file: " + dllFile);
+            _logger.LogError("Found blocked DLL file: {DllFile}", dllFile);
             result = false;
         }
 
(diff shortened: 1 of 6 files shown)


Shoko.Server/Services/SystemService.cs
```diff
@@ -336,7 +336,7 @@ private bool CheckBlockedFiles()
             if (!FileSystem.AlternateDataStreamExists(dllFile, "Zone.Identifier"))
                 continue;
 
-            _logger.LogError("Found blocked DLL file: " + dllFile);
+            _logger.LogError("Found blocked DLL file: {DllFile}", dllFile);
```
revam, line 339: For the future, try not to do unrelated edits when submitting PRs. @Foowy

revam (approved):

PR #1372, opened 2026-06-26
fix(tmdb): improve episode title matching for non-English content

Foowy:
## Summary

**Non-English title matching:**
- Expands the TMDB episode title candidates from `en-US` only to `en-US` plus the show's `OriginalLanguageCode` (already stored on `TMDB_Show`), so Japanese-only TMDB episodes can match by title instead of falling through to date-only matching. Titles in other languages are excluded to prevent false-positive fuzzy matches.
- Template titles ("Episode N") are excluded from title candidates in every language — TMDB stores these placeholders per language, but AniDB only stores them in English so they're already filtered on the AniDB side.
- Adds a language fallback for the AniDB episode title lookup: when no English title exists, falls back to the anime's primary/official language (Romaji→Japanese, Pinyin→ChineseSimplified, etc.), rather than skipping title matching entirely.

**Episode matching accuracy fixes** (merged in from #1377, found via a library-wide audit of existing "kinda match" cross-references):
- **Episode order reconciliation**: when two AniDB-adjacent episodes both matched on weak evidence alone (date-only or last-resort positional), a coincidental TMDB air-date mismatch could swap them. `ReconcileEpisodeOrderInversions` detects and corrects these after matching, iterating to convergence for runs of 3+ reversed episodes, without disturbing true adjacency between trusted (title-corroborated) matches.
- **Generic placeholder titles**: AniDB's auto-generated placeholder titles ("Episode 5", "Episode S1", "Episode C2", ...) are now excluded from title search for every episode type, not just normal episodes. If both sides only have a placeholder title, the episode isn't left unmatched — title search is simply skipped and the existing air-date and positional-fallback tiers (`TryAirDateMatch` / `FirstAvailable`) still apply, same as before this change.
- **Punctuation normalization**: `~` and `|` are normalized the same as the wave dash and `/`, so decorative/joining-character differences between AniDB and TMDB titles don't silently downgrade an otherwise-identical title to a lower match confidence tier.
- **Greedy candidate starvation**: `TryFindAnidbAndTmdbMatch` now also reports a confidence score (title distance plus air-date corroboration), and each matching pass orders its episodes by that score before matching so the strongest match gets first pick of a shared TMDB episode instead of whichever AniDB episode happened to be processed first. `RankByConfidence` computes this ranking once per pass and reuses the cached match in the pass loop when its chosen TMDB candidate hasn't since been claimed by an earlier (stronger) episode, instead of running the full title-search/air-date match twice per episode.

## Examples

**A Japanese show (OriginalLanguageCode = `ja`) with 5 TMDB titles per episode:**

| Language | Title | Used as candidate? |
|---|---|---|
| en-US | "The Lost Village" | ✅ |
| ja | "失われた村" | ✅ |
| zh-CN | "失去的村子" | ❌ filtered out |
| ko | "잃어버린 마을" | ❌ filtered out |
| fr | "Le village perdu" | ❌ filtered out |

**A Japanese show, untitled episode 3 — TMDB stores placeholders in all languages:**

| Language | Title | Used as candidate? |
|---|---|---|
| en-US | "Episode 3" | ❌ excluded as template |
| ja | "Episode 3" | ❌ excluded as template |

Matching still succeeds here via air date or positional fallback — the change only removes the false confidence of two unrelated placeholder strings matching each other.

## Test plan
- [x] Full test suite in `Shoko.Tests` passes (429 tests), including `TmdbLinkingServiceTests`, `SeriesSearchNormalizationTest`, `TmdbSearchServiceTests`
- [x] Manually traced `ReconcileEpisodeOrderInversions` false-adjacency handling and `originalLanguageCode` threading through the confidence-ranked pass helpers
- [x] Multi-agent code review pass across correctness, reuse, altitude, simplification, and efficiency angles; findings addressed above

## Rebase history
- 2026-07-04 14:45 UTC — rebased onto `master` @ `6dd36ed5e` (2 commits), now at `596656686`


Foowy:
This PR is to catch one off's but does have the impact of bringing is fully native information and with non-english information. Executive decision will need to be made on supporting shows with no current translations in TMDB or not.

Shoko.Server/Providers/TMDB/TmdbLinkingService.cs
```diff
@@ -804,9 +804,21 @@ public IReadOnlyList<ITmdbEpisodeCrossReference> MatchAnidbToTmdbEpisodes(int an
...
+            TitleLanguage.ThaiTranscription => TitleLanguage.Thai,
+            _ => mainTitle?.Language ?? TitleLanguage.English,
+        };
         var anidbTitle = _anidbEpisodeTitles.GetByEpisodeIDAndLanguage(anidbEpisode.EpisodeID, TitleLanguage.English)
-            .Where(title => !title.Title.Trim().Equals($"Episode {anidbEpisode.EpisodeNumber}", StringComparison.InvariantCultureIgnoreCase))
-            .FirstOrDefault()?.Title;
+            .FirstOrDefault(title => !title.Title.Trim().Equals($"Episode {anidbEpisode.EpisodeNumber}", StringComparison.InvariantCultureIgnoreCase))?.Title
+            ?? (fallbackLanguage != TitleLanguage.English
+                ? _anidbEpisodeTitles.GetByEpisodeIDAndLanguage(anidbEpisode.EpisodeID, fallbackLanguage)
+                    .FirstOrDefault(title => !title.Title.Trim().Equals($"Episode {anidbEpisode.EpisodeNumber}", StringComparison.InvariantCultureIgnoreCase))?.Title
```
revam: It's not going to be named Episode X in the origin language. And also, AniDB only stores "Episode X" titles in English, as far as my understanding goes. TMDb on the other hand seem to store it per language for some episodes, if my memory serves me right. Please check.

Shoko.Server/Providers/TMDB/TmdbLinkingService.cs
```diff
@@ -899,6 +911,12 @@ private static double CalculateAirDateProbability(DateOnly? firstDate, DateOnly?
         return 0;
     }
 
+    private static IReadOnlyList<string> GetEpisodeTitleCandidates(TMDB_Episode episode) =>
+        episode.GetAllTitles().Select(t => ReplaceTitle(t.Value))
```
revam: We would need to filter to English and the origin title, and also disallow default title templates in the origin languages.

revam: We already have the original language code stored for the TMDb show, so we can just use that with some filtering. And then use `"en-US"` (`"en"` for language, `"US"` for region/country) for (American) English.

revam (changes requested):
Please check how TMDb does it to be 100% sure we're handling it correctly. And also check the patterns needed to filter out any "default template titles" if it's even necessary in the first place.

Shoko.Server/Providers/TMDB/TmdbLinkingService.cs
```diff
@@ -899,6 +926,15 @@ private static double CalculateAirDateProbability(DateOnly? firstDate, DateOnly?
         return 0;
     }
 
+    private static IReadOnlyList<string> GetEpisodeTitleCandidates(TMDB_Episode episode, string originalLanguageCode) =>
+        episode.GetAllTitles()
+            .Where(t => (t.LanguageCode == "en" && t.CountryCode == "US") || t.LanguageCode == originalLanguageCode)
+            .Where(t => t.LanguageCode == "en" || !t.Value.Trim().Equals($"Episode {episode.EpisodeNumber}", StringComparison.InvariantCultureIgnoreCase))
```
revam: We can drop the `t.LanguageCode == "en" || ` free pass here. we don't want ANY "Episode X" template titles AT ALL for the title matching, in ANY language.

Shoko.Server/Providers/TMDB/TmdbLinkingService.cs
```diff
@@ -829,52 +817,91 @@ private CrossRef_AniDB_TMDB_Episode TryFindAnidbAndTmdbMatch(AniDB_Anime anime,
...
-        var airdateProbability = tmdbEpisodes
-            .Select(episode => new { episode, probability = CalculateAirDateProbability(anidbDate, episode.AiredAt) })
-            .Where(result => result.probability != 0)
-            .OrderByDescending(result => result.probability)
-            .ThenBy(result => result.episode.SeasonNumber == 0)
-            .ThenBy(result => result.episode.SeasonNumber)
-            .ThenBy(result => result.episode.EpisodeNumber)
+        var dateMatchedEpisodes = tmdbEpisodes
+            .Select(episode => (episode, probability: CalculateAirDateProbability(anidbDate, episode.AiredAt)))
+            .Where(r => r.probability > 0)
```
revam: why the `result` → `r` change?

revam (changes requested):
Rest of the changes seems fine, but drop the free pass and revert the unnecessary variable changes in the LINQ filtering.

Shoko.Server/Providers/TMDB/TmdbLinkingService.cs
Foowy: me shorthanding it for laziness, forgot to change back

Foowy:
rebased as of 13:25 02/07/2026

Foowy:
Good for QA/Review/Merge whenever

revam (approved):

revam:
seems fine from skimming the changes, so time to test it in the wilds.

PR #1375, opened 2026-06-29
repo(docker): modernise Dockerfile and build pipeline

Foowy:
## Summary

Replaces the two-stage SDK+runtime Dockerfile with a runtime-only image that expects pre-built artifacts staged into `./publish/<arch>/` by CI. Removes `Dockerfile.aarch64` — a single `Dockerfile` now covers both amd64 and arm64 via BuildKit's `TARGETARCH` arg.

**Why the old approach never cached**

Both old Dockerfiles opened with `COPY . /usr/src/app/source`, copying the entire source tree before NuGet restore. Any commit invalidated every downstream layer. The cache was uploaded after a full build; the next run hit the same invalidation point and rebuilt from scratch anyway.

**New approach**

- `dotnet publish` runs outside Docker in the existing CLI build jobs, with NuGet packages cached at the runner level via `actions/cache`.
- `docker-build` downloads the pre-built `Framework linux-x64` and `Framework linux-arm64` artifacts and stages them under `publish/amd64/` and `publish/arm64/` before calling `docker build`.
- `Dockerfile` `COPY ./publish/${TARGETARCH} .` copies the right slice for each platform in a single `build-push-action` call with `platforms: linux/amd64,linux/arm64`.
- Eliminates the separate `docker-push-manifest` job that previously assembled the manifest list from two per-arch images.

**Root cause of previous failure (PR #1365)**

CI showed green but the runtime container crashed on startup. The `docker-build` job tried to download a `Shoko.CLI_Framework_linux-arm64` artifact that did not exist — the old CLI matrix only included `Framework linux-x64`. With no arm64 artifact, `publish/arm64/` was empty when `docker build` ran; the `COPY` succeeded (empty dir) but the container had no binaries to start. This PR adds `Framework linux-arm64` to the matrix, producing the artifact `docker-build` expects.

**Other changes**

- Replace deprecated `brandedoutcast/publish-nuget` action with direct `dotnet pack` + `dotnet nuget push`, gated on `NUGET_API_KEY` — build is skipped entirely when no key is present.
- Fix `name`/`uses` ordering on all docker action steps (reviewer feedback from #1365).
- Add `--no-install-recommends` and `rm -rf /var/lib/apt/lists/*` to slim the runtime layer.

Also fixed the Dockerfile's "To build locally" doc comment: the documented `/p:Version="0.0.0.0"` placeholder crashed the app on startup with an unhelpful `NullReferenceException` (`PluginManager` force-unwraps a `null` `VersionInformation` when the assembly version is `<= 0.0.0.0`), and `InformationalVersion` was missing the escaped inner quotes MSBuild needs to parse a comma-containing property value.

**What was tested**

- `act push -n` (act v0.2.89) dry-run against `build-daily.yml`, confirming job wiring.
- Full real-world reproduction of the CLI matrix (`Standalone`/`Framework` × `linux-x64`/`linux-arm64`/`osx-arm64`) via dry-run, and live end-to-end verification of the arm64 docker artifact path:
  1. `Shoko.CLI.csproj` `<OutputPath>` → publish output lands in `Shoko.Server/bin/Release/net10.0/linux-arm64/publish/`
  2. Artifact upload `Shoko.CLI_Framework_linux-arm64` uses that exact path
  3. `docker-build` downloads it to `publish/arm64/`
  4. `COPY ./publish/${TARGETARCH} .` with `TARGETARCH=arm64` copies from `publish/arm64/`
- Live `dotnet publish` → `docker build` → `docker run` cycle on amd64: built the image with the real CI publish flags, ran the container, and confirmed `/api/v3/Init/Status` returns `200` within seconds of boot.
- Reproduced and fixed the Dockerfile doc-comment bug above by running the documented snippet verbatim (crashed), then the corrected version (booted and returned `200`).

Stowing this as draft for now

## Rebase history

- 2026-07-13: Rebased onto `master` (`d703dcf01`) to pick up unrelated upstream commits; no conflicts.
- 2026-07-18: Removed the post-build `docker-smoke-test` job from `build-daily.yml`/`build-release.yml`/`docker-manual.yml` (previously added in this PR) — see maintainer discussion for the reasoning; the manual verification above (live `docker build`/`docker run`/`Init/Status` check) still stands as evidence the pipeline produces a booting image.
- 2026-07-18: Dropped `retention-days: 1` on the daily CLI build artifacts — reverted to the default retention per revam's feedback wanting to pull a daily CLI build some time after it ran, not just within the first day.
- 2026-07-22: Rebased onto `master` to pick up unrelated upstream commits (`fix(api): only set image cache headers on successful responses`, `docs(api): correct DELETE File/{fileID} parameter documentation`); no conflicts, force-pushed.
- 2026-07-27 02:48 UTC — rebased onto `master` @ `a9126a417` (3 commits), now at `9227198e2`

- 2026-07-28 09:59 UTC — rebased onto `master` @ `fb1f411a6` (3 commits), now at `1f04e0033`

- 2026-07-29 14:53 UTC — rebased onto `master` @ `342cb8d88` (3 commits), now at `a1fcbeae4`
- 2026-08-03 03:02 UTC — rebased onto `master` @ `259dc7ed0` (3 commits), now at `abead2df8`


revam:
> **Post-build smoke test (new)**

We don't need a "smoke test" speed/build penalty _**for every build**_ if it's done right during the refactor, as that **will** slow down the release pipeline. it's also unfair to the other release types (windows/linux/mac cli/tray-service, windows installer) if we don't add a smoke test to test them. i think it's fine to rely on feedback from our "daily" users if the docker or other release types break, instead of slowing down the release pipeline. others can feel free to comment on this matter if they want.

\*_summons @da3dsoul from the depths of release management to give their take if possible_\*

.github/workflows/build-daily.yml
```diff
@@ -365,6 +372,7 @@ jobs:
         with:
           name: Shoko.CLI_${{ matrix.build_type }}_${{ matrix.rid }}
           path: Shoko.Server/bin/Release/net10.0/${{matrix.rid}}/publish/
+          retention-days: 1
```
revam: We want to be able to get the cli ver from the build a while later. just leave it on auto, no need to set it to 1 day.

Foowy: Ok

Foowy:
Closing — not mergeable, superseded/abandoned.

PR #1377, opened 2026-07-02
fix(tmdb): correct AniDB↔TMDB episode cross-reference accuracy bugs

Foowy:
## Summary
Fixes several root causes behind low-confidence (`*KindaMatches`) and outright-wrong AniDB↔TMDB episode cross-references, found via a library-wide audit of existing matches:

- **Episode order reconciliation**: when two AniDB-adjacent episodes both matched on weak evidence alone (date-only or last-resort positional), a coincidental TMDB air-date mismatch could swap them. `ReconcileEpisodeOrderInversions` now detects and corrects these after matching, iterating to convergence for runs of 3+ reversed episodes. A follow-up commit fixes a false-adjacency bug in that same method: it originally treated two weak matches as neighbors even when a trusted (title-corroborated) match sat between them in true AniDB order, which could produce an incorrect swap.
- **Generic placeholder titles**: AniDB's auto-generated placeholder titles ("Episode 5", "Episode S1", "Episode C2", ...) were only excluded from title search for normal episodes; specials/credits/trailers/etc. with no real title could leak through and coincidentally match an unrelated TMDB episode with zero real evidence.
- **Punctuation normalization**: `~` and `|` are common decorative/joining characters in episode titles (e.g. TMDB's `~Walpurgis~` vs AniDB's `: Walpurgis`, or TMDB's `A | B` vs AniDB's `A / B`) that weren't normalized the same as the wave dash and `/`, silently downgrading otherwise-identical titles to a lower match confidence tier.
- **Greedy candidate starvation**: each matching pass processed AniDB episodes in episode-number order and removed a TMDB episode from the shared candidate pool as soon as any episode claimed it, so when two AniDB episodes both matched the same TMDB episode within a pass, whichever came first won regardless of match strength. `TryFindAnidbAndTmdbMatch` now also reports a confidence score (title distance plus air-date corroboration), and each pass orders its episodes by that score before matching so the strongest match gets first pick.

## Test plan
- [x] Full relevant unit test suite passes in `SeriesSearchNormalizationTest` + `TmdbLinkingServiceTests`, including new regression tests for the false-adjacency fix and show-title normalization coverage
- [x] Traced call sites and mutation safety for `ReconcileEpisodeOrderInversions` and the `NormalizeForIndex` blast radius (whole-show auto-linking) by hand
- [x] Confirmed no external callers reference any of the extracted/renamed private helper methods

### DIFF
--- a/Shoko.Server/Providers/TMDB/TmdbLinkingService.cs
+++ b/Shoko.Server/Providers/TMDB/TmdbLinkingService.cs
@@ -764,6 +764,8 @@ public class TmdbLinkingService : ITmdbLinkingService
             }
         }
 
+        ReconcileEpisodeOrderInversions(anidbEpisodes, tmdbEpisodeDict, toAdd);
+
         if (!saveToDatabase)
         {
             _logger.LogDebug(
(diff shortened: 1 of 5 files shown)


Foowy:
moving back to draft, merge conflicts would exist with PR https://github.com/ShokoAnime/ShokoServer/pull/1372 , will wait for merge on prior and then refactor this change

Foowy:
Closing in favor of #1372. This branch's commits (episode order reconciliation, generic placeholder title filtering, `~`/`|` punctuation normalization, confidence-ranked matching passes) have been merged directly onto `tmdb/episode-title-matching`, combined with #1372's non-English title matching improvements. #1372 will go through full review with both sets of changes included.

PR #1383, opened 2026-07-12
feat(actions): implement plugin-registrable action system

revam:
Implemented the plugin-registrable actions system as specified by the RFC
discussion [#1380](https://github.com/ShokoAnime/ShokoServer/discussions/1380), allowing plugins (and core) to register discrete, invokable
units of work that appear in the Actions menu and can be triggered by ID via the
API.

**Abstractions layer** (Shoko.Abstractions.Action):
- Added `IExecutableAction` base interface with `Name` (nullable; inferred from
  class name split on PascalCase if not set), `Description`, `Category` (hard
  enum `ActionCategory`, defaults to `Mischievous`), and `RequiresConfirmation`
  properties.
- Added scope sub-interfaces distinguishing system-level (admin) from user-level
  variants: `IExecutableGlobalSystemAction`, `IExecutableGlobalUserAction`,
  `IExecutableGroupSystemAction`, `IExecutableGroupUserAction`,
  `IExecutableSeriesSystemAction`, `IExecutableSeriesUserAction`,
  `IExecutableEpisodeSystemAction`, `IExecutableEpisodeUserAction`.
- Added `ExecutableActionInfo` DTO with stable UUIDv5 derived from class name +
  plugin ID namespace.
- Added `IActionService` for registration (`AddParts`), discovery
  (`GetActions`), direct execution, and queued scheduling.
- Expanded `ActionCategory` enum with explicit hex values (e.g.
  `Maintenance = 0xF1`, `Mischievous = 0xF2`, `Destructive = 0xFE`,
  `PluginInferred = 0xFF`) and doc comments on all members.
- Added `ActionScope` as a `[Flags]` enum with `System`, `User`, `Global`,
  `Group`, `Series`, `Episode` base flags and combined values (`SystemAndGlobal`,
  `UserAndGlobal`, `SystemAndGroup`, `UserAndGroup`, `SystemAndSeries`,
  `UserAndSeries`, `SystemAndEpisode`, `UserAndEpisode`).
- Added design overview docs for the abstractions:
  `ActionScope.md`, `ActionCategory.md`, and `IExecutableAction.md`
  (plugin author guide).

**Registration & DI**:
- `PluginManager.InitPlugins()` calls `IActionService.AddParts()` via
  `GetExports<IExecutableAction>`, same pattern as other plugin extension
  points.
- Action instances are transient via `ActivatorUtilities`; lifetime management
  is the plugin's responsibility.
- Stores action types, not cached instances; fresh instance created from type on
  each execution via `IPluginManager.GetExport<T>`.

**Service layer** (`ActionService`):
- Implements `IActionService` with `_registeredActions` list.
- `AddParts` resolves plugin info, derives UUIDs, collects all matching scopes
  via `typeof(T).IsAssignableFrom`, and propagates `RequiresConfirmation` -
  `GetActions` accepts `IEnumerable<ActionScope>?`,
  `IEnumerable<ActionCategory>?`, `IEnumerable<string>?` for set-based filtering
  (scope filtering via `HasFlag`-based overlap, category/category-name contains).
- Result ordering: category value (ensures deterministic core category
  ordering), then category name (ensures deterministic plugin ordering), then
  name (ensures deterministic action ordering).
- `Execute*` methods resolve a fresh instance each time, validate scope presence
  upfront, throw `InvalidOperationException` on mismatch - `ScheduleExecuteOf*`
  validates scope before enqueuing via `ExecuteActionJob`.
- Added 7 new methods migrated from the API controller: `RunImport_SyncVotes`,
  `PurgeAllTmdbLinks`, `UpdateAniDBCalendar`, `PlexSyncAll`,
  `AddAllManualLinksToMyList`, `GetAniDBNotifications`, `AVDumpMismatchedFiles`.

**Global action implementations** (Shoko.Server.Services.Actions):
- 35 `IExecutableGlobalSystemAction` classes across 5 domains:
  - Import (3): RunImport, ImportNewFiles, RemoveMissingFiles
  - AniDB (11): DownloadMissingAniDBAnimeData, Creators,
    UpdateAllAniDBInfo, UpdateMissingAniDBFileInfo,
    RefreshAniDBMovedFiles, VerifyAllRelations, SyncAniDBMyList,
    SyncAniDBVotes, UpdateAniDBCalendar,
    AddAllManualLinksToMyList, GetAniDBNotifications,
    AVDumpMismatchedFiles
  - TMDB (10): SearchForMatches, UpdateAllTMDBMovies (+WithImages),
    UpdateAllTMDBShows (+WithImages), DownloadMissingPeople,
    PurgeAllUnusedMovies, Collections, Shows,
    ShowAlternateOrderings, PurgeAllLinks
  - Images (3): UpdateAllImages, ValidateAllImages,
    PurgeAllUnusedTMDBImages
  - Maintenance (8): UpdateAllMediaInfo, UpdateSeriesStats,
    RecreateAllGroups, RenameAllGroups, CreateMissingSeries,
    PurgeAllUsedReleases, PurgeAllUnusedReleases, PlexSyncAll
- Destructive actions set `RequiresConfirmation = true`
- `RunImportAction` and `SyncAnidbMyListAction` use `IJobFactory`

**Series action implementations** (Shoko.Server.Services.Actions.*.Series):
- 17 `IExecutableSeriesSystemAction` classes across 4 sub-namespaces:
  - AniDB.Series (4): UpdateAnidbInfo (cache+remote fallback),
    UpdateAnidbInfoRemote (remote, respects checks),
    UpdateAnidbInfoForce (bypasses time check + HTTP bans,
    requires confirmation), UpdateAnidbInfoXmlCache (local XML only)
  - Tmdb.Series (7): AutoSearchTmdbMatch, UpdateTmdbInfo,
    UpdateTmdbImagesForce, RefreshTmdbMovies,
    DownloadTmdbMovieImages, AutoMatchTmdbEpisodes,
    ResetTmdbEpisodeMappings (requires confirmation)
  - Import.Series (3): RescanFiles, RehashFiles, RelocateFiles
  - Destructive.Series (3): DeleteKeepFiles, DeleteRemoveFiles,
    DeleteAllData (all require confirmation)
- Actions are self-contained: they inject services directly and
  operate on the `IShokoSeries` passed at invocation

**Group action implementations** (Shoko.Server.Services.Actions.Import.Group):
- 3 `IExecutableGroupSystemAction` classes mirroring the series file actions:
  RescanGroupFiles, RehashGroupFiles, RelocateGroupFiles
- Each iterates all series within the group to operate on their files

**Queue job** (`ExecuteActionJob`):
- Stores `ActionID`, `UserID`, `GroupID`, `AnimeID`, and `EpisodeID`
- Uses `PostInit` to hydrate display names via `IActionService`,
  `IUserService`, and `IMetadataService`
- `ScopeToName` infers the scope label from which of
  (GroupID, AnimeID, EpisodeID, UserID) are set,
  covering all 8 scope combinations
- Routes to the appropriate `IActionService.Execute*` method

**API layer** (v3):
- `GET /api/v3/Action/Actions` — lists global/user-scoped actions grouped
  by `CategoryName` (B2 shape), filtered by user permissions
- `POST /api/v3/Action/{actionID}/Execute?scope=` — invokes a
  global/user action; scope defaults by role priority; system-scoped
  actions require admin, user-scoped variants require any auth
- `GET /api/v3/Action/Group/Actions` — lists group-scoped actions
- `POST /api/v3/Action/Group/{groupID}/Action/{actionID}/Execute?scope=`
  — invokes a group/group-user action with group existence validation
- `GET /api/v3/Action/Series/Actions` — lists series-scoped actions
- `POST /api/v3/Action/Series/{seriesID}/Action/{actionID}/Execute?scope=`
  — invokes a series/series-user action with series existence
  validation
- `GET /api/v3/Action/Episode/Actions` — lists episode-scoped
  actions (hookup for future)
- `POST /api/v3/Action/Episode/{episodeID}/Action/{actionID}/Execute?scope=`
  — invokes an episode/episode-user action with episode existence
  validation
- Scope-aware auth: resolvedScope.HasFlag(ActionScope.System)
  gates admin actions uniformly across all entity levels
- Plugin info included in `ActionInfo` response DTO
- Old hardcoded endpoints marked `[Obsolete]`; implementation
  delegated to `ActionService`

**Differences from RFC #1380**:

**Interface hierarchy (scope inference via sub-interfaces):**
The RFC specifies a single `IExecutableAction` with a `Scope` property and a parameterless `Execute(CancellationToken)`. It does not address how entity context (the series, group, or episode to operate on) or the invoking user are actually passed into `Execute` — Gap 18 dismisses it as "an implementation detail."

This implementation instead defines 8 scope-specific sub-interfaces, each with an `Execute` signature that provides exactly the context the action needs:

- `IExecutableGlobalSystemAction` → `Execute(CancellationToken)` — no context
- `IExecutableGlobalUserAction` → `Execute(IUser, CancellationToken)` — user context
- `IExecutableSeriesSystemAction` → `Execute(IShokoSeries, CancellationToken)` — series context
- `IExecutableSeriesUserAction` → `Execute(IShokoSeries, IUser, CancellationToken)` — both
- (and so on for Group and Episode)

An action declares which scopes it supports by implementing the corresponding sub-interface(s). `ActionService.AddParts` detects every implemented interface via `typeof(T).IsAssignableFrom` and registers each as a supported scope on `ExecutableActionInfo.Scopes`. A single class can support multiple scopes (e.g. both `IExecutableGlobalSystemAction` and `IExecutableSeriesSystemAction`) by implementing both interfaces.

This design was chosen over the RFC's single-interface approach because:
- **DX and intent**: the interface an action implements is a self-documenting declaration of what it needs. No guesswork about whether `Execute` will receive a series, a user, or neither.
- **The RFC doesn't specify how entity/user context reaches Execute.** A single parameterless method with only `CancellationToken` provides no mechanism for the service to supply the series, group, or user the action should operate on. The sub-interface approach makes this explicit at the type level.
- **The scope set on `ExecutableActionInfo` is inferred from the interfaces, not declared as a property.** This eliminates a class of bugs where a plugin sets `Scope` to `Series` but forgets to handle the series parameter. Bare `IExecutableAction` without any sub-interface is not registered — at least one sub-interface must be implemented.

**Permission modeling (folded into ActionScope):**
The RFC's revised Gap 8 specifies a separate `ActionPermission` enum (`Admin`/`User`) orthogonal to `Scope`. This implementation folds permission into `ActionScope` itself via `[Flags]`:

```csharp
[Flags]
public enum ActionScope : byte
{
    System = 1 << 0,    // admin gate
    User   = 1 << 1,    // any authenticated user
    Global = 1 << 2,    // scope-level flags
    Group  = 1 << 3,
    Series = 1 << 4,
    Episode = 1 << 5,
}
```

Combined values like `SystemAndGlobal` describe both the entity scope and the permission level in a single value. The admin check at the API layer is a single uniform `resolvedScope.HasFlag(ActionScope.System)` regardless of which entity level the action targets — no per-entity-level conditionals needed.

Keeping base flags and entity-level flags on the same enum was a deliberate choice to simplify the `ExecutableActionInfo.Scopes` set: a multi-scope action (e.g. implementing both `IExecutableGlobalSystemAction` and `IExecutableGlobalUserAction`) stores `{ SystemAndGlobal, UserAndGlobal }` — both values carry the entity scope (`Global`) alongside the permission level (`System` or `User`), so a listing endpoint or plugin can filter by either axis with a single `HasFlag` call without joining across two enums.

This also extends the RFC's original `Global`/`Series` scope pair with `Group` and `Episode` scopes to match four of the entity levels Shoko supports. A `Video`-level scope was left out for now but can be added later if deemed necessary.

**Other differences:**
- `GetActions` uses set-based overlap filtering instead of single scope matching
- `CategoryName` string on `ExecutableActionInfo` carries the display label; `PluginInferred` category uses the plugin's name as the label rather than a core-owned `Generic` fallback (RFC Gap 1)
- `RequiresConfirmation` added as a UI hint for destructive actions (not specified in RFC but a natural follow-on)
- Direct execution (`Execute*` methods) exists on the service; the API always schedules through the job queue system, but plugins may use `Execute*` for immediate in-process execution
- Actions store types, not instances; fresh instance per invocation
- `ActionCategory` default changed back from `PluginInferred` to `Mischievous`, realigning with the original RFC's intent for a fixed core-owned fallback
- `Name` is nullable and inferred from the class name when not set (the RFC treats it as required)

Refs: [#1380](https://github.com/ShokoAnime/ShokoServer/discussions/1380)

### DIFF
--- a/Shoko.Abstractions/Action/ActionCategory.md
+++ b/Shoko.Abstractions/Action/ActionCategory.md
@@ -1,0 +1,51 @@
+# ActionCategory — Design Overview
+
+`ActionCategory` is a hard-validated enum (`: byte`) with fixed members.
+Plugins cannot register new categories at runtime — adding a genuinely new
+category requires a PR against core to extend the enum.
+
+## Members
+
+| Member | Value | Description |
+|--------|-------|-------------|
+| `Import` | `0x01` | File import and scanning actions. |
+| `AniDB` | `0x21` | AniDB metadata and synchronization actions. |
+| `TMDB` | `0x22` | TMDB metadata and synchronization actions. |
+| `AniList` | `0x23` | AniList metadata and synchronization actions. |
+| `Sync` | `0x31` | Data synchronization actions across providers. |
+| `Images` | `0x71` | Image download and management actions. |
+| `Maintenance` | `0xF1` | System maintenance actions. |
+| `Mischievous` | `0xF2` | Uncategorized or miscellaneous actions. |
+| `Destructive` | `0xFE` | Destructive operations such as purging data. |
+| `PluginInferred` | `0xFF` | Category label is inferred from the owning plugin's name. |
+
+## Value spacing
+
+The hex values are intentionally spaced into ranges to allow future additions
+without renumbering:
+
+- `0x01–0x3F` — provider-specific categories (Import, AniDB, TMDB, AniList, Sync)
+- `0x71–0x7F` — system categories (Images)
+- `0xF1–0xFF` — special categories (Maintenance, Mischievous, Destructive,
+  PluginInferred)
+
+## `Mischievous` vs `PluginInferred`
+
+- **`Mischievous`** is the default fallback. An action that doesn't explicitly
+  set a category lands here. Its label is "Mischievous", owned centrally by
+  core.
+- **`PluginInferred`** is an opt-in choice. When an action uses this category,
+  `ActionService.AddParts` sets `ExecutableActionInfo.CategoryName` to the
+  owning plugin's display name instead of the enum member name.
+
+  This means `PluginInferred` effectively delegates the category label itself to
+  the plugin — distinct from the RFC's original `Generic` proposal, which
+  specified a fixed core-owned label. `Mischievous` exists as the true
+  core-owned fallback that aligns with the original RFC intent.
+
+## Category display names
+
+Each enum member's display name is its own `ToString()` value, except for
+`PluginInferred`, where the display name is the plugin's name. The
+`ExecutableActionInfo.CategoryName` field carries the resolved display label
+for API consumers.
--- a/Shoko.Abstractions/Action/ActionScope.md
+++ b/Shoko.Abstractions/Action/ActionScope.md
@@ -1,0 +1,65 @@
+# ActionScope — Design Overview
+
+`ActionScope` is a `[Flags]` enum that encodes two orthogonal concerns in a
+single value:
+
+1. **Permission level** — whether an action requires admin privileges or is open
+   to any authenticated user.
+2. **Entity scope** — the entity type the action operates on (Global, Group,
+   Series, or Episode).
+
+## Base flags
+
+| Flag | Bit | Meaning |
+|------|-----|---------|
+| `System` | `1 << 0` | Admin-only action. Requires `[Authorize("admin")]`. |
+| `User` | `1 << 1` | Any authenticated user may invoke. |
+| `Global` | `1 << 2` | Not tied to any particular entity. |
+| `Group` | `1 << 3` | Scoped to a group. |
+| `Series` | `1 << 4` | Scoped to a series. |
+| `Episode` | `1 << 5` | Scoped to an episode. |
+
+## Combined (named) values
+
+Every value stored in an action's `ExecutableActionInfo.Scopes` set is always a
+combination of exactly one permission flag (`System` or `User`) plus exactly one
+entity flag (`Global`, `Group`, `Series`, or `Episode`). The named combined
+values exist for client convenience:
+
+| Named value | Permission | Entity |
+|-------------|------------|--------|
+| `SystemAndGlobal` | System | Global |
+| `UserAndGlobal` | User | Global |
+| `SystemAndGroup` | System | Group |
+| `UserAndGroup` | User | Group |
+| `SystemAndSeries` | System | Series |
+| `UserAndSeries` | User | Series |
+| `SystemAndEpisode` | System | Episode |
+| `UserAndEpisode` | User | Episode |
+
+**Important:** bare flags like bare `Global` or bare `System` are never stored in
+an action's scope set. They exist only as filter values for querying.
+
+## Filtering
+
+Because every stored value is a combination, `HasFlag`-based filtering allows
+querying by either axis independently:
+
+- `GetActions(scopes: [ActionScope.System])` — returns all admin actions across
+  all entity levels.
+- `GetActions(scopes: [ActionScope.Global])` — returns all global actions
+  (both admin and user variants).
+- `GetActions(scopes: [ActionScope.SystemAndGroup])` — returns actions that
+  require admin privileges and target a group.
+
+This is the approach used in `ActionService.GetActions` and the listing
+endpoints in the API controller.
+
+## Why a single enum instead of separate Scope + Permission enums?
+
+The RFC's revised Gap 8 specifies a separate `ActionPermission` enum orthogonal
+to `ActionScope`. Keeping both on a single `[Flags]` enum simplifies the
+`ExecutableActionInfo.Scopes` set: a multi-scope action stores values like
+`{ SystemAndGlobal, UserAndGlobal }` — both carry the entity scope (`Global`)
+alongside the permission level. A listing endpoint can filter by either axis
+with a single `HasFlag` call, without joining across two separate enums.
--- a/Shoko.Abstractions/Action/Enums/ActionCategory.cs
+++ b/Shoko.Abstractions/Action/Enums/ActionCategory.cs
@@ -1,0 +1,63 @@
+using System.Text.Json.Serialization;
+using Newtonsoft.Json.Converters;
+
+namespace Shoko.Abstractions.Action.Enums;
+
+/// <summary>
+///   Defines the category of an executable action. Categories group related
+///   actions together in the UI and can be used to filter the action list.
+/// </summary>
+[JsonConverter(typeof(JsonStringEnumConverter))]
+[Newtonsoft.Json.JsonConverter(typeof(StringEnumConverter))]
+public enum ActionCategory : byte
+{
+    /// <summary>
+    ///   File import and scanning actions.
+    /// </summary>
+    Import = 0x01,
+
+    /// <summary>
+    ///   AniDB metadata and synchronization actions.
+    /// </summary>
+    AniDB = 0x21,
+
+    /// <summary>
+    ///   TMDB metadata and synchronization actions.
+    /// </summary>
+    TMDB = 0x22,
+
+    /// <summary>
+    ///   AniList metadata and synchronization actions.
+    /// </summary>
+    AniList = 0x23,
+
+    /// <summary>
+    ///   Data synchronization actions across providers.
+    /// </summary>
+    Sync = 0x31,
+
+    /// <summary>
+    ///   Image download and management actions.
+    /// </summary>
+    Images = 0x71,
+
+    /// <summary>
+    ///   System maintenance actions.
+    /// </summary>
+    Maintenance = 0xF1,
+
+    /// <summary>
+    ///   Uncategorized or miscellaneous actions.
+    /// </summary>
+    Mischievous =0xF2,
+
+    /// <summary>
+    ///   Destructive operations such as purging data.
+    /// </summary>
+    Destructive = 0xFE,
+
+    /// <summary>
+    ///   The category should be inferred from the plugin that provides the action.
+    /// </summary>
+    PluginInferred = 0xFF,
+}
--- a/Shoko.Abstractions/Action/Enums/ActionCategory.cs
+++ b/Shoko.Abstractions/Action/Enums/ActionCategory.cs
@@ -47,9 +47,10 @@ public enum ActionCategory : byte
     Maintenance = 0xF1,
 
     /// <summary>
-    ///   Uncategorized or miscellaneous actions.
+    ///   Uncategorized or miscellaneous actions. Actions which does not set
+    ///   their own category will be placed in this category by default.
     /// </summary>
-    Mischievous =0xF2,
+    Mischievous = 0xF2,
 
     /// <summary>
     ///   Destructive operations such as purging data.
--- a/Shoko.Abstractions/Action/Enums/ActionScope.cs
+++ b/Shoko.Abstractions/Action/Enums/ActionScope.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Text.Json.Serialization;
 using Newtonsoft.Json.Converters;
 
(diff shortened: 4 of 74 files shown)


hidden4003:
1. Scope enum ballooned: spec resolved ActionScope as exactly Global | Series (Gap 20, brand new in v3). The PR ships 8 flat values — Global/GlobalUser/Group/GroupUser/Series/SeriesUser/Episode/EpisodeUser — introducing Group and Episode scoping and "User" variants of every level. None of that exists in the spec at all. This is the biggest scope expansion in the PR.

@revam  Why do we need that many?

2. Permissions model contradicts Gap 8/19 as written: the RFC was explicit and re-confirmed twice — "all actions require admin privileges... no per-action role granularity," and for series actions specifically, "no basis for opening them to non-admin users" since they act on shared objects. The PR's ...User scope variants are explicitly non-admin ("any auth for user-scoped variants"). That's a direct reversal of a gap the RFC author and maintainer both signed off on, not just an extension — this is the one I'd push back hardest on in review, since it wasn't discussed anywhere in the thread.

Spec needs to be changed if that is really the case.

3. Interface shape changed from a Scope property to a family of marker interfaces: the RFC's interface had ActionScope Scope { get; } directly on IExecutableAction. The PR instead uses IExecutableGlobalAction, IExecutableSeriesAction, etc., and scope is inferred via typeof(T).IsAssignableFrom. This isn't called out in the "Differences" section even though it's a real structural departure — functionally it achieves the same routing goal, but a plugin author reading the RFC's published interface would write code that doesn't compile against this PR.

Either code or spec needs to be changed.

4. ID derivation adds plugin ID as namespace: spec said UUIDv5 from "the action's fully-qualified class name" only. PR derives from class name plus plugin ID namespace. Arguably better (avoids cross-plugin collisions the RFC worried about elsewhere), but again, not what Gap 4/5 says.

looks like a good change spec needs to be updated.

revam:
> 1. Scope enum ballooned: spec resolved ActionScope as exactly Global | Series (Gap 20, brand new in v3). The PR ships 8 flat values — Global/GlobalUser/Group/GroupUser/Series/SeriesUser/Episode/EpisodeUser — introducing Group and Episode scoping and "User" variants of every level. None of that exists in the spec at all. This is the biggest scope expansion in the PR.
> 
> @revam Why do we need that many?
> 

It was two variants (admin, user) per scope level, across four levels (global  group, series, episode). each scope corresponds to a a sub interface with a caller signature for that scope.

> 2. Permissions model contradicts Gap 8/19 as written: the RFC was explicit and re-confirmed twice — "all actions require admin privileges... no per-action role granularity," and for series actions specifically, "no basis for opening them to non-admin users" since they act on shared objects. The PR's ...User scope variants are explicitly non-admin ("any auth for user-scoped variants"). That's a direct reversal of a gap the RFC author and maintainer both signed off on, not just an extension — this is the one I'd push back hardest on in review, since it wasn't discussed anywhere in the thread.
> 
> Spec needs to be changed if that is really the case.
> 

User actions did come up in the thread, but more from a misunderstanding then something else. But thinking about it, I decided that some things would make sense to have as actions per user, though the examples I can come up with on the spot mostly resolves around user data, e.g. per user plex sync or MyList sync. But as user actions would not require admin perms, as they're not admin commands. the api controller should enforce that admin actions can only be ran by admins. the description for admin vs. any user on each scope is there for plugin authors to see who should have access to the action scopes, if a plugin were to run an action on behalf of a user.

> 3. Interface shape changed from a Scope property to a family of marker interfaces: the RFC's interface had ActionScope Scope { get; } directly on IExecutableAction. The PR instead uses IExecutableGlobalAction, IExecutableSeriesAction, etc., and scope is inferred via typeof(T).IsAssignableFrom. This isn't called out in the "Differences" section even though it's a real structural departure — functionally it achieves the same routing goal, but a plugin author reading the RFC's published interface would write code that doesn't compile against this PR.
> 
> Either code or spec needs to be changed.

The scope is kinda still in the impl., but the way Execute was called on the RFC version allowed for no way to supply the series for a series scoped action. so instead i opted for sub interfaces for registration, and scopes on the action infos for the service and api consumers.

> 4. ID derivation adds plugin ID as namespace: spec said UUIDv5 from "the action's fully-qualified class name" only. PR derives from class name plus plugin ID namespace. Arguably better (avoids cross-plugin collisions the RFC worried about elsewhere), but again, not what Gap 4/5 says.
> 
> looks like a good change spec needs to be updated.

all other uuidv5s throughout the plugin system are namespaced by their plugin id, for reference.

hidden4003:
> User actions did come up in the thread, but more from a misunderstanding then something else. But thinking about it, I decided that some things would make sense to have as actions per user, though the examples I can come up with on the spot mostly resolves around user data, e.g. per user plex sync or MyList sync. But as user actions would not require admin perms, as they're not admin commands. the api controller should enforce that admin actions can only be ran by admins. the description for admin vs. any user on each scope is there for plugin authors to see who should have access to the action scopes, if a plugin were to run an action on behalf of a user.

From what I see in the API Actions controller currently there is no checking whether user is admin or not, therefore there is no concept of admin action vs user action in current code. We can of course introduce it but it seems like a bit out of scope for the system since we never really tested multi-user setups and as far as I know we can only have a single AniDB user in the server anyway i.e. multiple shoko users can't use different AniDB user logins as it is currently. 

I feel like the whole concept for scope has to either be scrapped or scope just be a constant - `Admin` for the moment.

Would like to hear from @da3dsoul and @harshithmohan regarding this to make sure my understanding of API code is correct.

da3dsoul:
I agree that user doesn't make sense for now, but if it's not hard to make user-based actions here, it would be nice to not have it be a path that needs to change if/when multi-user is maybe more supported. AniDB is hard to do multi-user, because we are API limited, but the queue can now handle it better without completely ruining the experience, as long as you aren't actively running a big import. Other list providers should be per-user when implemented

harshithmohan:
I feel like you are focussing too much on having multi-user AniDB support. Ignore that for a moment, any other plugin, be it an Anilist Sync or a MAL sync plugin or whatever can take advantage of user actions with the current codebase.

harshithmohan:
**Three smaller RFC deviations around defaults, response shape, and category metadata:**

**1. `Generic` → `PluginInferred`**

The RFC's default was a single "Generic" bucket for all uncategorized actions. The PR replaces it with `PluginInferred`, which groups each plugin's uncategorized actions under the plugin's name. Intentional — but if many plugins ship actions without picking a core category, the UI ends up with a category per plugin, which could get cluttered. The RFC's catch-all avoided that. Not a blocker, just something to keep an eye on as plugin count grows.

**2. Response model additions are fine, but `description` on the category object is missing**

`RequiresConfirmation` and `PluginInfo` on each action are useful additions — no issue there. But the RFC's category group object included a `description` field (e.g. `"description": "Miscellaneous actions"`) that the PR drops entirely. `ActionCategoryGroup` only has `Name` and `Actions`. Without it, the UI can't show a category-level tooltip, and adding one later would mean a response shape change.

**3. No central registry for category names/descriptions — ties into 2.**

The RFC specified category metadata would be "owned centrally by core." The PR derives names from `category.ToString()` with no descriptions anywhere. For v1 this works (enum names are self-explanatory), but it means there's no single place to curate labels or add descriptions later. Adding `description` to `ActionCategoryGroup` and a central lookup for it would resolve both 2. and 3.

revam:
> **Three smaller RFC deviations around defaults, response shape, and category metadata:**
> 
> **1. `Generic` → `PluginInferred`**
> 
> The RFC's default was a single "Generic" bucket for all uncategorized actions. The PR replaces it with `PluginInferred`, which groups each plugin's uncategorized actions under the plugin's name. Intentional — but if many plugins ship actions without picking a core category, the UI ends up with a category per plugin, which could get cluttered. The RFC's catch-all avoided that. Not a blocker, just something to keep an eye on as plugin count grows.
> 
> **2. Response model additions are fine, but `description` on the category object is missing**
> 
> `RequiresConfirmation` and `PluginInfo` on each action are useful additions — no issue there. But the RFC's category group object included a `description` field (e.g. `"description": "Miscellaneous actions"`) that the PR drops entirely. `ActionCategoryGroup` only has `Name` and `Actions`. Without it, the UI can't show a category-level tooltip, and adding one later would mean a response shape change.
> 
> **3. No central registry for category names/descriptions — ties into 2.**
> 
> The RFC specified category metadata would be "owned centrally by core." The PR derives names from `category.ToString()` with no descriptions anywhere. For v1 this works (enum names are self-explanatory), but it means there's no single place to curate labels or add descriptions later. Adding `description` to `ActionCategoryGroup` and a central lookup for it would resolve both 2. and 3.

@harshithmohan I thought we decided not to add any category description in discord? If we still need it, then I can add it, but I'll need the text to add for each category except `PluginInferred` if the current ones in the enum definition are not satisfying enough. I'll also note that the current UI impl. doesn't have any, and idk where we would add it to the current UI (a tooltip does **not** sound like a _good_ idea, for reference).

hidden4003:
**Execute signature + permission modeling — need your input**

The current impl (PR #1383, latest revision) diverges from RFC Gap 18/20 more than the spec currently documents: `Execute` isn't parameterless — it's 8 separate signatures (`Global/Group/Series/Episode` × `System/User`), and permission is folded into `ActionScope` itself as a `[Flags]` combo rather than a separate property.

Two questions, really one:

**1. Should `Execute` stay split across scope-specific interfaces, or collapse to one per entity with a nullable `IUser?`?**

- **A — status quo (8 interfaces):** `Execute()`, `Execute(IUser)`, `Execute(IShokoSeries)`, `Execute(IShokoSeries, IUser)`, etc. Compile-time guarantee that a system action can't touch a user it was never given.
- **B — 4 interfaces, `IUser?` always passed:** `Execute(IUser? user, ...)`, `Execute(IShokoSeries series, IUser? user, ...)`. Null when system/scheduled-invoked, non-null for user-tier calls. Fewer interfaces, but pushes the "did I handle null correctly" check to runtime instead of the type system.

**2. Should `Permission` (admin vs. user) be its own property, independent of `Scope`/entity?**

Right now it's baked into `ActionScope` (`SystemAndSeries`, `UserAndSeries`, etc.) — my preference is to pull it back out as a standalone `Permission` field so scope and permission stay orthogonal, but that's a bigger ask since it touches what's already merged. Wanted your take before pushing for it.

B and "pull Permission out" aren't the same axis — you could do A+separate-Permission, B+separate-Permission, or leave both as-is. Curious which combination you'd actually want to maintain going forward.

harshithmohan:
> > **Three smaller RFC deviations around defaults, response shape, and category metadata:**
> > **1. `Generic` → `PluginInferred`**
> > The RFC's default was a single "Generic" bucket for all uncategorized actions. The PR replaces it with `PluginInferred`, which groups each plugin's uncategorized actions under the plugin's name. Intentional — but if many plugins ship actions without picking a core category, the UI ends up with a category per plugin, which could get cluttered. The RFC's catch-all avoided that. Not a blocker, just something to keep an eye on as plugin count grows.
> > **2. Response model additions are fine, but `description` on the category object is missing**
> > `RequiresConfirmation` and `PluginInfo` on each action are useful additions — no issue there. But the RFC's category group object included a `description` field (e.g. `"description": "Miscellaneous actions"`) that the PR drops entirely. `ActionCategoryGroup` only has `Name` and `Actions`. Without it, the UI can't show a category-level tooltip, and adding one later would mean a response shape change.
> > **3. No central registry for category names/descriptions — ties into 2.**
> > The RFC specified category metadata would be "owned centrally by core." The PR derives names from `category.ToString()` with no descriptions anywhere. For v1 this works (enum names are self-explanatory), but it means there's no single place to curate labels or add descriptions later. Adding `description` to `ActionCategoryGroup` and a central lookup for it would resolve both 2. and 3.
> 
> @harshithmohan I thought we decided not to add any category description in discord? If we still need it, then I can add it, but I'll need the text to add for each category except `PluginInferred` if the current ones in the enum definition are not satisfying enough. I'll also note that the current UI impl. doesn't have any, and idk where we would add it to the current UI (a tooltip does **not** sound like a _good_ idea, for reference).

I forgot about that 😔 
Leave it off then

harshithmohan:
~~Why did we move away from sub-interfaces? This one is even worse.
Sub-interfaces made sense for the objective~~
Apparently, I can't read

hidden4003:
> Why did we move away from sub-interfaces? This one is even worse. Sub-interfaces made sense for the objective

We did not move off anything yet it says "— need your input" in the comment so if you don't want to I can just put the interfaces in the spec and it will be reconciled with implementation.

PR #1385, opened 2026-07-13
fix(api): reject path traversal in external subtitle filename

Foowy:
## Summary
- `GetExternalSubtitle` (`FileController`) passed the raw `filename` route parameter straight into `Path.Combine` with no traversal/absolute-path check — an authenticated caller (or anonymous, if `AllowAnonymousFileStreamingInAPIv3` is on) could read arbitrary files on the host via `../` sequences or a rooted path.
- The first pass at a fix simply rejected any `filename` containing `/` or `\` — but that also blocked legitimate subtitles living in a subfolder next to the video, while doing nothing to stop encoded traversal segments embedded elsewhere in the name. Replaced with a regex (`PathTraversalPattern`) that specifically rejects `.`/`..` path segments, plus a whitelist check against the video's actual known external subtitle filenames (`MediaInfo.TextStreams`) — `filename` can now only ever resolve to a subtitle genuinely linked to that file, not just any traversal-free string.
- SonarCloud flagged the traversal regex as an unbounded-ReDoS risk; added a 250ms match timeout, matching the convention already used by `LogService`'s filter regex.
- An earlier revision of this PR also added an `AllowedSeries` (restricted/parental-control profile) check to the file-streaming endpoints, since they were missing it unlike other endpoints in this controller — that was dropped to keep this PR scoped to the traversal fix only; it's a separate, lower-severity gap that can be addressed independently.

Found during a local security review of the file-streaming surface.

## Test plan
- [x] `dotnet build Shoko.Server.sln -c Release` — 0 errors

## Rebase history
- 2026-07-27 02:48 UTC — rebased onto `master` @ `a9126a417` (4 commits), now at `1f7d99355`

### DIFF
--- a/Shoko.Server/API/v3/Controllers/FileController.cs
+++ b/Shoko.Server/API/v3/Controllers/FileController.cs
@@ -84,7 +84,7 @@ public class FileController(
 
     internal const string FileForbiddenForUser = "Accessing File is not allowed for the current user";
 
-    private static readonly Regex PathTraversalPattern = new(@"(^|(?<=[\\/]))\.\.?[\\/]", RegexOptions.Compiled);
+    private static readonly Regex PathTraversalPattern = new(@"(^|(?<=[\\/]))\.\.?[\\/]", RegexOptions.Compiled, TimeSpan.FromMilliseconds(250));
 
     /// <summary>
     /// Get or search through the files accessible to the current user.
--- a/Shoko.Server/API/v3/Controllers/FileController.cs
+++ b/Shoko.Server/API/v3/Controllers/FileController.cs
@@ -688,6 +688,16 @@ public class FileController(
     public ActionResult GetFileStreamWithDirectory([FromRoute, Range(1, int.MaxValue)] int fileID, [FromRoute] string? filename = null, [FromQuery] bool streamPositionScrobbling = false)
         => GetFileStreamInternal(fileID, filename, streamPositionScrobbling);
 
+    /// <summary>
+    /// Checks whether the current user is allowed to view the series the given file is cross-referenced to,
+    /// matching the restriction already enforced on other endpoints via <see cref="JMMUser.AllowedSeries"/>.
+    /// </summary>
+    private bool IsSeriesAllowedForFile(VideoLocal file)
+    {
+        var series = file.EpisodeCrossReferences.FirstOrDefault(xref => xref.AnimeID is not 0)?.AnimeSeries;
+        return series == null || User.AllowedSeries(series);
+    }
+
     [NonAction]
     public ActionResult GetFileStreamInternal(int fileID, string? filename = null, bool streamPositionScrobbling = false)
     {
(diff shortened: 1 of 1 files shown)


Foowy:
Purely security check, most users will not see this, but we might as well not be a weak point

Shoko.Server/API/v3/Controllers/FileController.cs
```diff
@@ -753,6 +753,9 @@ public ActionResult GetExternalSubtitle([FromRoute, Range(1, int.MaxValue)] int
         if (!SettingsProvider.GetSettings().Web.AllowAnonymousFileStreamingInAPIv3 && User is null)
             return Unauthorized();
 
+        if (filename.IndexOfAny(['/', '\\']) >= 0 || filename is "." or "..")
```
revam: '/' and '\\' are percentage encoded when read here when `{filename}` doesn't contain a star at the start of it's capture region. this check will not be helpful. also, if we did allow the characters, then I'd opt for a regex rejecting pattern like `/(^|(?<=[\\\/]))\.\.?[\\\/]/` instead of only checking if the filename is exactly '.' or '..' or if it contains a directory separator.

Shoko.Server/API/v3/Controllers/FileController.cs
```diff
@@ -761,8 +764,7 @@ public ActionResult GetExternalSubtitle([FromRoute, Range(1, int.MaxValue)] int
         {
             var path = place.FileInfo?.Directory?.FullName;
             if (path == null) continue;
-            path = Path.Combine(path, filename);
-            var subFile = new FileInfo(path);
+            var subFile = new FileInfo(Path.Combine(path, filename));
```
revam, line 774: also no need for this change, as they're functionally the same.

revam (changes requested):

Shoko.Server/API/v3/Controllers/FileController.cs
revam: maybe also validate the file name against the external subtitle files stored in the media info object for the video, to make sure we're not allowing it to read any other files other than the approved external subtitles linked to the video.

revam: and the reason i would not reject on a directory separator is in case the subtitle file is in a sub-folder on disc.

Foowy: forgot about that, so used to subtitle being in same base directory of adjacent video

Shoko.Server/API/v3/Controllers/FileController.cs
Foowy: that was code tidying, one less line, my personal preferences sneaking in.

revam: the change can stay, but it also wasn't necessary for the fix — which is why i sent my earlier comment.

Foowy:
Used a 250ms match timeout for `PathTraversalPattern`, matching the timing already used by `LogService`'s `_logFilterRegexMatchTimeout`.

revam (approved):

PR #1387, opened 2026-07-15
feat(api): example OIDC single sign-on implementation

Foowy:
## Summary

Example implementation of optional OIDC single sign-on, offered as a discussion starting point — not opinion-locked. See [ShokoAnime/ShokoServer#1386](https://github.com/ShokoAnime/ShokoServer/discussions/1386) for the broader SSO conversation. Opened as draft specifically to be picked apart before considering it mergeable.

- Adds `Settings.Oidc` (disabled by default, config-file only — not exposed in the settings UI/API)
- `OidcAuthController` — `/api/v3/Auth/Oidc/Challenge` and `/Callback`, authorization-code flow with PKCE against the provider's discovery document
- Sign-in requires an explicit prior link (`GET /Auth/Oidc/Link` / `POST /Auth/Oidc/Unlink`) — no auto-matching by username/email
- Optional `OidcSettings.AutoCreateUsers` (default off) provisions a new local account on first sign-in
- Minted token expiry matches the OIDC ID token's own `exp` claim; `Unlink` invalidates all tokens minted under that provider link
- `ServerStatus.OidcEnabled` lets the WebUI conditionally show an SSO button without revealing provider details
- `docs/oidc-sso-setup.md` walks through IdP registration and account linking

## Security notes

- PKCE (`S256`); issuer pinned to configured `Authority` rather than trusting the discovery document's self-declared issuer
- `redirect_uri` built from an explicit `OidcSettings.PublicUrl` setting, not derived from request headers
- `returnUrl` validated via `Url.IsLocalUrl` before being carried through signed state (closes an open-redirect path)
- Discovery document cached per-authority with refresh-and-retry on validation failure, so IdP key rotation doesn't cause a sign-in outage
- Unique index on `JMMUser.ExternalAuthID` (all DB backends) closes a concurrent-link race

## Test plan

- [x] `dotnet test Shoko.Tests/Shoko.Tests.csproj` — 517 passed, 0 failed
- [ ] Manual end-to-end test against a real OIDC provider (feedback on the flow itself would help shape this first)

## Rebase history
- 2026-08-16 16:41 UTC — rebased onto `master` @ `4f0e3c52b` (11 commits), now at `666a8063f`
- 2026-08-23 16:23 UTC — rebased onto `upstream/master` @ `b6fdba59b` (11 commits), now at `99d56cc1a`

### DIFF
--- a/Shoko.Server/API/v3/Controllers/OidcAuthController.cs
+++ b/Shoko.Server/API/v3/Controllers/OidcAuthController.cs
@@ -16,7 +16,9 @@ using Microsoft.IdentityModel.JsonWebTokens;
 using Microsoft.IdentityModel.Protocols;
 using Microsoft.IdentityModel.Protocols.OpenIdConnect;
 using Microsoft.IdentityModel.Tokens;
+using Shoko.Abstractions.Exceptions;
 using Shoko.Abstractions.User.Services;
+using Shoko.Abstractions.User.Update;
 using Shoko.Server.API.Annotations;
 using Shoko.Server.Models.Shoko;
 using Shoko.Server.Repositories.Cached;
@@ -37,6 +39,7 @@ namespace Shoko.Server.API.v3.Controllers;
 public class OidcAuthController(
     ISettingsProvider settingsProvider,
     JMMUserRepository userRepository,
+    AuthTokensRepository authTokensRepository,
     IUserService userService,
     IHttpClientFactory httpClientFactory,
     IDataProtectionProvider dataProtectionProvider
@@ -86,9 +89,19 @@ public class OidcAuthController(
         if (User.ExternalAuthID is null)
             return NoContent();
 
+        InvalidateProviderTokens(User.JMMUserID, User.ExternalAuthID);
         User.ExternalAuthID = null;
         userRepository.Save(User);
         return NoContent();
+    }
+
+    // Tokens are keyed by device name "OIDC — {authority} — {subject}", so a provider-scoped
+    // prefix match invalidates every token minted for this user under the given authority
+    // without needing to look up the OIDC settings, which may have already changed.
+    private void InvalidateProviderTokens(int userID, string externalAuthID)
+    {
+        var authority = externalAuthID.Split("::", 2)[0];
+        authTokensRepository.DeleteWithUserIDAndDevicePrefix(userID, $"OIDC — {authority}");
     }
 
     private async Task<ActionResult> StartAuthorizeAsync(string? returnUrl, int? linkUserID, string? forwardedProto)
@@ -130,10 +143,10 @@ public class OidcAuthController(
             return RedirectToWebUiWithError($"OIDC provider returned an error: {error}");
 
         if (string.IsNullOrEmpty(code) || string.IsNullOrEmpty(state))
-            return BadRequest("Missing code or state.");
+            return RedirectToWebUiWithError("Missing code or state.");
 
         if (UnprotectState(state) is not { } statePayload || DateTime.UtcNow - statePayload.CreatedAt > StateLifetime)
-            return BadRequest("Invalid or expired sign-in attempt. Please try again.");
+            return RedirectToWebUiWithError("Invalid or expired sign-in attempt. Please try again.");
 
         var (settings, configuration, configError) = await GetEnabledConfigurationAsync();
         if (configError is not null)
@@ -144,10 +157,10 @@ public class OidcAuthController(
             return RedirectToWebUiWithError(exchangeError);
 
         var (claims, validationError) = await ValidateIdTokenAsync(configuration, settings, idToken!, statePayload.Nonce);
-        if (validationError is not null)
-            return RedirectToWebUiWithError(validationError);
+        if (validationError is not null || claims is null)
+            return RedirectToWebUiWithError(validationError ?? "ID token validation failed.");
 
-        var rawSubject = claims!.FindFirst("sub")?.Value;
+        var rawSubject = claims.FindFirst("sub")?.Value;
         if (string.IsNullOrEmpty(rawSubject))
             return RedirectToWebUiWithError("ID token is missing a subject claim.");
 
@@ -157,17 +170,19 @@ public class OidcAuthController(
 
         var (user, userError) = statePayload.LinkUserID is { } linkUserID
             ? LinkUser(linkUserID, externalAuthID)
-            : ResolveUser(externalAuthID);
+            : await ResolveUserAsync(externalAuthID, rawSubject, settings);
         if (userError is not null)
             return RedirectToWebUiWithError(userError);
 
         // Match the Shoko token's lifetime to the OIDC token's own expiry rather than a
         // fixed value — a fresh login always mints a new token, so several can coexist
         // with different expirations without any one of them ever being non-expiring.
-        if (GetExpiration(claims!) is not { } expiresAt || expiresAt <= DateTime.UtcNow.AddMinutes(1))
+        if (GetExpiration(claims) is not { } expiresAt || expiresAt <= DateTime.UtcNow.AddMinutes(1))
             return RedirectToWebUiWithError("ID token is missing a valid expiration.");
 
-        var apiToken = await userService.GenerateApiTokenForUser(user!, $"OIDC — {settings.Authority}", expiresAt);
+        // Subject is part of the device name (not just externalAuthID) so a provider-scoped
+        // Unlink() can target exactly the tokens minted for this identity via prefix match.
+        var apiToken = await userService.GenerateApiTokenForUser(user!, $"OIDC — {settings.Authority} — {rawSubject}", expiresAt);
         return RedirectToWebUiWithToken(apiToken.Token, statePayload.ReturnUrl);
     }
 
@@ -222,13 +237,36 @@ public class OidcAuthController(
 
     // Sign-in only ever resolves a user that was already explicitly linked via Link() —
     // never matches by username or email, which would let anyone controlling (or
-    // impersonating) the IdP take over a same-named local account.
-    private (JMMUser? User, string? Error) ResolveUser(string externalAuthID)
+    // impersonating) the IdP take over a same-named local account. The one opt-in exception
+    // is AutoCreateUsers, which provisions a brand new account rather than linking an
+    // existing one, so it can't be used to take over anything.
+    private async Task<(JMMUser? User, string? Error)> ResolveUserAsync(string externalAuthID, string rawSubject, OidcSettings settings)
     {
         var user = userRepository.GetByExternalAuthID(externalAuthID);
-        return user is not null
-            ? (user, null)
-            : (null, "No local Shoko account is linked to this SSO identity. Sign in locally and link your account first.");
+        if (user is not null)
+            return (user, null);
+
+        if (!settings.AutoCreateUsers)
+            return (null, "No local Shoko account is linked to this SSO identity. Sign in locally and link your account first.");
+
+        if (userRepository.GetByUsername(rawSubject) is not null)
+            return (null, $"Cannot auto-create a user for subject \"{rawSubject}\" — that username is already taken.");
+
+        try
+        {
+            var created = (JMMUser)await userService.CreateUser(new UserUpdate
+            {
+                Username = rawSubject,
+                Password = Convert.ToHexString(RandomNumberGenerator.GetBytes(32)),
+            });
+            created.ExternalAuthID = externalAuthID;
+            userRepository.Save(created);
+            return (created, null);
+        }
+        catch (GenericValidationException ex)
+        {
+            return (null, $"Could not auto-create a user for subject \"{rawSubject}\": {ex.Message}");
+        }
     }
 
     private (JMMUser? User, string? Error) LinkUser(int linkUserID, string externalAuthID)
--- a/Shoko.Server/API/v3/Controllers/OidcAuthController.cs
+++ b/Shoko.Server/API/v3/Controllers/OidcAuthController.cs
@@ -1,0 +1,264 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Net.Http.Json;
+using System.Security.Claims;
+using System.Security.Cryptography;
+using System.Text;
+using System.Text.Json;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.DataProtection;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.WebUtilities;
+using Microsoft.IdentityModel.JsonWebTokens;
+using Microsoft.IdentityModel.Protocols;
+using Microsoft.IdentityModel.Protocols.OpenIdConnect;
+using Microsoft.IdentityModel.Tokens;
+using Shoko.Abstractions.User.Services;
+using Shoko.Server.API.Annotations;
+using Shoko.Server.Models.Shoko;
+using Shoko.Server.Repositories.Cached;
+using Shoko.Server.Settings;
+
+namespace Shoko.Server.API.v3.Controllers;
+
+/// <summary>
+/// Optional OpenID Connect single sign-on. Disabled unless configured in
+/// Settings.Oidc. Never creates local accounts — a matching local account
+/// (by username or email) must already exist; SSO only links to it and
+/// mints the same kind of API token the local sign-in endpoint issues.
+/// </summary>
+[ApiController]
+[Route("/api/v{version:apiVersion}/Auth/Oidc")]
+[ApiV3]
+public class OidcAuthController(
+    ISettingsProvider settingsProvider,
+    JMMUserRepository userRepository,
+    IUserService userService,
+    IHttpClientFactory httpClientFactory,
+    IDataProtectionProvider dataProtectionProvider
+) : ControllerBase
+{
+    private const string ProtectorPurpose = "Shoko.Server.OidcAuth.State";
+    private static readonly TimeSpan StateLifetime = TimeSpan.FromMinutes(10);
+
+    private OidcSettings Settings => settingsProvider.GetSettings().Oidc;
+
+    private sealed record StatePayload(string Nonce, string? ReturnUrl, DateTime CreatedAt);
+
+    /// <summary>
+    /// Redirects the browser to the configured OIDC provider's authorization
+    /// endpoint.
+    /// </summary>
+    [HttpGet("Challenge")]
+    [AllowAnonymous]
+    public async Task<ActionResult> Challenge(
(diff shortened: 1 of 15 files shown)


Foowy:
and back to appeasing the  sonarqube gods

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
```diff
@@ -0,0 +1,264 @@
...
+
+        var subject = claims!.FindFirst("sub")?.Value;
+        if (string.IsNullOrEmpty(subject))
+            return RedirectToWebUiWithError("ID token is missing a subject claim.");
+
+        var (user, userError) = ResolveUser(claims, subject);
+        if (userError is not null)
+            return RedirectToWebUiWithError(userError);
+
+        var apiToken = await userService.GenerateApiTokenForUser(user!, "OIDC");
```
revam: Can you set the expire date on the token, and maybe also append the authority ID and/or external id to the key? e.g. `$"OIDC — {settings.Authority} — {subject}"`

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
```diff
@@ -0,0 +1,264 @@
...
+        }
+        catch
+        {
+            return null;
+        }
+    }
+
+    private ActionResult RedirectToWebUiWithToken(string token, string username, string? returnUrl)
+    {
+        var basePath = settingsProvider.GetSettings().Web.WebUIPublicPath.TrimEnd('/');
```
revam: Don't trim the end of the web ui public path. also, no need to always get it if we have a return url. I'll recommend doing `returlUrl ??= settingsProvider.GetSettings().Web.WebUIPublicPath` instead, then using `returnUrl`, so no `basePath` and `target` variables.

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
```diff
@@ -0,0 +1,264 @@
...
+        var target = string.IsNullOrWhiteSpace(returnUrl) ? basePath : returnUrl;
+        return Redirect($"{target}#oidcToken={Uri.EscapeDataString(token)}&oidcUsername={Uri.EscapeDataString(username)}");
+    }
+
+    private ActionResult RedirectToWebUiWithError(string message)
+    {
+        var basePath = settingsProvider.GetSettings().Web.WebUIPublicPath.TrimEnd('/');
```
revam: See above comment.

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
```diff
@@ -0,0 +1,264 @@
...
+        var tokenNonce = claims.FindFirst("nonce")?.Value;
+        if (tokenNonce is null || !CryptographicOperations.FixedTimeEquals(Encoding.UTF8.GetBytes(tokenNonce), Encoding.UTF8.GetBytes(expectedNonce)))
+            return (null, "ID token nonce mismatch.");
+
+        return (claims, null);
+    }
+
+    private (JMMUser? User, string? Error) ResolveUser(ClaimsIdentity claims, string subject)
+    {
+        var user = userRepository.GetByExternalAuthID(subject);
```
revam: seems like we may have cross-authority collusion if we ever switch the authority. I'll recommend using the authority id as a prefix for the subject to prevent this.

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
```diff
@@ -0,0 +1,264 @@
...
+        {
+            return null;
+        }
+    }
+
+    private ActionResult RedirectToWebUiWithToken(string token, string username, string? returnUrl)
+    {
+        var basePath = settingsProvider.GetSettings().Web.WebUIPublicPath.TrimEnd('/');
+        var target = string.IsNullOrWhiteSpace(returnUrl) ? basePath : returnUrl;
+        return Redirect($"{target}#oidcToken={Uri.EscapeDataString(token)}&oidcUsername={Uri.EscapeDataString(username)}");
```
revam: no need to send our local username with the token. the web ui will fetch the user details from the API if necessary. we could show the upstream user/nick name though if we wanted.

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
```diff
@@ -0,0 +1,264 @@
...
+    private string BuildRedirectUri(string? forwardedProto)
+    {
+        // Request.Scheme reflects the connection Kestrel actually accepted, which is
+        // http when Shoko sits behind a TLS-terminating reverse proxy. Honor
+        // X-Forwarded-Proto so the redirect_uri we send matches what's registered
+        // with the OIDC provider.
+        var scheme = !string.IsNullOrWhiteSpace(forwardedProto)
+            ? forwardedProto.Split(',')[0].Trim()
+            : Request.Scheme;
+        return $"{scheme}://{Request.Host}/api/v3/Auth/Oidc/Callback";
```
revam: i feel like we should be able to override the redirect url in the settings so it 100% matches what we set in the provider.

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
```diff
@@ -0,0 +1,264 @@
...
+            return (user, null);
+
+        // First SSO login for this external identity — link to a
+        // pre-existing local account by username or email, never create one.
+        // Only trust the email claim for matching if the provider marked it
+        // verified; an unverified email lets any IdP user claim someone else's.
+        var emailVerified = bool.TryParse(claims.FindFirst("email_verified")?.Value, out var verified) && verified;
+        var candidateNames = new[]
+        {
+            claims.FindFirst("preferred_username")?.Value,
```
revam: so if you have a local admin account named `Default`, and an attacker somehow manages to comprimise the OIDC authority then the only thing they need to take over the local admin is name their preferred username as `Default`? Why aren't we checking if the user below already is linked to an external auth id, and is instead overwriting it's stored id with the new id from the sign in attempt?

Also, i feel like the linking should not be possible for existing users during sign in, and we should instead have a **separate** flow for existing users to tie their local only account to an authority. We should also have a way to untie the account from the external authority.

And we should create new accounts for new users which we don't recognise from the subject/auth-user-id, during sign in. Again, **not** link them to an existing acocunt. If the preferred username is taken, then we would get the next best username, or we can just use the email address as the username, and if the email is in use, show an error as we don't want to duplicate or create similar usernames using emails automagically.

revam (commented):
A few comments so far.

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
harshithmohan: > i feel like we should be able to override the redirect url in the settings so it 100% matches what we set in the provider.

You set the callback URL in the provider depending on what the application takes, not the other way around.
So that shouldn't be customizable in our application.

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
harshithmohan: Email based account matching? What emails? 🤔 

Also, as revam mentioned, it would be better to have a separate flow for linking existing users rather than linking them on first time OAuth login. That only works from a security perspective when we do have an identifier to match that verifiably belongs to a particular user, like email (which doesn't exist). Doing it by username means anybody can put any existing Shoko user's name and it will match to that.

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
Foowy: Yes, what kind of range are were looking at. 7-14-30 (Feel these are safe ranges) 60-90 days?

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
Foowy: Will update

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
Foowy: Ditto to above

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
Foowy: Will update

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
harshithmohan: I think it should be same as the expiry of the oidc access token but that can be refreshed too. Ideally we shouldn't be generating a normal api token (I asume that's what it's doing) and be able to use the actual token itself for auth. Needs more discussion

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
Foowy: Would it be better to just drop it? Looks like WebUI already pulls it from api. Just pass the token in the redirect?

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
Foowy: Will see what i can come up with. Agree with the security issue. Will make changes to see if i can get it more inline to both your comments.

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
revam: @harshithmohan @hidden4003 thoughts from the frontend team?

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
revam: I say match the ocid token which is granting our token. and by setting an expiration we could have multiple tokens with the same key with different expiration dates, so you can't hijack an exiting token, and also won't get an everlasting/non-expiring token.

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
```diff
@@ -0,0 +1,323 @@
...
+[ApiController]
+[Route("/api/v{version:apiVersion}/Auth/Oidc")]
+[ApiV3]
+public class OidcAuthController(
+    ISettingsProvider settingsProvider,
+    JMMUserRepository userRepository,
+    IUserService userService,
+    IHttpClientFactory httpClientFactory,
+    IDataProtectionProvider dataProtectionProvider
+) : ControllerBase
```
revam: Extend BaseController instead, so we don't need the `User` virtual property and so all the APIv3 controllers use our existing base controller convention.

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
Foowy: The Shoko API token minted after OIDC sign-in/link now expires exactly when the OIDC ID token does (read from its exp claim) instead of a fixed 30-day window, per @revam's suggestion. If the ID token has no valid exp, the flow errors out rather than minting a token with a made-up lifetime.

Also confirmed the token store already supports this without changes — AuthTokensRepository.CreateExpiringApiKey (the path used here) never deduped by device name, so multiple tokens can already coexist under the same `OIDC — {authority}` key, each with its own expiry. A fresh login just adds a new row; nothing gets silently reused or extended.

This can again be altered or changed to something else.

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
Foowy: sorry forgot to keep commits broken apart and auto-squashed out of habit, but will make those changes

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
Foowy: I have not touched this code yet As both comments are counter to each other., waiting for further discussion.

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
Foowy: The second commit includes my suggestion of dropping it, but can be changed if i read the code wrong

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
harshithmohan: > I say match the ocid token which is granting our token. and by setting an expiration we could have multiple tokens with the same key with different expiration dates, so you can't hijack an exiting token, and also won't get an everlasting/non-expiring token.

~~What happens when the shoko token gets expired before the OIDC token does?~~
nvm, I just read Foowy's message, that should work, I guess

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
harshithmohan: Yes, we already call the API to get the user name, we don't need it here

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
harshithmohan: I assume revam's thumbup means we can keep it as is

revam: i didn't have anything else to add, so you assumed correctly

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
```diff
@@ -0,0 +1,314 @@
...
+        var (user, userError) = statePayload.LinkUserID is { } linkUserID
+            ? LinkUser(linkUserID, externalAuthID)
+            : ResolveUser(externalAuthID);
+        if (userError is not null)
+            return RedirectToWebUiWithError(userError);
+
+        // Match the Shoko token's lifetime to the OIDC token's own expiry rather than a
+        // fixed value — a fresh login always mints a new token, so several can coexist
+        // with different expirations without any one of them ever being non-expiring.
+        if (GetExpiration(claims!) is not { } expiresAt || expiresAt <= DateTime.UtcNow.AddMinutes(1))
```
revam: you can add a `claims is null` check to the if statement above to prevent the need of this assertion

revam: not the userError one, the validationError one

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
```diff
@@ -0,0 +1,314 @@
...
+            return RedirectToWebUiWithError("ID token is missing a valid expiration.");
+
+        var apiToken = await userService.GenerateApiTokenForUser(user!, $"OIDC — {settings.Authority}", expiresAt);
```
revam: so we're not going to include the upstream subject in the token key? also related but unrelated, but we should probably invalidate all tokens tied to a pr9vider when unlinking the user from said provider, andnwhen disabling or switching the provider.

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
```diff
@@ -0,0 +1,314 @@
...
+    /// Removes the external identity link from the currently signed-in
+    /// local account, if any.
+    /// </summary>
+    [HttpPost("Unlink")]
+    [Authorize]
+    public ActionResult Unlink()
+    {
+        if (User.ExternalAuthID is null)
+            return NoContent();
+
```
revam, line 92: so invalidate all auth tokens for the provider and user here, which we can grab from the prefix of the external user id.

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
```diff
@@ -0,0 +1,314 @@
...
+        [FromHeader(Name = "X-Forwarded-Proto")] string? forwardedProto = null)
+    {
+        if (!string.IsNullOrEmpty(error))
+            return RedirectToWebUiWithError($"OIDC provider returned an error: {error}");
+
+        if (string.IsNullOrEmpty(code) || string.IsNullOrEmpty(state))
+            return BadRequest("Missing code or state.");
+
+        if (UnprotectState(state) is not { } statePayload || DateTime.UtcNow - statePayload.CreatedAt > StateLifetime)
+            return BadRequest("Invalid or expired sign-in attempt. Please try again.");
```
revam: I feel like we should be redirecting to the webui for _all_ the errors here, so the web ui can render them instead of the user seeing a white page with plain text. sounds like a better ux to me. @harshithmohan thoughts?

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
```diff
@@ -0,0 +1,314 @@
...
+
+        var (settings, configuration, configError) = await GetEnabledConfigurationAsync();
+        if (configError is not null)
```
revam, line 163: the same as above would apply here.

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
```diff
@@ -0,0 +1,314 @@
...
+
+        var rawSubject = claims!.FindFirst("sub")?.Value;
+        if (string.IsNullOrEmpty(rawSubject))
+            return RedirectToWebUiWithError("ID token is missing a subject claim.");
+
+        // Prefix with the authority so switching Settings.Oidc.Authority can never make an
+        // identity minted by a different provider resolve to the same external auth ID.
+        var externalAuthID = $"{settings.Authority}::{rawSubject}";
+
+        var (user, userError) = statePayload.LinkUserID is { } linkUserID
```
revam, line 188: Can we add an option (default off/disabled) to allow for auto-adding new users based on the subject when the link id is not set in the challenge/callback payload? Their password should be auto generated and not left blank. Saying that just in case it wasn't obvious.

revam: If the subject is not available then we can somehow increment it until it is, unless it's an email, in which case we should hard abort.

revam: Actually, same would apply to numerical ids... maybe just hard abort if unavailable in general.

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
harshithmohan: Yes

Shoko.Server/API/v3/Controllers/OidcAuthController.cs
harshithmohan: > Can we add an option (default off/disabled) to allow for auto-adding new users based on the subject when the link id is not set in the challenge/callback payload? Their password should be auto generated and not left blank. Saying that just in case it wasn't obvious.

It would be the numerical ID or the email by default most of the times. I think the option is not worth it

harshithmohan: @Foowy alredy added it before I could comment 😔

Foowy:
Added documentation and further changes to match current specs and safeguards

PR #1389, opened 2026-07-22
fix(tmdb): add season-scoped nearest-air-date fallback for episode linking

Foowy:
## Summary
- `TryFindAnidbAndTmdbMatch` previously fell straight from the strict ±2-day `TryAirDateMatch` window to a blind positional `FirstAvailable` guess, which mismatched delayed/compressed episodes onto the wrong TMDB slot (e.g. AniDB episode 2 grabbing episode 5's TMDB entry).
- Added `TryNearestAirDateMatch`, tagged with a new `MatchRating.DateKindaMatches`, using the unbounded `CalculateAirDateDistance` to pick the closest remaining candidate by air date instead of by list position, bounded by `MaxFallbackDifferenceInDays` (120) so it can't confidently link episodes separated by an unrelated multi-month gap.
- Scoped the fallback candidate pool to `anchorSeasonNumber`, resolved from whichever neighboring AniDB episode already has a confirmed TMDB link, so a back-to-back multi-cour release (e.g. season 1 ending in March, season 2 starting in June) can't have this pull an unresolved episode across the season boundary. When neighbors disagree on season, the closer one by air date wins instead of always preferring the previous episode.
- `ResolveRankedMatch`'s cache-reuse now also invalidates when a same-pass neighbor resolves a new anchor mid-pass, not just when the cached TMDB candidate gets claimed.
- Renamed `MatchRating.NearestDateMatches` to `DateKindaMatches` to match the existing `*KindaMatches` naming convention (`TitleKindaMatches`, `DateAndTitleKindaMatches`), per harshithmohan's review.

## Follow-up fixes (post-review)
- `anidbEpisodesByTypeNumber` is now built with `GroupBy`+`First` instead of `ToDictionary`, which threw on any duplicate `(EpisodeType, EpisodeNumber)` pair.
- The loose (±120 day) `TryNearestAirDateMatch` fallback is now gated by `isSpecial`, matching the positional `FirstAvailable` fallback beside it — a special with no title match could previously auto-link to whatever TMDB episode was nearest in air date up to four months away, with zero corroborating evidence. Covered by two new tests.
- Collapsed two duplicate copies of the special-episode predicate (a closure and a static method reimplementing the same check) into one shared `IsSpecialEpisode(EpisodeType, AnimeType)` helper.
- Rewrote the `nearestAirdate` LINQ chain as a query expression so the null-check is visible to nullable analysis across the following `select`, clearing a CS8629 warning.
- Wrapped new method signatures that exceeded the repo's 160-character line limit.

## Test plan
- [x] `dotnet build Shoko.Server.sln`
- [x] `dotnet test Shoko.Tests/Shoko.Tests.csproj --filter "FullyQualifiedName~TmdbLinkingServiceTests"` (11/11 passing, including 4 gap-handling tests)
- [x] `dotnet test Shoko.Tests/Shoko.Tests.csproj` (521/521 passing)

## Rebase history
- 2026-08-10 00:53 UTC — rebased onto `master` @ `7b1523c88` (6 commits), now at `9aeae1017`
- 2026-08-13 17:22 UTC — rebased onto `master` @ `d4dff0c4d` (8 commits), now at `a84f981e6`

### DIFF
--- a/Shoko.Abstractions/Metadata/Enums/MatchRating.cs
+++ b/Shoko.Abstractions/Metadata/Enums/MatchRating.cs
@@ -53,4 +53,9 @@ public enum MatchRating : byte
     ///   Date and Title are close, but not exact.
     /// </summary>
     DateAndTitleKindaMatches = 8,
+
+    /// <summary>
+    ///   No title or in-window date evidence; matched to the closest available air date as a last resort.
+    /// </summary>
+    NearestDateMatches = 9,
 }
--- a/Shoko.Abstractions/Metadata/Enums/MatchRating.cs
+++ b/Shoko.Abstractions/Metadata/Enums/MatchRating.cs
@@ -57,5 +57,5 @@ public enum MatchRating : byte
     /// <summary>
     ///   No title or in-window date evidence; matched to the closest available air date as a last resort.
     /// </summary>
-    NearestDateMatches = 9,
+    DateKindaMatches = 9,
 }
--- a/Shoko.Server/Providers/TMDB/TmdbLinkingService.cs
+++ b/Shoko.Server/Providers/TMDB/TmdbLinkingService.cs
@@ -472,6 +472,17 @@ public class TmdbLinkingService : ITmdbLinkingService
         var toSkip = new HashSet<int>();
         var toAdd = new List<CrossRef_AniDB_TMDB_Episode>();
         var crossReferences = new List<CrossRef_AniDB_TMDB_Episode>();
+        // Kept alongside crossReferences (not derived from it) so ResolveAnchorSeason can look up a
+        // neighbor's primary (Ordering == 0) link in O(1) instead of rescanning the whole list, which
+        // would otherwise get quadratically slower as crossReferences grows across passes.
+        var primaryLinkByAnidbEpisodeId = new Dictionary<int, CrossRef_AniDB_TMDB_Episode>();
+
+        void AddCrossReference(CrossRef_AniDB_TMDB_Episode xref)
+        {
+            crossReferences.Add(xref);
+            if (xref.TmdbEpisodeID != 0 && xref.Ordering == 0)
+                primaryLinkByAnidbEpisodeId[xref.AnidbEpisodeID] = xref;
+        }
         var secondPass = new List<AniDB_Episode>();
         var fourthPass = new List<AniDB_Episode>();
         var thirdPass = new List<AniDB_Episode>();
@@ -483,6 +494,7 @@ public class TmdbLinkingService : ITmdbLinkingService
             .OrderBy(episode => episode.EpisodeType)
             .ThenBy(episode => episode.EpisodeNumber)
             .ToDictionary(episode => episode.EpisodeID);
+        var anidbEpisodesByTypeNumber = anidbEpisodes.Values.ToDictionary(episode => (episode.EpisodeType, episode.EpisodeNumber));
         var tmdbEpisodeDict = _tmdbEpisodes.GetByTmdbShowID(tmdbShowId)
             .ToDictionary(episode => episode.TmdbEpisodeID);
         var tmdbEpisodes = tmdbEpisodeDict.Values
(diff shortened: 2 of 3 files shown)


da3dsoul:
I know we talked about this a bunch. @revam were you good with the decisions made? I'm pretty sure I was in agreement with them

da3dsoul:
@Foowy one thing to check: I just pushed some fixes for Seasons recently. We should reuse that if we aren't

Shoko.Abstractions/Metadata/Enums/MatchRating.cs
```diff
@@ -53,4 +53,9 @@ public enum MatchRating : byte
     ///   Date and Title are close, but not exact.
     /// </summary>
     DateAndTitleKindaMatches = 8,
+
+    /// <summary>
+    ///   No title or in-window date evidence; matched to the closest available air date as a last resort.
+    /// </summary>
+    NearestDateMatches = 9,
```
revam: adding this requires a webui change, so if possible, then just shove it under DateKindaMatches, or consult the webui team to add it. 

@harshithmohan @hidden4003 thoughts on adding it?

Shoko.Server/Providers/TMDB/TmdbLinkingService.cs
```diff
@@ -516,30 +528,80 @@ public IReadOnlyList<ITmdbEpisodeCrossReference> MatchAnidbToTmdbEpisodes(int an
...
 
+        // Finds the TMDB season an episode's adjacent AniDB neighbor (previous and next) already linked
+        // to, so the nearest-air-date fallback in TryFindAnidbAndTmdbMatch can be scoped to that season
+        // instead of searching across all seasons. Without this, a back-to-back multi-cour release (e.g.
+        // season 1 ending in March, season 2 starting in June) can have less than the fallback's date
+        // tolerance between season boundaries, letting an unresolved season-1 episode get pulled onto a
+        // season-2 episode instead of staying unlinked. Returns null for OVAs, where there's only one pool.
+        // Looks up primaryLinkByAnidbEpisodeId (O(1), Ordering == 0 only) rather than crossReferences
+        // directly, so a multi-part neighbor's secondary/duplicate link can't be picked up by mistake.
+        int? ResolveAnchorSeason(AniDB_Episode episode)
```
revam: any reason preventing this from not being an inline function?

Shoko.Server/Providers/TMDB/TmdbLinkingService.cs
```diff
@@ -516,30 +528,80 @@ public IReadOnlyList<ITmdbEpisodeCrossReference> MatchAnidbToTmdbEpisodes(int an
...
+            return nextDistance < previousDistance ? next.Value.Season : previous.Value.Season;
+        }
+
         // Ranks episodes by their best candidate match without consuming it, so passes can process the
         // strongest matches first and let weaker duplicate claims fall back to their next-best candidate
         // instead of losing a shared TMDB episode purely by AniDB episode order. The match found here is
         // cached and handed back to the caller so the loop that actually consumes candidates doesn't have
         // to run the same title-search/air-date computation a second time for every episode.
-        List<(AniDB_Episode Episode, CrossRef_AniDB_TMDB_Episode CrossRef)> RankByConfidence(IEnumerable<AniDB_Episode> episodes) =>
+        List<(AniDB_Episode Episode, CrossRef_AniDB_TMDB_Episode CrossRef, int? AnchorSeasonNumber)> RankByConfidence(IEnumerable<AniDB_Episode> episodes) =>
```
revam: same question here.

Shoko.Server/Providers/TMDB/TmdbLinkingService.cs
```diff
@@ -516,30 +528,80 @@ public IReadOnlyList<ITmdbEpisodeCrossReference> MatchAnidbToTmdbEpisodes(int an
...
-        // TMDB candidate hasn't since been claimed by an earlier (stronger) episode in the same pass —
-        // in that case re-running the deterministic match against the unchanged remaining pool is
-        // guaranteed to produce the same result, so the recompute is skipped.
-        CrossRef_AniDB_TMDB_Episode ResolveRankedMatch(AniDB_Episode episode, CrossRef_AniDB_TMDB_Episode cached, List<TMDB_Episode> episodeList) =>
-            cached.TmdbEpisodeID != 0 && episodeList.Any(candidate => candidate.TmdbEpisodeID == cached.TmdbEpisodeID)
+        // TMDB candidate hasn't since been claimed by an earlier (stronger) episode in the same pass AND
+        // the episode's anchor season hasn't changed since — a same-pass neighbor getting linked earlier
+        // in this same loop can newly resolve an anchor that didn't exist when RankByConfidence ran, so
+        // pool-membership alone isn't enough to prove the cached match is still correct.
+        CrossRef_AniDB_TMDB_Episode ResolveRankedMatch(AniDB_Episode episode, CrossRef_AniDB_TMDB_Episode cached, int? cachedAnchorSeasonNumber, List<TMDB_Episode> episodeList)
```
revam: and here.

Shoko.Server/Providers/TMDB/TmdbLinkingService.cs
```diff
@@ -909,7 +995,7 @@ private static bool TryAirDateMatch(
         out CrossRef_AniDB_TMDB_Episode crossRef,
```
revam: mark this as nullable and add the attribute to say it's not null when true.

Shoko.Server/Providers/TMDB/TmdbLinkingService.cs
```diff
@@ -928,32 +1014,67 @@ private static bool TryAnyTitleMatch(
...
```
revam: same as above

revam (changes requested):
comments

revam:
> I know we talked about this a bunch. @revam were you good with the decisions made? I'm pretty sure I was in agreement with them

adding it is fine. impl. wise i added some comments above.

Shoko.Abstractions/Metadata/Enums/MatchRating.cs
Foowy: Could fold it in, it's only used in 3 spots (including 1 test) - but nothing existing quite fits. FirstAvailable gets rejected by IsAcceptableAutoMatch, so that'd quietly kill this whole fallback. DateAndTitleKindaMatches is a guess and implies info we dont have. DateMatches is probably the least-bad squeeze if we want to pass on a WebUI change. 

If not, WebUI would be the next step.

Foowy:
> @Foowy one thing to check: I just pushed some fixes for Seasons recently. We should reuse that if we aren't

Checked it out — that fix is for AniDB broadcast-season/year classification (the Winter/Spring/Summer/Fall library filters). This PR's "season" is about matching AniDB episodes to the right TMDB TV season instead. Nothing to reuse that i can see. Correct me if i missed something

Shoko.Server/Providers/TMDB/TmdbLinkingService.cs
Foowy: These are already local functions, just called from two spots each (main pass + fallback pass). Inlining would mean copy-pasting the same logic twice per function instead of simplifying/de-duplicating.

Shoko.Server/Providers/TMDB/TmdbLinkingService.cs
Foowy: See Above

Shoko.Server/Providers/TMDB/TmdbLinkingService.cs
Foowy: See Above

da3dsoul:
> > @Foowy one thing to check: I just pushed some fixes for Seasons recently. We should reuse that if we aren't
> 
> Checked it out — that fix is for AniDB broadcast-season/year classification (the Winter/Spring/Summer/Fall library filters). This PR's "season" is about matching AniDB episodes to the right TMDB TV season instead. Nothing to reuse that i can see. Correct me if i missed something

Just double checking

Shoko.Server/Providers/TMDB/TmdbLinkingService.cs
revam: i meant the opposite, extract them into proper class members/methods instead of inline/local functions.

Shoko.Abstractions/Metadata/Enums/MatchRating.cs
Foowy: @harshithmohan @hidden4003 - Re-pinging you in on the NearestDateMatches addition to MatchRating in Shoko.Abstractions/Metadata/Enums/MatchRating.cs.

Confirmed this is backward-compatible on the WebUI side - no change is required for it to keep working. MatchRatingType in src/core/types/api/episode.ts is a string-keyed TS enum consumed by MatchRating.tsx. That component's getAbbreviation() switch has a default: return ['', ''] fallback, and none of the cx() background-color conditions match an unrecognized value either - so an episode with NearestDateMatches renders with no crash, just no color badge and no tooltip abbreviation. It falls through cleanly rather than breaking.

That said, I can create a PR on WebUI for this if you want - mainly just adding the enum value, a badge abbreviation, and a color tier so it's not blending into "uncategorized." Let me know either way.

harshithmohan: I don't see any issue with adding it to the webui. But maybe it should be named `DataKindaMatches` instead of `NearestDateMatches` to match the others for consistency.
So make a PR to the webui *after* this is merged into server. Also, if it's an enum on the webui, it should be changed to a string union instead as part of the changes

Foowy: Changes made and draft pr on webui that i will move to review if/when this PR gets merged

Foowy:
Any other changes needed or should i archive this for now?

revam (approved):

PR #1391, opened 2026-07-22
feat(tmdb): TMDbLib master upgrade support, with restricted auto-match genre fix

Foowy:
## Summary
- `CollectCandidates`/`CollectMovieCandidates` required every TMDB search hit to carry the "animation" genre tag before it was even considered a candidate. TMDB's genre metadata is sparse for R18/adult content, so a correct match was often discarded before scoring ran.
- Restricted titles now also accept a match on `original_language == "ja"` - TMDB-assigned metadata, not community-tagged, so it doesn't share the genre field's sparsity problem. This restores real content-type filtering for restricted search instead of bypassing the check outright.
- Extracted the near-duplicate cast/crew diffing in `UpdateMovieCastAndCrew`/`UpdateEpisodeCastAndCrew` into shared generic helpers (`DiffCast`/`DiffCrew`), then further extracted `UpdateMovieCastAndCrew`'s per-person update and orphan-cleanup logic into their own methods. Fixes three SonarCloud cognitive-complexity findings (35→, 23→, 17→, all against a limit of 15). Same add/update/remove/save/delete/logging behavior.
- Dropped two redundant `(int)` casts left over from TMDbLib's pre-upgrade string-typed episode-number fields.
- Removed `Dockerfile.tmdblib-dev` from history entirely - a local-dev-only container helper that never belonged in an upstream PR, previously flagged by SonarCloud (root-user default, line-length). Dropped via `git rebase -i` on the commit that introduced it, not just deleted going forward.

**Draft - pending an update to [TMDbLib](https://github.com/jellyfin/TMDbLib).** This branch now also carries the TMDbLib-upgrade work (submodule build against TMDbLib's unreleased `master`, external-ID/5xx/null-guard fixes, `episode_type` storage) so the genre fix can be validated against it ahead of an actual release.

## TMDbLib upgrade improvements

Built against TMDbLib `master` (jellyfin/TMDbLib, submodule pinned `a4a3fbb`) via a git submodule + `ProjectReference` swap, since no released package exists past `3.0.0` yet:

- External-ID handling (`TvdbId`) aligned with TMDbLib's `string?` → `int?` type fix - also simpler, since Shoko's local model was already `int?`.
- 5xx responses now route through the retry policy (`TMDbServerException`/`TMDbServiceUnavailableException`), matching TMDbLib master's new exception types instead of the old `GeneralHttpException` catch-all.
- Null-guards added for TMDb response fields (translations, credits, image lists) that TMDbLib master no longer force-unwraps.
- Stores TMDB's real `episode_type` field instead of Shoko's season-number-based guess.
- Verified via a parallel A/B test (3.0.0 baseline vs TMDbLib master, full library re-scan): TMDB request counts identical, 12/14 DB tables byte-identical. A 2-poster mapping gap seen in the initial pass (downloaded-but-unlinked images) did not reproduce on a clean re-scan against the current tip - resolved by an earlier null-guard fix, no code change needed.
- Submodule tracked continuously - bumped 2026-09-14 from `4744732` to `a4a3fbb` (3 upstream dependency bumps, no API changes), rebuilt and retested clean.

**Caveat:** all of the above is verified against an unreleased TMDbLib `master` snapshot, not a published package. The actual TMDbLib release may drift further before it ships - this work (including the exception types, the `TvdbId` type, and the null-guard sites) will need re-verification against the real release before merge, and specifics here may change.

## SonarCloud: comment kept intentionally

SonarCloud's `S125` ("remove this commented out code") flags this comment in `UpdateMovieCastAndCrew`/`UpdateEpisodeCastAndCrew`:

> A present-but-empty `credits` append can deserialize to null members under STJ; treat that as "no cast/crew this pass" rather than purging existing rows.

It's a false positive -`S125` is pattern-matching the comment's shape (backticked words, semicolons), not actually detecting disabled code. The comment explains a real edge case: TMDB sometimes sends back a `credits` field that looks empty instead of just omitting it. If the code treated that as "there's really no cast/crew," it would delete every cast/crew row we already have on file - even though the real cause might just be a flaky or partial API response. So instead it skips that pass and leaves existing data alone. The comment exists so nobody later "fixes" that check by making it delete data again. Left in place; should be marked Won't Fix/False Positive in SonarCloud rather than deleted.

## Test plan
- [x] `dotnet build Shoko.Server.sln` (against TMDbLib master via submodule)
- [x] `dotnet test Shoko.Tests/Shoko.Tests.csproj` (3940/3940 passing)

## Rebase history
- 2026-09-16 02:11 UTC - rebased onto `upstream/master` @ `2fa9089a2` (13 commits), now at `ac8eec9d4`
- 2026-09-19 20:53 UTC - rebased onto `upstream/master` @ `0fe4fc582` (13 commits), now at `19cd740a6`

### DIFF
--- a/Shoko.Server/Providers/TMDB/TmdbSearchService.cs
+++ b/Shoko.Server/Providers/TMDB/TmdbSearchService.cs
@@ -264,7 +264,7 @@ public partial class TmdbSearchService : ITmdbSearchService
 
         // Attempt #1: full title + year
         (results, _) = await SearchMoviesRaw(query, includeRestricted: includeRestricted, year: year).ConfigureAwait(false);
-        CollectMovieCandidates(candidates, results, seen, candidateCount);
+        CollectMovieCandidates(candidates, results, seen, candidateCount, includeRestricted);
 
         // Attempt #2: sequel-suffix stripped + year
         var strippedTitle = SequelSuffixRemovalRegex().Match(query) is { Success: true } regexResult
@@ -272,7 +272,7 @@ public partial class TmdbSearchService : ITmdbSearchService
         if (!string.IsNullOrEmpty(strippedTitle) && candidates.Count < candidateCount)
         {
             (results, _) = await SearchMoviesRaw(strippedTitle, includeRestricted: includeRestricted, year: year).ConfigureAwait(false);
-            CollectMovieCandidates(candidates, results, seen, candidateCount);
+            CollectMovieCandidates(candidates, results, seen, candidateCount, includeRestricted);
         }
 
         // Attempts #3–4: year-free fallbacks mirroring #1–2.
@@ -282,12 +282,12 @@ public partial class TmdbSearchService : ITmdbSearchService
         var yearFreeCap = candidateCount * 2;
 
         (results, _) = await SearchMoviesRaw(query, includeRestricted: includeRestricted).ConfigureAwait(false);
-        CollectMovieCandidates(candidates, results, seen, yearFreeCap);
+        CollectMovieCandidates(candidates, results, seen, yearFreeCap, includeRestricted);
 
         if (!string.IsNullOrEmpty(strippedTitle) && candidates.Count < yearFreeCap)
         {
             (results, _) = await SearchMoviesRaw(strippedTitle, includeRestricted: includeRestricted).ConfigureAwait(false);
-            CollectMovieCandidates(candidates, results, seen, yearFreeCap);
+            CollectMovieCandidates(candidates, results, seen, yearFreeCap, includeRestricted);
         }
 
         if (candidates.Count == 0)
@@ -575,7 +575,7 @@ public partial class TmdbSearchService : ITmdbSearchService
 
         // Attempt #1: full title + year
         (results, _) = await SearchShowsRaw(originalTitle, includeRestricted: restricted, year: airDate.Year).ConfigureAwait(false);
-        CollectCandidates(candidates, results, seen, candidateCount);
+        CollectCandidates(candidates, results, seen, candidateCount, restricted);
 
         // Attempt #2: sequel-suffix stripped + year
         var strippedTitle = SequelSuffixRemovalRegex().Match(originalTitle) is { Success: true } regexResult
@@ -583,7 +583,7 @@ public partial class TmdbSearchService : ITmdbSearchService
         if (!string.IsNullOrEmpty(strippedTitle) && candidates.Count < candidateCount)
         {
             (results, _) = await SearchShowsRaw(strippedTitle, includeRestricted: restricted, year: airDate.Year).ConfigureAwait(false);
-            CollectCandidates(candidates, results, seen, candidateCount);
+            CollectCandidates(candidates, results, seen, candidateCount, restricted);
         }
 
         // Attempt #3: subtitle stripped + year.
@@ -597,7 +597,7 @@ public partial class TmdbSearchService : ITmdbSearchService
         if (!string.IsNullOrEmpty(titleWithoutSubTitle) && candidates.Count < candidateCount)
         {
             (results, _) = await SearchShowsRaw(titleWithoutSubTitle, includeRestricted: restricted, year: airDate.Year).ConfigureAwait(false);
-            CollectCandidates(candidates, results, seen, candidateCount);
+            CollectCandidates(candidates, results, seen, candidateCount, restricted);
         }
 
         // Attempts #4–6: year-free fallbacks mirroring #1–3.
@@ -609,18 +609,18 @@ public partial class TmdbSearchService : ITmdbSearchService
         var yearFreeCap = candidateCount * 2;
 
         (results, _) = await SearchShowsRaw(originalTitle, includeRestricted: restricted).ConfigureAwait(false);
-        CollectCandidates(candidates, results, seen, yearFreeCap);
+        CollectCandidates(candidates, results, seen, yearFreeCap, restricted);
 
         if (!string.IsNullOrEmpty(strippedTitle) && candidates.Count < yearFreeCap)
         {
             (results, _) = await SearchShowsRaw(strippedTitle, includeRestricted: restricted).ConfigureAwait(false);
-            CollectCandidates(candidates, results, seen, yearFreeCap);
+            CollectCandidates(candidates, results, seen, yearFreeCap, restricted);
         }
 
         if (!string.IsNullOrEmpty(titleWithoutSubTitle) && candidates.Count < yearFreeCap)
         {
             (results, _) = await SearchShowsRaw(titleWithoutSubTitle, includeRestricted: restricted).ConfigureAwait(false);
-            CollectCandidates(candidates, results, seen, yearFreeCap);
+            CollectCandidates(candidates, results, seen, yearFreeCap, restricted);
         }
 
         if (candidates.Count == 0)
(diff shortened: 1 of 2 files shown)


Foowy:
Placing into draft pending updated TMDBLib

Foowy:
CodeQL/Integration Testing/and Units Tests will fail until merge with production TMDB Lib, ignoring for now

Shoko.Server/Models/TMDB/TMDB_Episode.cs
```diff
@@ -113,6 +113,13 @@ public int? RuntimeMinutes
     /// </summary>
     public TimeSpan? Runtime { get; set; }
 
+    /// <summary>
+    /// TMDB's own episode classification (<c>standard</c>, <c>finale</c>,
+    /// <c>mid_season</c>, <c>special</c>), verbatim from the API. <see langword="null"/>
+    /// until the episode is refreshed against a library build that fetches it.
+    /// </summary>
+    public string? TmdbEpisodeType { get; set; }
```
revam, line 121: Add an enum, byte backed with json serializers set, in the abstractions under the metadata tmdb namespace. and use that instead of an opaque string.

Shoko.Server/Providers/TMDB/TmdbMetadataService.cs
```diff
@@ -600,101 +610,178 @@ public async Task<bool> UpdateMovie(TmdbMovieUpdateOptions options)
         }
     }
 
-    private async Task<bool> UpdateMovieCastAndCrew(TMDB_Movie tmdbMovie, MovieCredits credits, bool forceRefresh, bool downloadImages)
+    // Shared by UpdateMovieCastAndCrew/UpdateEpisodeCastAndCrew — cast ordering is our own
+    // zero-based counter, not TMDb's `order` field, so existing sort order survives re-fetches
+    // even when TMDb reshuffles theirs.
+    private static (List<TCast> ToSave, List<TCast> ToRemove, int Added) DiffCast<TCredit, TCast>(
```
revam, line 616: if it's shared, then move it to the shared section at the bottom.

Shoko.Server/Providers/TMDB/TmdbSearchService.cs
```diff
@@ -770,24 +770,30 @@ private static bool MovieMatchesYear(TMDbLib.Objects.Movies.Movie movie, int yea
...
+    // Restricted titles also accept a Japanese original_language match — TMDB's genre tags are sparse
+    // for adult content, but original_language is TMDB-assigned metadata, not community-tagged.
+    private static void CollectCandidates(List<SearchTv> candidates, List<SearchTv> results, HashSet<int> seen, int candidateCount, bool isRestricted = false)
     {
         foreach (var result in results)
         {
             if (candidates.Count >= candidateCount) break;
             if (!seen.Add(result.Id)) continue;
-            if (!result.GetGenres().Contains(AnimationGenre, StringComparer.OrdinalIgnoreCase)) continue;
+            var isAnimeLike = result.GetGenres().Contains(AnimationGenre, StringComparer.OrdinalIgnoreCase) ||
```
revam, line 781: are we going to be biased towards japanese animation works? what about chinese, or korean?

Shoko.Server/Shoko.Server.csproj
```diff
@@ -50,6 +50,7 @@
   <ItemGroup>
     <ProjectReference Include="..\Shoko.Abstractions\Shoko.Abstractions.csproj" />
     <ProjectReference Include="..\Shoko.QueueProcessor\Shoko.QueueProcessor.csproj" />
+    <ProjectReference Include="..\ext\TMDbLib\TMDbLib\TMDbLib.csproj" />
```
revam, line 53: we will not be using a git submodule for the final builds. fine during testing while we wait for their upstream release, but the gh action doesn't need it unless you're somehow testing that in production? (I'll recommend not, for db migrations' sake)

Shoko.Tests/Providers/TMDB/TmdbAppendToResponseDeserializationTests.cs
```diff
@@ -0,0 +1,177 @@
...
+/// <summary>
+/// Guards the TMDbLib master upgrade (jellyfin/TMDbLib#626 + System.Text.Json migration).
+/// The upgrade keeps the exact same TMDB requests (verified by the reference-library
+/// re-scan, tasks T018/T023 — identical request counts and byte-for-byte row parity), so
+/// the only new risk is the wire→model binding: STJ presence handling of appended
+/// sub-resources and the <c>tvdb_id</c> type change (string → int). These replay the
+/// <c>append_to_response</c> payload shapes that <see cref="Shoko.Server.Providers.TMDB.TmdbMetadataService"/>
+/// requests and assert each sub-resource Shoko reads still deserializes.
+/// </summary>
+public class TmdbAppendToResponseDeserializationTests
```
revam, line 18: why do we have tests for the internals of the tmdblib inside our repo? this is fine during the wait for their release, but they should be removed once upstream is released and we switch over to it.

Shoko.Tests/Providers/TMDB/TmdbEpisodeTypeTests.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace Shoko.Tests.Providers.TMDB;
+
+/// <summary>
+/// User Story 5 / O4: capture TMDB's own <c>episode_type</c>. TMDbLib master exposes it on
+/// <see cref="TvSeasonEpisode.EpisodeType"/> (and <c>TvEpisodeBase</c>); before the upgrade
+/// Shoko had no episode-type signal from TMDB. This only persists the value (and surfaces it
+/// on the v3 DTO) — it deliberately does NOT feed episode matching (that would be a
+/// match-selection change, out of scope here; tracked as a follow-up).
+/// </summary>
+public class TmdbEpisodeTypeTests
```
revam, line 15: again, upstream tests, and also the enum change.

Shoko.Tests/Providers/TMDB/TmdbExternalIdTests.cs
```diff
@@ -0,0 +1,91 @@
+using Shoko.Server.Models.TMDB;
+using Shoko.Server.Providers.TMDB;
+using TMDbLib.Objects.General;
+using Xunit;
+
+namespace Shoko.Tests.Providers.TMDB;
```
revam, line 6: again, upstream tests.

Shoko.Server/Mappings/TMDB/TMDB_EpisodeMap.cs
```diff
@@ -24,6 +24,7 @@ public TMDB_EpisodeMap()
         Map(x => x.SeasonNumber).Not.Nullable();
         Map(x => x.EpisodeNumber).Not.Nullable();
         Map(x => x.RuntimeMinutes).Column("Runtime");
+        Map(x => x.TmdbEpisodeType).Nullable();
```
revam, line 27: needs a `.CustomType<TmdbEpisodeType>()` once it is enum backed.

Shoko.Server/Databases/SQLite.cs
```diff
@@ -1023,6 +1023,7 @@ WHERE CrossReferences LIKE '%AnidbEpisodeID%'
         new(171,  1, "CREATE TABLE Anilist_Anime_Suggestion ( Anilist_Anime_SuggestionID INTEGER PRIMARY KEY AUTOINCREMENT, AnilistAnimeID INTEGER NOT NULL, SuggestedAnilistAnimeID INTEGER NOT NULL, Rating INTEGER NOT NULL, Ordering INTEGER NOT NULL );"),
         new(171,  2, "CREATE UNIQUE INDEX UIX_Anilist_Anime_Suggestion_AnimeID_SuggestedID ON Anilist_Anime_Suggestion(AnilistAnimeID, SuggestedAnilistAnimeID);"),
         new(171,  3, "CREATE INDEX IX_Anilist_Anime_Suggestion_SuggestedID ON Anilist_Anime_Suggestion(SuggestedAnilistAnimeID);"),
+        new(172,  1, "ALTER TABLE TMDB_Episode ADD COLUMN TmdbEpisodeType TEXT NULL DEFAULT NULL;"),
```
revam, line 1026: integer, once enum backed.

Shoko.Server/API/v3/Models/TMDB/TmdbEpisode.cs
```diff
@@ -105,6 +105,14 @@ public class TmdbEpisode
     /// </summary>
     public TimeSpan? Runtime { get; init; }
 
+    /// <summary>
+    /// TMDB's own episode classification (<c>standard</c>, <c>finale</c>,
+    /// <c>mid_season</c>, <c>special</c>), verbatim from the API. <c>null</c>
+    /// until the episode has been refreshed since the field was added.
+    /// </summary>
+    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+    public string? TmdbEpisodeType { get; init; }
```
revam, line 114: enum.

revam (commented):
Mid draft observations.

PR #1399, opened 2026-08-03
fix(db): add missing primary keys on SQL Server for 38 tables

Foowy:
## Summary

SQLite and MySQL both declare `PRIMARY KEY` on the identity column for these tables; the SQL Server migration file never did, leaving them as heaps on that backend only. Confirmed live via `sys.dm_db_missing_index_details` against a production instance — `TMDB_Episode_Cast` and `TMDB_Episode` alone accounted for the two highest-impact missing-index findings, forcing RID lookups instead of clustered-key seeks on every query against their ID columns.

Adds migration 180: `ALTER TABLE ... ADD CONSTRAINT PRIMARY KEY CLUSTERED` for all 38 affected tables (mostly `TMDB_*` and a handful of `AniDB_*`/cross-reference tables).

## Test plan

- `dotnet build Shoko.Server.sln` — clean, 0 errors
- Dry-ran the actual `ADD CONSTRAINT` statements in a rolled-back transaction against a live production SQL Server database — no conflicts, confirms existing IDENTITY values are unique as expected
- Not run against SQLite/MySQL migration paths since those backends are unaffected — no schema change needed there

### DIFF
--- a/Shoko.Server/Databases/SQLServer.cs
+++ b/Shoko.Server/Databases/SQLServer.cs
@@ -1054,6 +1054,49 @@ public class SQLServer(SystemService systemService) : BaseDatabase<SqlConnection
                      FROM StoredReleaseInfo sri
                      WHERE sri.CrossReferences LIKE '%AnidbEpisodeID%'
                      """),
+        // These tables were created without a primary key on SQL Server only — SQLite and MySQL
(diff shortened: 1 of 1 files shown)


Foowy:
@da3dsoul tagging you for review since this touches SQL Server schema/migrations, which is normally your area.

PR #1401, opened 2026-08-10
repo(workflows): replace archived publish-nuget action, resolve Node 20 deprecation

Foowy:
## Summary
- `brandedoutcast/publish-nuget` is archived (unmaintained since 2022) and its Node 20 runtime is now being force-run on Node 24 by GitHub's runners, surfacing a `Node.js 20 is deprecated` annotation on every daily/release build.
- Replaced it in both `build-daily.yml` and `build-release.yml` with a composite `./.github/actions/publish-nuget` action that wraps `dotnet pack` / `dotnet nuget push --skip-duplicate` plus the existing `rickstaa/action-create-tag` step, preserving tag-on-publish behavior.
- Also bumped `build-daily.yml`'s explicitly pinned `node-version: 20` (used for the release-info npm packages) to 24 — a separate source of the same warning.
- Checked all other actions across both workflows: everything else already floats on a major-version tag (`@v7`, `@v4`, etc.) and auto-tracks latest, so nothing else needed bumping.

## Test plan
- [x] `build-daily.yml`'s composite action pattern already exercised locally against the branch diff
- [ ] Verify a real daily build run picks up the new composite action cleanly (opening as draft pending this)
- [ ] Confirm release-workflow NuGet publish jobs behave identically (skip-duplicate, tag creation) once run on a real release

## Rebase history
- 2026-08-13 17:22 UTC — rebased onto `master` @ `d4dff0c4d` (1 commits), now at `123ddbc72`

### DIFF
--- a/.github/actions/publish-nuget/action.yml
+++ b/.github/actions/publish-nuget/action.yml
@@ -1,0 +1,49 @@
+name: Publish NuGet package
+description: >
+  Packs and pushes a NuGet package, skipping already-published versions, and
+  tags the commit on a successful publish. Replaces the archived
+  brandedoutcast/publish-nuget action.
+
+inputs:
+  project-file-path:
+    description: Path to the .csproj to pack and publish.
+    required: true
+  nuget-key:
+    description: NuGet.org API key.
+    required: true
+  tag-format:
+    description: >
+      Git tag to create on publish, with * replaced by the package version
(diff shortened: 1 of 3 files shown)


Foowy:
@revam could you take a look when you get a chance?

revam (approved):
A bit late, since I haven't really been ti GH this week. This seems fine though. Good work.

PR #1404, opened 2026-08-16
feat: add plugin-registrable action system (RFC #1380)

revam:
## Summary

Implements the plugin-registrable action system from RFC [#1380](https://github.com/ShokoAnime/ShokoServer/discussions/1380) (spec v7).

- `IExecutableAction` with required `Name`/`Permission`, optional `Validate` pre-check, `RequiresConfirmation` UI hint, and the `SeriesAction`/`GroupAction`/`EpisodeAction` base classes with an internal `IScopedAction` context setter, plus `IActionCaller` for caller identity.
- `ActionService` registration with load-time validation (permission must be declared on the type itself; `IScopedAction` bypasses are rejected), UUIDv5 action IDs namespaced by plugin ID.
- `ActionExecutionJob` — every action always runs through the job queue, deduped by action + scope entity + caller.
- New endpoints: `GET /api/v3/Action?scope=`, `POST /api/v3/Action/{actionId}`, and `POST /api/v3/{Group|Series|Episode}/{id}/Action/{actionId}` — 200 ack / 400 validation rejection / 404 unknown entity, no tracking ID.
- Ported ~60 existing actions from `ActionController` and the scoped controllers to the new shape, using the Gap 23 two-bucket split (variant classes for closed parameter sets, defaults-only or settable properties for open-ended ones). Legacy endpoints remain available, marked `[Obsolete]`.
- Core actions are discovered through the plugin manager like any other plugin — no separate core registration.

## Beyond the RFC (bolt-ons)

The RFC defines the HTTP API and the registration model, but never addresses how **plugins** list or invoke actions programmatically. This PR adds:

- **`IActionService` in `Shoko.Abstractions`** — plugins can enumerate registered actions (`GetActions`/`GetActionInfo`) and invoke them via typed `InvokeAsync` overloads for global/series/group/episode scopes. Invocation still goes through the queue; the result is `null` on acceptance or an `ActionValidationResult` reason on rejection.
- **Invoke-time scope validation** — invoking an action with a mismatched scope (or no entity for a scoped action) is rejected up front instead of failing later with a cast error inside the job.
- **Caller semantics** — `IActionCaller` actions require a non-null caller and are rejected otherwise; trusted programmatic invocations without a caller skip the `Permission` gate.
- **Permission enforcement on the HTTP invoke path** — the RFC models `Permission` but the reference controllers don't enforce it; here admin-only actions are rejected for non-admin callers.

## Notes

- `Shoko.Abstractions` gains `InternalsVisibleTo("Shoko.Server")` so the load-time `IScopedAction` guard (Gap 28) is enforceable from the server.
- Full solution builds; `Shoko.Tests` passes (517/517).

### DIFF
--- a/Shoko.Abstractions/Actions/ActionCategory.cs
+++ b/Shoko.Abstractions/Actions/ActionCategory.cs
@@ -1,0 +1,68 @@
+using System.Text.Json.Serialization;
+using Newtonsoft.Json.Converters;
+
+namespace Shoko.Abstractions.Actions;
+
+/// <summary>
+///   The category of an executable action. Categories group related actions
+///   together in the UI and can be used to filter the action list.
+/// </summary>
+/// <remarks>
+///   This is a closed, core-owned enum — a plugin cannot invent a new
+///   core-owned category at runtime; adding one requires a PR against core.
+///   <see cref="Miscellaneous"/> is the shared fallback for any action that
+///   declares no category. <see cref="PluginInferred"/> is an explicit opt-in
+///   a plugin uses to request its own dedicated group; its display label is
+///   always the owning plugin's own name.
+/// </remarks>
+[JsonConverter(typeof(JsonStringEnumConverter))]
+[Newtonsoft.Json.JsonConverter(typeof(StringEnumConverter))]
+public enum ActionCategory : byte
+{
+    /// <summary>
+    ///   File import and scanning actions.
+    /// </summary>
+    Import = 0x01,
+
+    /// <summary>
+    ///   AniDB metadata and synchronization actions.
+    /// </summary>
+    AniDB = 0x21,
+
+    /// <summary>
+    ///   TMDB metadata and synchronization actions.
+    /// </summary>
+    TMDB = 0x22,
+
+    /// <summary>
+    ///   Data synchronization actions across providers.
+    /// </summary>
+    Sync = 0x31,
+
+    /// <summary>
+    ///   Image download and management actions.
+    /// </summary>
+    Images = 0x71,
+
+    /// <summary>
+    ///   System maintenance actions.
+    /// </summary>
+    Maintenance = 0xF1,
+
+    /// <summary>
+    ///   The shared fallback category for any action — core or plugin — that
+    ///   declares no category.
+    /// </summary>
+    Miscellaneous = 0xF2,
+
+    /// <summary>
+    ///   Destructive operations such as purging data.
+    /// </summary>
+    Destructive = 0xFE,
+
+    /// <summary>
+    ///   An explicit opt-in a plugin uses to request its own dedicated group.
+    ///   The display label is always the owning plugin's own name.
+    /// </summary>
+    PluginInferred = 0xFF,
+}
--- a/Shoko.Abstractions/Actions/ActionPermission.cs
+++ b/Shoko.Abstractions/Actions/ActionPermission.cs
@@ -1,0 +1,24 @@
+using System.Text.Json.Serialization;
+using Newtonsoft.Json.Converters;
+
+namespace Shoko.Abstractions.Actions;
+
+/// <summary>
+///   The permission required to invoke an executable action. Orthogonal to
+///   <see cref="ActionScope"/> — an action of any scope may require either
+///   permission level.
+/// </summary>
+[JsonConverter(typeof(JsonStringEnumConverter))]
+[Newtonsoft.Json.JsonConverter(typeof(StringEnumConverter))]
+public enum ActionPermission
+{
+    /// <summary>
+    ///   Only administrators may invoke the action.
+    /// </summary>
+    Admin,
+
+    /// <summary>
+    ///   Any authenticated user may invoke the action.
+    /// </summary>
+    User,
+}
--- a/Shoko.Abstractions/Actions/ActionScope.cs
+++ b/Shoko.Abstractions/Actions/ActionScope.cs
@@ -1,0 +1,45 @@
+using System.Text.Json.Serialization;
+using Newtonsoft.Json.Converters;
+
+namespace Shoko.Abstractions.Actions;
+
+/// <summary>
+///   The entity level an executable action is bound to.
+/// </summary>
+/// <remarks>
+///   <para>
+///     Global actions implement <see cref="IExecutableAction"/> directly.
+///     Scoped actions derive from one of the three base classes
+///     (<see cref="SeriesAction"/>, <see cref="GroupAction"/>,
+///     <see cref="EpisodeAction"/>), which fix the scope at compile time.
+///   </para>
+///   <para>
+///     Actions available for a given scope do not vary by <em>which</em>
+///     series/group/episode is being viewed, only by which entity type it is,
+///     so listings are scope-filtered, not entity-filtered.
+///   </para>
+/// </remarks>
+[JsonConverter(typeof(JsonStringEnumConverter))]
+[Newtonsoft.Json.JsonConverter(typeof(StringEnumConverter))]
+public enum ActionScope
+{
+    /// <summary>
+    ///   The action is not tied to a particular entity.
+    /// </summary>
+    Global,
+
+    /// <summary>
+    ///   The action operates on a group.
+    /// </summary>
+    Group,
+
+    /// <summary>
+    ///   The action operates on a series.
+    /// </summary>
+    Series,
+
+    /// <summary>
+    ///   The action operates on an episode.
+    /// </summary>
+    Episode,
+}
--- a/Shoko.Abstractions/Actions/ActionValidationResult.cs
+++ b/Shoko.Abstractions/Actions/ActionValidationResult.cs
@@ -1,0 +1,11 @@
+namespace Shoko.Abstractions.Actions;
+
+/// <summary>
+///   The result of a failed <see cref="IExecutableAction.Validate"/> call.
+///   Returned by actions to reject an invocation before it is enqueued; the
+///   API maps it to a 400 response with <see cref="Reason"/> as the message.
+/// </summary>
+/// <param name="Reason">
+///   The reason the invocation was rejected.
+/// </param>
+public sealed record ActionValidationResult(string Reason);
--- a/Shoko.Abstractions/Actions/EpisodeAction.cs
+++ b/Shoko.Abstractions/Actions/EpisodeAction.cs
@@ -1,0 +1,38 @@
+using System.Threading;
+using System.Threading.Tasks;
+using Shoko.Abstractions.Metadata.Shoko;
+
+namespace Shoko.Abstractions.Actions;
+
+/// <summary>
+///   Base class for episode-scoped executable actions. Derive from this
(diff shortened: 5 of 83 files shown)


hidden4003:
## Review: PR #1404 vs. RFC spec v7

Reviewed against [`v7-draft-op.md`](specs/plugin-registrable-actions/v7-draft-op.md) (all 31 tracked gaps resolved, posted to discussion #1380 on 2026-08-08). The core types, `ActionService.AddParts` load-time validation (Permission declaration check, `IScopedAction` guard), the 5-route endpoint shape, and the bucket-1/bucket-2 classification across the ~60 ported actions all track the spec closely. Two issues found below.

### 1. Gap 23 bucket-2 parameter population is not implemented

**Files:** `Shoko.Server/Scheduling/Jobs/Actions/ActionExecutionJob.cs:79`, `Shoko.Abstractions/Actions/Services/IActionService.cs`

The spec's "Execution service (Gap 15/18/23)" section specifies that `ActionExecutionJob` populates a bucket-2 action's free-form properties from `JobDataJson` before calling `Execute`, specifically to support invocation "by a plugin calling an action programmatically." In this implementation, `ActionExecutionJob.Execute()` resolves the target action from DI and calls `Execute()` directly — only the job's own four properties (`ActionId`/`Scope`/`ScopeEntityId`/`CallerUserId`) are ever populated from job data; the action's own settable properties (e.g. `PurgeAllTmdbLinksAction.RemoveShowLinks`/`RemoveMovieLinks`/`ResetAutoLinkingState`) are never touched.

Compounding this, `IActionService.InvokeAsync` — the PR's own plugin-facing addition, built specifically for programmatic invocation — has no overload that accepts a parameter payload of any kind.

**Effect:** every bucket-2 action can only ever run with its compiled-in default values, regardless of caller (WebUI, HTTP API, or plugin via `IActionService`). The scenario the spec explicitly carved this mechanism out for is currently unreachable.

### 2. Unknown `actionId` is handled inconsistently across the four invoke routes

**Files:** `Shoko.Server/API/v3/Controllers/ActionController.cs:103-104` vs. `SeriesActionController.cs:37-41`, `GroupActionController.cs`, `EpisodeActionController.cs`

The Global route (`ActionController.Invoke`) checks `GetActionInfo(actionId) is null` before invoking and returns a clean `404`. The three scoped controllers (`Series`/`Group`/`Episode`) only validate that the *entity* exists, then call `InvokeAsync` directly. For an unregistered action ID, this reaches `ActionService.InvokeCoreAsync`, which throws `KeyNotFoundException("No action registered for {actionId}")` with nothing to catch it — there's no global exception→status-code middleware in the API pipeline (`APIExtensions.cs` only wires Sentry capture and `UseDeveloperExceptionPage`).

**Effect:** the same "unknown action ID" case returns a clean `404` on the Global route but surfaces as an unhandled exception (500-class response) on the three scoped routes, despite the spec presenting all four invoke routes as symmetric.

---

No other findings rose to reportable severity — permission and category assignments across the ported actions matched their prior `[Authorize]` state, and the scope-validation, caller-identity, and `IScopedAction` hardening logic behaved as specified.

hidden4003:
## Review: PR #1404 update (follow-up)

Re-reviewed against the latest push. Both previously reported issues are resolved:

- Unknown `actionId` on the Series/Group/Episode invoke routes now returns a clean `404`, matching the Global route.
- Bucket-2 invocation parameters are now supported end-to-end: `ActionExecutionJob` gained a `Parameters` dictionary populated via `JsonConvert.PopulateObject` before `Execute()` runs, and `IActionService` gained parallel parameterized overloads for all four scopes. The HTTP endpoints correctly remain parameter-less, keeping WebUI out of scope per the spec.

The parameter-population fix introduces one new issue, plus a related design gap worth noting.

### 1. Dedup key does not include invocation parameters

**File:** `Shoko.Server/Scheduling/Jobs/Actions/ActionExecutionJob.cs:63`

`ActionExecutionJob` carries explicit `[JobKeyMember]` attributes on `ActionId`, `ScopeEntityId`, and `CallerUserId`. The new `Parameters` property has no such attribute. Per `Shoko.QueueProcessor`'s documented dedup rules, once explicit `[JobKeyMember]` annotations are present they replace the "all public settable properties" fallback — so `Parameters` never participates in the dedup key. Job-data merging on a dedup collision is opt-in (via `IJobMerge` or a registered merge handler), and `ActionExecutionJob` implements neither.

**Effect:** invoking the same action on the same scope and caller twice in quick succession with different parameters collapses the second call into a no-op against the already-queued first job. The second caller's parameters are silently discarded — only the first invocation's values ever execute — while `InvokeAsync` still returns `null` (accepted) to the second caller, incorrectly implying their specific request was queued.

### 2. `Validate()` runs before parameters are populated

**File:** `Shoko.Server/Services/ActionService.cs` (`InvokeCoreAsync`, ~line 355)

The probe instance used for the synchronous pre-enqueue `Validate()` call is never populated with the caller-supplied `parameters` — population only happens later, inside `ActionExecutionJob`, on a separately-resolved instance. No action currently shipped in this PR combines bucket-2 settable properties with a `Validate()` override, so this doesn't affect any existing behavior. It does mean a future bucket-2 action that wants to validate against its own parameter values won't be able to — `Validate()` will always observe default property values rather than what the caller actually passed.

---

Recommend addressing (1) before merge, since it silently drops caller-supplied data under a plausible concurrent-invocation pattern. (2) is a design note for future bucket-2 actions with `Validate()` logic, not a current defect.

harshithmohan (changes requested):
## Request changes: planning/spec artifacts and broken comments left in code

Non-functional — only comments. These reference an external plan/spec or a "current" legacy state, and won't make sense to a future reader once the migration context is gone.

### 1. "Gap N" spec references (16)

These point at a numbered planning doc ("Gap 1/29", "Gap 23 bucket 2", …) that isn't in the repo. The numbering is meaningless without that doc; the surrounding prose should be kept but the `Gap N` tokens dropped (reword the sentence to state the "what/why" directly).

- `Shoko.Server/Services/ActionService.cs`
  - `:187` `Gap 28`
  - `:203` `Gap 21`
  - `:204` `Gap 5`
  - `:214` `Gap 8/19`
  - `:225` `Gap 1/29`
  - `:278` `Gap 23 bucket 2`
  - `:365` `Gap 24`
  - `:367` `Gap 9`
  - `:387` `Gap 15`
  - `:388` `Gap 9`
  - `:404` `Gap 7`
- `Shoko.Server/Scheduling/Jobs/Actions/ActionExecutionJob.cs`
  - `:58` `Gap 23 bucket 2`
  - `:91` `Gap 9`
  - `:94` `Gap 23 bucket 2`
  - `:118` `Gap 11`
- `Shoko.Server/Actions/Import/RemoveMissingFilesNoMyListSyncAction.cs`
  - `:13` `Gap 23 bucket 1`

### 2. "Matches today" temporal parity notes (45)

One-liners above the `Permission` property in nearly every action class, e.g. `// Matches today: [Authorize("admin")] on the legacy ... endpoint`. "Today" is transient and migration-scoped. The permission choice is non-obvious, so don't just delete them — reword to a permanent invariant, e.g.:

```csharp
// Legacy endpoint had no admin gate; keep User-level to avoid regressing callers.
```

Affected files (45):

<details>
<summary>Full list</summary>

- `AniDB/VerifyAllRelationsAction.cs`
- `AniDB/AddAllManualLinksToMyListAction.cs`
- `AniDB/UpdateMissingAnidbFileInfoAction.cs`
- `AniDB/GetAnidbNotificationsAction.cs`
- `AniDB/RefreshAnidbMovedFilesAction.cs`
- `AniDB/UpdateAllAnidbInfoAction.cs`
- `AniDB/AVDumpMismatchedFilesAction.cs`
- `AniDB/SyncAnidbMyListAction.cs`
- `AniDB/DownloadMissingAnidbCreatorsAction.cs`
- `AniDB/DownloadMissingAnidbAnimeDataAction.cs`
- `AniDB/UpdateAnidbCalendarAction.cs`
- `Images/DownloadAllImagesAction.cs`
- `Images/ValidateAllImagesAction.cs`
- `Images/PurgeAllUnusedTmdbImagesAction.cs`
- `Images/UpdateAllImagesAction.cs`
- `Import/ImportNewFilesAction.cs`
- `Import/RunImportAction.cs`
- `Import/RemoveMissingFilesAction.cs`
- `Import/RemoveMissingFilesNoMyListSyncAction.cs`
- `Maintenance/PurgeAllUsedReleasesAction.cs`
- `Maintenance/CreateMissingSeriesAction.cs`
- `Maintenance/RecreateAllGroupsAction.cs`
- `Maintenance/UpdateAllMediaInfoAction.cs`
- `Maintenance/PurgeAllUnusedReleasesAction.cs`
- `Maintenance/PlexSyncAllAction.cs`
- `Maintenance/RenameAllGroupsAction.cs`
- `Maintenance/UpdateSeriesStatsAction.cs`
- `Tmdb/UpdateAllTmdbMoviesWithImagesAction.cs`
- `Tmdb/UpdateAllTmdbMoviesAction.cs`
- `Tmdb/PurgeAllUnusedTmdbShowsAction.cs`
- `Tmdb/SearchForTmdbMatchesAction.cs`
- `Tmdb/PurgeAllTmdbMovieCollectionsAction.cs`
- `Tmdb/UpdateAllTmdbShowsWithImagesAction.cs`
- `Tmdb/PurgeAllTmdbLinksAction.cs`
- `Tmdb/PurgeAllUnusedTmdbMoviesAction.cs`
- `Tmdb/PurgeAllTmdbShowAlternateOrderingsAction.cs`
- `Tmdb/DownloadMissingTmdbPeopleAction.cs`
- `Tmdb/UpdateAllTmdbShowsAction.cs`
- `Tmdb/Series/AutoMatchTmdbEpisodesSeriesAction.cs`
- `Tmdb/Series/AutoSearchTmdbSeriesAction.cs`
- `Tmdb/Series/UpdateTmdbImagesForceSeriesAction.cs`
- `Tmdb/Series/UpdateTmdbInfoSeriesAction.cs`
- `Tmdb/Series/RefreshTmdbMoviesSeriesAction.cs`
- `Tmdb/Series/ResetTmdbEpisodeMappingsSeriesAction.cs`
- `Tmdb/Series/DownloadTmdbMovieImagesSeriesAction.cs`

</details>

### 3. Stale/contradictory comment — `ActionService.cs:188-190`

```csharp
// three base classes. Redundant once IScopedAction is `internal` (a plugin
// assembly literally cannot implement an internal interface from another
// assembly), but kept as a documented, load-time-checked guard.
```

`IScopedAction` is *already* `internal` (`Shoko.Abstractions/Actions/IScopedAction.cs:13`), so the "once … internal" framing is stale — the guard is already redundant today. Either remove the guard (and comment) or correct the comment to reflect that it's currently redundant and intentionally kept.

### 4. Typo — `Shoko.Abstractions/Actions/IExecutableAction.cs:90`

```csharp
///   report something log, same as the rest of the queue already does.
```

Missing word — read as "report something **to the** log" (or "report something **via** logging").

harshithmohan (approved):

PR #1410, opened 2026-08-17
feat(api): move release group endpoints into ReleaseInfo with source filtering

harshithmohan:
## Summary

Moves the AniDB release group endpoints into the release info API and generalizes them so they are no longer hardcoded to AniDB groups. Clients can now filter release groups by their `Source` (e.g. `AniDB`, `User`, or plugin-defined sources) instead of always receiving AniDB-only results.

### `GET /api/v3/ReleaseInfo/ReleaseGroup`

Lists the release groups known to Shoko.

| Parameter | Type | Default | Description |
| --- | --- | --- | --- |
| `pageSize` | `int` | `20` | Number of results per page. Set to `0` to disable pagination. |
| `page` | `int` | `1` | Page index. |
| `includeMissing` | `IncludeOnlyFilter` | `false` | `false` = used groups, `only` = unused groups, `true` = all. |
| `source` | comma-delimited `string` | *(all sources)* | Filters to the given group sources. |

Example: `GET /api/v3/ReleaseInfo/ReleaseGroup?source=AniDB&includeMissing=true`

### `GET /api/v3/ReleaseInfo/ReleaseGroup/{source}/{groupID}`

Returns a single release group identified by its source and ID.

- `source` — the release group source (e.g. `AniDB`, `User`).
- `groupID` — the release group ID.

Returns `404` when no group matches.

Example: `GET /api/v3/ReleaseInfo/ReleaseGroup/AniDB/69`

## Deprecated endpoints

- `GET /api/v3/AniDB/ReleaseGroup`
- `GET /api/v3/AniDB/ReleaseGroup/{groupID}`

Both are now marked `[Obsolete]` and will be removed in a future version. Their behavior is unchanged until then.

## Why `source` + `ID` for the detail endpoint

A release group's identity is the `(GroupID, GroupSource)` pair — `GroupID` alone is **not** unique across sources (the repository index is keyed on `(GroupID, GroupSource)`, and results are deduplicated by that pair). The detail endpoint therefore takes both a `source` and a `groupID`.

This also fixes a latent bug in the old `GET /api/v3/AniDB/ReleaseGroup/{groupID}` endpoint, which pattern-matched the returned collection as a single `IReleaseInfo` (never true), so it always returned `404`.

## Behavior changes for existing consumers

- `ExpressionDiscovery` (`HasReleaseGroupNameExpression` autocomplete) now returns **used** release groups from **all** sources, instead of only AniDB used groups. The v1 `GET /api/v1/ReleaseGroups` endpoint is unchanged (still returns AniDB used groups only).

## Repository changes

`StoredReleaseInfoRepository`'s three release group queries (`GetReleaseGroups`, `GetUsedReleaseGroups`, `GetUnusedReleaseGroups`) now accept an optional `IReadOnlySet<string>?` source filter and share a single `GetReleaseGroupsCore` implementation. The old parameterless AniDB-only overloads were removed; the obsolete AniDB endpoint and the v1 endpoint now pass `"AniDB"` explicitly.

## Testing

- `dotnet build Shoko.Server.sln` — succeeds with 0 errors.

### DIFF
--- a/Shoko.Server/API/v1/Implementations/ShokoServiceImplementation_Utilities.cs
+++ b/Shoko.Server/API/v1/Implementations/ShokoServiceImplementation_Utilities.cs
@@ -126,7 +126,7 @@ public partial class ShokoServiceImplementation
     [HttpGet("ReleaseGroups")]
     public List<string> GetAllReleaseGroups()
     {
-        return RepoFactory.StoredReleaseInfo.GetUsedReleaseGroups().Select(r => r.Name).ToList();
+        return RepoFactory.StoredReleaseInfo.GetUsedReleaseGroups(null).Select(r => r.Name).ToList();
     }
 
     [HttpGet("File/DeleteMultipleFilesWithPreferences/{userID}")]
--- a/Shoko.Server/API/v1/Implementations/ShokoServiceImplementation_Utilities.cs
+++ b/Shoko.Server/API/v1/Implementations/ShokoServiceImplementation_Utilities.cs
@@ -126,7 +126,7 @@ public partial class ShokoServiceImplementation
     [HttpGet("ReleaseGroups")]
     public List<string> GetAllReleaseGroups()
     {
-        return RepoFactory.StoredReleaseInfo.GetUsedReleaseGroups(null).Select(r => r.Name).ToList();
+        return RepoFactory.StoredReleaseInfo.GetUsedReleaseGroups(new HashSet<string> { "AniDB" }).Select(r => r.Name).ToList();
     }
 
     [HttpGet("File/DeleteMultipleFilesWithPreferences/{userID}")]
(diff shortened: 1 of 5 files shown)


Shoko.Server/API/v1/Implementations/ShokoServiceImplementation_Utilities.cs
```diff
@@ -126,7 +126,7 @@ public List<CL_AniDB_Anime> SearchAnimeWithFilename(int uid, [FromForm] string q
     [HttpGet("ReleaseGroups")]
     public List<string> GetAllReleaseGroups()
     {
-        return RepoFactory.StoredReleaseInfo.GetUsedReleaseGroups().Select(r => r.Name).ToList();
+        return RepoFactory.StoredReleaseInfo.GetUsedReleaseGroups(null).Select(r => r.Name).ToList();
```
revam: I think for legacy reasons, this should still be scoped to "AniDB" only.

revam (approved):

PR #1411, opened 2026-08-18
fix(search): match titles when the query carries a trailing qualifier

revam:
Since `dc486893c` ("Better Fuzzy Search"), a query carrying a trailing token the stored title lacks — most visibly a ` (<year>)` — returns nothing at all.

| Query | Stored title | Before | After |
|---|---|---|---|
| `Sousou no Frieren (2023)` | `Sousou no Frieren` | **miss** | match |
| `Boku no Hero Academia (2024)` | `Boku no Hero Academia` | **miss** | match |
| `Bleach (2004)` | `Bleach` | **miss** | match |
| `Steins;Gate (2011)` | `Steins;Gate` | **miss** | match |
| `Sousou no Frieren` | `Sousou no Frieren (2023)` | match | match |

The mirror direction always worked, which is why this went unnoticed — a series *stored* with a year in its title is still found by a query without one.

## Cause

`MinEditDistInText` is semi-global over the **text only**: row 0 is all zeros (free text-prefix skip) and the result is the minimum over the last row (free text-suffix skip). Slop in the **query** is charged in full. Two gates then reject the pair:

1. **The length gate** — `normalized.Length >= normalizedQuery.Length - maxErrors` skips the comparison outright when the query is longer, so no distance is ever computed.
2. **The error cap** — `GetMaxErrors` caps at 3, and appending a year is always exactly 5 edits. Even with the gate removed it could never pass, at *any* query length.

The parentheses are innocent: `NormalizeForIndex` already maps `(`, `)`, `[` and `]` to spaces, so `[2023]` and a bare `2023` fail identically. The cliff isn't year-specific either — `S1` and `(TV)` squeaked through at exactly `dist == maxErrors`, while anything four or more characters longer fell off.

Before `dc486893c` the Sørensen–Dice path tolerated all of these; it compared bigram sets, so five unmatched characters out of twenty-two barely moved the score.

## Fix

`SeriesSearch.TryGetFuzzyDistance` replaces the length gate and the direct distance call at both fuzzy call sites (`FuzzySearchIndex.SearchLatin`, `SeriesSearch.SearchCollection`):

- **title ≥ query** — unchanged, `MinEditDistInText(query, title)`.
- **query > title** — arguments swapped, moving the free prefix/suffix onto the query.
- **length guard** — the swapped path requires `title.Length * 2 >= query.Length`, so a two-letter title like `GK` can't match nearly every query. It applies **only** to the swapped direction, so a short typo'd query still matches a long title exactly as before (`belach` → `Gekijouban Bleach: Memories of Nobody`).

`GetMaxErrors` is untouched — the cap was never the right lever, since the cost of a year is constant regardless of query length.

## Behavioural note for reviewers

This removes the cliff for **all** trailing qualifiers, not just years — `(TV)`, `S1`, and similar slop now match on the same footing. That restores the pre-`dc486893c` Dice behaviour, but it is a slightly wider change than "years now work", so it's worth a second opinion on whether the `* 2` guard is the right threshold.

## Testing

- `Shoko.Tests` — 542/542 pass.
- Reverting only the two source files while keeping the new tests fails **13** of them, so the tests are not vacuous.
- `Debug` and `Release` both build clean, 0 warnings.

New coverage in `FuzzySearchIndexTests`: `TryGetFuzzyDistance` in both directions plus both guard rejections; end-to-end cases for `(2004)`, `[2024]`, a bare year, `(TV)`, and a typo stacked on a trailing year; and cases pinning the mirror direction, `ExactMatch` staying false for slop matches, strict (non-fuzzy) mode staying strict, and short titles not matching much longer queries.

---

Investigated, written and verified by Claude.

### DIFF
--- a/Shoko.Server/Utilities/FuzzySearchIndex.cs
+++ b/Shoko.Server/Utilities/FuzzySearchIndex.cs
@@ -143,9 +143,8 @@ public class FuzzySearchIndex<T>
                         Result = item,
                     };
                 }
-                else if (fuzzy && normalized.Length >= normalizedQuery.Length - maxErrors && SeriesSearch.IsLatinScript(normalized))
+                else if (fuzzy && SeriesSearch.IsLatinScript(normalized) && SeriesSearch.TryGetFuzzyDistance(normalizedQuery, normalized, out var dist))
                 {
-                    var dist = SeriesSearch.MinEditDistInText(normalizedQuery, normalized);
                     if (dist > maxErrors)
                         continue;
 
--- a/Shoko.Server/Utilities/SeriesSearch.cs
+++ b/Shoko.Server/Utilities/SeriesSearch.cs
@@ -148,6 +148,42 @@ public static class SeriesSearch
         return min;
     }
 
+    /// <summary>
+    /// Computes the edit distance between a normalized query and a normalized title,
+    /// treating leading and trailing slop on whichever of the two is longer as free.
+    /// </summary>
+    /// <remarks>
+    /// <see cref="MinEditDistInText"/> only skips a free prefix/suffix in the text, so a
+    /// query carrying tokens the title lacks — e.g. "Sousou no Frieren (2023)" against
+    /// "Sousou no Frieren" — is charged for every extra character and blows past
+    /// <see cref="GetMaxErrors"/>. Searching the longer of the two for the shorter one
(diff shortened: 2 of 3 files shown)


da3dsoul:
Good job with the tests. If they pass, lgtm

PR #1412, opened 2026-08-18
fix: missing-episode stats going stale, hiding series from the Missing Episodes page

Foowy:
## Summary

Reported symptom: the dashboard's "Missing Episodes (Total)" tile shows a count greater than zero, but clicking it opens the Missing Episodes page (`/webui/utilities/missing-episodes`) and it says **"No series with missing episodes!"** — completely empty, not just a lower count.

Root cause: `VideoReleaseService.SaveReleaseForVideo` never triggers a stats recalculation for the anime a file was just linked to. `AnimeSeries.MissingEpisodeCount` — the field the dashboard sums — only gets refreshed via AniDB MyList sync, group-status refresh, or a manual API call, never via the normal file-import pipeline. So as soon as a file is imported for a previously-missing episode, the cached count goes stale and stays wrong until something unrelated happens to touch that series. Meanwhile `MissingEpisodes/Series` recomputes live on every request, so it correctly shows nothing — hence the disconnect between the two.

Along the way I also found and fixed a related but separate issue: `GetMissing()`'s raw SQL required `AirDate != 0`, which silently drops episodes with an unknown air date even when the anime itself has already finished airing. `AniDB_Episode.HasAired` (used by the cached stat) correctly treats those as aired. This didn't cause the reported symptom, but it's the same category of bug (live query and cached stat disagreeing) and was straightforward to align while in this code.

## Changes

- `Shoko.Server/Services/VideoReleaseService.cs`: enqueue `RefreshAnimeStatsJob` for each affected anime after a release is saved — the same job `GetAniDBReleaseGroupStatusJob` already uses for this purpose.
- `Shoko.Server/Repositories/Cached/AnimeEpisodeRepository.cs`: loosen the `AirDate` filter in the missing-episodes SQL to a coarse pre-filter, and apply the exact `HasAired` check in C#, following this file's existing SQL-narrows/C#-filters pattern (see `GetWithMultipleReleases`/`GetWithDuplicateFiles` in the same file).

## Test plan

- [x] Reproduced on a local-dev build against a real library: dashboard showed `MissingEpisodes: 1` while `MissingEpisodes/Series` returned `Total: 0`.
- [x] Confirmed via direct DB query that the series in question had zero episodes actually matching "missing" criteria — the persisted count was simply stale.
- [x] Applied the fix, ran `/api/stats_update` once to clear existing stale data, confirmed dashboard and live query now agree (`0` / `0`).
- [x] `dotnet build Shoko.Server.sln` — clean, 0 warnings, 0 errors.

### DIFF
--- a/Shoko.Server/Repositories/Cached/AnimeEpisodeRepository.cs
+++ b/Shoko.Server/Repositories/Cached/AnimeEpisodeRepository.cs
@@ -217,7 +217,8 @@ GROUP BY
 
     // Group_CompletionStatus.Complete = 3, Group_CompletionStatus.Finished = 5
     // EpisodeType.Episode = 1
-    // AirDate is stored as Unix timestamp (seconds). LastEpisodeNumber >= EpisodeNumber approximates HasGroupReleasedEpisode.
+    // AirDate = 0 means unknown; kept as a candidate here and resolved exactly via HasAired in GetMissing().
+    // LastEpisodeNumber >= EpisodeNumber approximates HasGroupReleasedEpisode.
     // GS.GroupID (int) = SRI.GroupID (varchar) relies on implicit int↔string coercion present in SQLite, MySQL, and SQL Server.
     private const string MissingEpisodesQuery = @"
 SELECT AE.AnimeEpisodeID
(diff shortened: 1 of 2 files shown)


Foowy:
@da3dsoul this touches VideoReleaseService and the missing-episodes query, flagging you for review.

PR #1421, opened 2026-09-03
Unify missing episode definition and refresh AniDB group status periodically

harshithmohan:
## Summary

Addresses GitHub Discussion #1418 — the definition of "missing episode" was inconsistent across the v3 API, and `AniDB_GroupStatus` data went stale without the legacy v1 API.

### 1. Refresh `AniDB_GroupStatus` during periodic anime updates (`e2d0e2d8d`)

`AniDB_GroupStatus` was only refreshed by the legacy v1 API, so on a v3-only install the cached release-group progress stopped updating and "released-but-absent" missing-episode counts under-counted.

- `GetUpdatedAniDBAnimeJob` now enqueues `GetAniDBReleaseGroupStatusJob` for each updated series, propagating `ForceRefresh` so scheduled runs keep the `ShouldSkip()` ended-anime gate intact (no unnecessary UDP traffic for ended anime).

### 2. Unify the missing-episode definition (`6ec2e3c23`)

`/v3/Series` → `Sizes.Missing.Episodes` used an aired-but-absent check while `/v3/MissingEpisodes` and the persisted counts used a released-but-absent check via `AniDB_GroupStatus`, so the same series could report different missing counts.

- Added `AnimeEpisodeExtensions.IsMissingEpisode` as the single source of truth: aired, not hidden, no local file, and either no group status data or a group with `CompletionState` complete/finished or `LastEpisodeNumber` past the episode number.
- Applied the group status gate to `Sizes.Missing.Episodes` in both `GenerateSeriesSizes` and `GenerateGroupSizes` so series and group counts agree. `GenerateSeriesSizes` now *requires* the per-anime status dictionary so future callers cannot fall back to the divergent aired-but-absent logic.
- Replaced the non-collecting `GetMissing` raw SQL with cache-based filtering through the shared predicate; the collecting variant stays as raw SQL.
- Added `AniDB_GroupStatusRepository.GetByAnimeIDs` with batched (1000-per-query) lookups.

Specials are intentionally unchanged (still bare `HasAired`), per the existing comment in `AnimeSeriesService` about there being no reliable way to check specials.

## Verification

- `dotnet build Shoko.Server.sln` — 0 errors, 0 warnings
- `dotnet test Shoko.Tests/Shoko.Tests.csproj` — 2607 passed, 1 failed (`YearlySeasonsTests.RealWorldRegression_LateSeasonPremiere_IsClassifiedOnlyUnderItsOwnSeason`, pre-existing on master and unrelated to these changes — fixed separately in #1420)

## ⚠️ Known caveat: group-status N+1 on list endpoints

Both DTO constructors now query the (uncached) `AniDB_GroupStatus` repository:

- `new Series(...)` runs one `GetByAnimeID` query **per series**
- `GenerateGroupSizes` runs one batched query **per group**

On single-item endpoints (`GET /v3/Series/{id}` etc.) this is one extra indexed query against a small table — negligible. But list endpoints (`GET /v3/Series` with `pageSize=100`, `/v3/Group`, filter endpoints, dashboard, tree) now issue one extra query per constructed DTO, where previously they made zero group-status queries.

**Question for maintainers:** should this be fixed in this PR (batch-load statuses once per page via `GetByAnimeIDs` and thread them through the `Series`/`Group` constructors), or left as-is/a follow-up? A batch-loading fix has already been implemented and validated locally, so it can be added quickly if preferred — it's deliberately not part of this PR to keep the diff focused.

### DIFF
--- a/Shoko.Server/Extensions/AnimeEpisodeExtensions.cs
+++ b/Shoko.Server/Extensions/AnimeEpisodeExtensions.cs
@@ -1,0 +1,36 @@
+using System.Collections.Generic;
+using System.Linq;
+using Shoko.Abstractions.Metadata.Enums;
+using Shoko.Server.Models.AniDB;
+using Shoko.Server.Models.Shoko;
+
+namespace Shoko.Server.Extensions;
+
+/// <summary>
+/// Extension methods for <see cref="AnimeEpisode"/>.
+/// </summary>
+public static class AnimeEpisodeExtensions
+{
+    /// <summary>
+    /// Determines whether an aired episode with no local files should be counted as missing,
+    /// based on the cached AniDB group release statuses for its anime.
+    /// </summary>
+    /// <param name="episode">The episode to evaluate.</param>
+    /// <param name="groupStatuses">Group statuses already scoped to the episode's anime. An empty list is treated as missing.</param>
+    /// <returns><see langword="true"/> if the episode is considered missing; otherwise, <see langword="false"/>.</returns>
+    /// <remarks>
+    /// This predicate is intended for <see cref="EpisodeType.Episode"/> episodes only; callers must
+    /// pre-filter by episode type AND file presence before calling it.
+    /// </remarks>
+    public static bool IsMissingEpisode(this AnimeEpisode episode, IReadOnlyList<AniDB_GroupStatus> groupStatuses)
+    {
+        if (episode.IsHidden) return false;
+        var anidb = episode.AniDB_Episode;
+        if (anidb == null) return false;
+        if (!anidb.HasAired) return false;
+
+        return groupStatuses.Count == 0 || groupStatuses.Any(gs =>
+            gs.CompletionState is (int)GroupCompletionStatus.Complete or (int)GroupCompletionStatus.Finished
+            || gs.LastEpisodeNumber >= anidb.EpisodeNumber);
+    }
+}
--- a/Shoko.Server/Repositories/Cached/AnimeEpisodeRepository.cs
+++ b/Shoko.Server/Repositories/Cached/AnimeEpisodeRepository.cs
@@ -6,6 +6,7 @@ using NHibernate;
 using NutzCode.InMemoryIndex;
 using Shoko.Abstractions.Extensions;
 using Shoko.Server.Databases;
+using Shoko.Server.Extensions;
 using Shoko.Server.Models.Shoko;
 
 using EpisodeType = Shoko.Abstractions.Metadata.Enums.EpisodeType;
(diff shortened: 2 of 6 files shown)


Foowy:
@harshithmohan High-level Claude code review of this PR (opus, high effort):

**Correctness**
- `Shoko.Server/Repositories/Cached/AnimeEpisodeRepository.cs:289` - the `collecting=true` branch of `GetMissing` dropped the precise `HasAired` recheck that the `collecting=false` branch still applies, so the two branches can now disagree on what counts as "aired" for episodes with unknown air dates on ongoing anime. Worth confirming this divergence is intentional.

**Design/standards conformance**
- `Shoko.Server/Extensions/AnimeEpisodeExtensions.cs:91` - `IsMissingEpisode` documents "callers must pre-filter by episode type" as a precondition but doesn't enforce it. Only two call sites exist today and both comply, but nothing stops a future caller from passing a Special/Credits episode and getting silently wrong results.
- `Shoko.Server/API/v3/Helpers/ModelHelper.cs:237` - the new `GenerateSeriesSizes` signature is 172 chars, over the repo's documented 160-char line-length limit.
- `Shoko.Server/API/v3/Helpers/ModelHelper.cs:328` - minor: `IsMissingEpisode` re-resolves `episode.AniDB_Episode` even though the caller already has it in a local var one line up. No correctness issue, just a redundant lookup.

**Gotcha (already flagged by the PR author)**
- `Shoko.Server/API/v3/Models/Shoko/Series.cs:137` - every `Series` DTO now issues an extra uncached `AniDB_GroupStatus.GetByAnimeID` query, so paginated list endpoints (e.g. `GET /v3/Series`) go from 0 to N extra queries per page as library size grows.

Net: one real correctness question (the aired-check divergence), one soft-contract gap around episode-type pre-filtering, a line-length lint violation, and the N+1 query concern the author already flagged.

harshithmohan:
Thanks for the review. Addressing each point:

**`AnimeEpisodeRepository.cs:289` (aired-check divergence)** — This is intentional; the two paths answer different questions. The `collecting=true` SQL path is *evidence-driven*: an episode with an unknown air date only qualifies because of the `StoredReleaseInfo` `EXISTS` clause — a release group with a matching `GroupStatus` has actually released files for this anime (up to this episode number or completed it), so we treat the episode as aired and collectible regardless of whether AniDB has filled in the air date yet. Unknown air date on AniDB means either "not scheduled yet" or "actually unknown" — release evidence is the stronger signal in both cases. The `collecting=false` path answers the stricter "what's genuinely missing" question via `HasAired`, so it stays conservative. The divergence won't be aligned.

**`AnimeEpisodeExtensions.cs:91` (episode-type precondition)** — Accepted; it's a soft contract enforced at both call sites today. Keeping it as documented rather than adding a runtime guard.

**`ModelHelper.cs:237` (line length)** — Fixed in f6a6287ea.

**`ModelHelper.cs:328` (redundant lookup)** — Fixed in f6a6287ea; `IsMissingEpisode` now takes the `AniDB_Episode` explicitly and callers pass what they already have.

**`Series.cs:137` (N+1 on `AniDB_GroupStatus.GetByAnimeID`)** — Known trade-off, flagged during development. Open to a batch-loading follow-up if it shows up in profiling, but leaving it for now.

Shoko.Server/Extensions/AnimeEpisodeExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// </summary>
+    /// <param name="episode">The episode to evaluate.</param>
+    /// <param name="anidb">The AniDB episode for <paramref name="episode"/>.</param>
+    /// <param name="groupStatuses">Group statuses already scoped to the episode's anime. An empty list is treated as missing.</param>
+    /// <returns><see langword="true"/> if the episode is considered missing; otherwise, <see langword="false"/>.</returns>
+    /// <remarks>
+    /// This predicate is intended for <see cref="EpisodeType.Episode"/> episodes only; callers must
+    /// pre-filter by episode type AND file presence before calling it.
+    /// </remarks>
+    public static bool IsMissingEpisode(this AnimeEpisode episode, AniDB_Episode anidb, IReadOnlyList<AniDB_GroupStatus> groupStatuses)
```
revam: Place this on AnimeEpisode itself. We don't need extensions for our own models.

Shoko.Server/Repositories/Cached/AnimeEpisodeRepository.cs
```diff
@@ -326,26 +285,43 @@ public IEnumerable<AnimeEpisode> GetMissing(bool collecting, int? animeID = null
...
+            .Where(a => a.AirDate == 0 || a.AirDate < currentTime)
+            .ToDictionary(a => a.EpisodeID);
+
+        // Filter Shoko episodes by valid AniDB episode and optional anime ID
+        var episodes = GetAll()
+            .Where(e => !e.IsHidden && e.AniDB_EpisodeID > 0 && anidbEpisodes.ContainsKey(e.AniDB_EpisodeID))
+            .ToList();
+
+        if (animeID.HasValue)
+            episodes = episodes.Where(e => anidbEpisodes[e.AniDB_EpisodeID]!.AnimeID == animeID.Value).ToList();
```
revam: can you fix this sonar complaint?

Shoko.Server/Repositories/Cached/AnimeEpisodeRepository.cs
```diff
@@ -326,26 +285,43 @@ public IEnumerable<AnimeEpisode> GetMissing(bool collecting, int? animeID = null
...
+
+        // Load group statuses and file cross-references in batch
+        var animeIDs = episodes.Select(e => anidbEpisodes[e.AniDB_EpisodeID]!.AnimeID).Distinct().ToList();
```
revam: same sonar complaint here.

Shoko.Server/Repositories/Cached/AnimeEpisodeRepository.cs
```diff
@@ -326,26 +285,43 @@ public IEnumerable<AnimeEpisode> GetMissing(bool collecting, int? animeID = null
...
+        var groupStatusesByAnime = RepoFactory.AniDB_GroupStatus.GetByAnimeIDs(animeIDs);
+        var episodeIDsWithFiles = RepoFactory.CrossRef_File_Episode.GetAll()
+            .Select(f => f.EpisodeID)
+            .ToHashSet();
+
+        // Apply the shared missing-episode predicate
+        return episodes
+            .Where(e => !episodeIDsWithFiles.Contains(e.AniDB_EpisodeID) &&
+                e.IsMissingEpisode(anidbEpisodes[e.AniDB_EpisodeID], groupStatusesByAnime.GetValueOrDefault(anidbEpisodes[e.AniDB_EpisodeID]!.AnimeID) ?? []))
```
revam: and here.

revam (changes requested):
minor changes, but can you address them?

revam (approved):

PR #1426, opened 2026-09-14
fix(abstractions): make LanguageExtensions unknown-language tracking thread-safe

Foowy:
## Summary
- `GetTitleLanguage()` mutated an unsynchronized static `HashSet<string>` to dedupe unknown-language reports; concurrent callers (e.g. API requests building File DTOs) corrupted it and crashed with `InvalidOperationException`
- Replaced it with a `ConcurrentDictionary<string, byte>` used as a set, so no external synchronization is needed at any call site

Opened at Revam's request.

Corrected the new concurrency test to assert on results (SonarCloud S2699 flagged it as assertion-free).


Shoko.Abstractions/Extensions/LanguageExtensions.cs
```diff
@@ -17,7 +17,8 @@ public static class LanguageExtensions
     /// </summary>
     public static Action<string>? OnUnknownLanguage;
 
-    private static readonly HashSet<string> _reportedUnknowns = new(StringComparer.OrdinalIgnoreCase);
+    // ConcurrentDictionary used as a thread-safe set; concurrent callers otherwise corrupt a plain HashSet.
```
revam: comment can go

revam (changes requested):

revam (approved):

PR #1441, opened 2026-09-15
chore(deps): bump outdated NuGet packages

Foowy:
Bumps all outdated NuGet packages across the solution to their latest available patch/minor versions.

## Bumped

**Shoko.Server**
| Package | From | To | Changelog |
|---|---|---|---|
| `Asp.Versioning.Mvc.ApiExplorer` | 10.0.0 | 10.2.1 | [Releases](https://github.com/dotnet/aspnet-api-versioning/releases) |
| `FluentNHibernate` | 3.4.1 | 3.5.0 | [Releases](https://github.com/FluentNHibernate/fluent-nhibernate/releases/tag/v3.5.0) |
| `Magick.NET-Q8-AnyCPU` | 14.16.0 | 14.17.1 | [Releases](https://github.com/dlemstra/Magick.NET/releases) |
| `Microsoft.AspNetCore.Mvc.NewtonsoftJson` | 10.0.10 | 10.0.12 | [.NET 10 release notes](https://github.com/dotnet/core/tree/main/release-notes/10.0) |
| `Microsoft.AspNetCore.ResponseCaching` | 2.3.11 | 2.3.13 | [Releases](https://github.com/dotnet/aspnetcore/releases) |
| `Microsoft.AspNetCore.SignalR` | 1.2.11 | 1.2.13 | [Releases](https://github.com/dotnet/aspnetcore/releases) |
| `Microsoft.AspNetCore.SignalR.Protocols.NewtonsoftJson` | 10.0.10 | 10.0.12 | [.NET 10 release notes](https://github.com/dotnet/core/tree/main/release-notes/10.0) |
| `Microsoft.CodeAnalysis.Analyzers` | 5.6.0 | 5.9.0 | [Releases](https://github.com/dotnet/roslyn-analyzers/releases) |
| `Microsoft.CodeAnalysis.CSharp` | 5.6.0 | 5.9.0 | [Releases](https://github.com/dotnet/roslyn/releases) |
| `Microsoft.CodeAnalysis.NetAnalyzers` | 10.0.302 | 10.0.401 | [Releases](https://github.com/dotnet/roslyn-analyzers/releases) |
| `Microsoft.Data.SqlClient` | 7.0.2 | 7.0.3 | [Releases](https://github.com/dotnet/SqlClient/blob/main/CHANGELOG.md) |
| `Microsoft.Data.Sqlite` | 10.0.10 | 10.0.12 | [.NET 10 release notes](https://github.com/dotnet/core/tree/main/release-notes/10.0) |
| `Microsoft.Extensions.DependencyInjection.Abstractions` | 10.0.10 | 10.0.12 | [.NET 10 release notes](https://github.com/dotnet/core/tree/main/release-notes/10.0) |
| `MySqlConnector` | 2.6.1 | 2.6.2 | [Releases](https://mysqlconnector.net/changelog/) |
| `NLog` | 6.1.4 | 6.2.0 | [Releases](https://github.com/NLog/NLog/releases/tag/v6.2.0) |
| `NLog.Extensions.Logging` | 6.1.4 | 6.2.0 | [Releases](https://github.com/NLog/NLog.Extensions.Logging/releases) |
| `NLog.Web.AspNetCore` | 6.1.4 | 6.2.0 | [Releases](https://github.com/NLog/NLog.Web/releases) |
| `Polly` | 8.7.0 | 8.8.0 | [Releases](https://github.com/App-vNext/Polly/releases/tag/8.8.0) |
| `Sentry` / `Sentry.AspNetCore` / `Sentry.Extensions.Logging` / `Sentry.NLog` | 6.8.0 | 6.11.0 | [CHANGELOG](https://github.com/getsentry/sentry-dotnet/blob/main/CHANGELOG.md) |
| `SharpCompress` | 0.50.1 | 0.50.4 | [Releases](https://github.com/adamhathcock/sharpcompress/releases) |

**Shoko.TrayService**
| Package | From | To | Changelog |
|---|---|---|---|
| `Avalonia` / `Avalonia.Desktop` / `Avalonia.Themes.Fluent` | 12.1.0 | 12.1.2 | [Releases](https://github.com/AvaloniaUI/Avalonia/releases) |
| `Tmds.DBus.Protocol` | 0.94.2 | 0.95.1 | [Releases](https://github.com/tmds/Tmds.DBus/releases) |

**Shoko.QueueProcessor / Shoko.QueueProcessor.Tests**
| Package | From | To | Changelog |
|---|---|---|---|
| `Microsoft.EntityFrameworkCore` / `.Design` / `.Sqlite` / `.SqlServer` | 10.0.10 | 10.0.12 | [.NET 10 release notes](https://github.com/dotnet/core/tree/main/release-notes/10.0) |
| `Microsoft.Extensions.DependencyInjection.Abstractions` / `Hosting.Abstractions` / `Logging.Abstractions` | 10.0.10 | 10.0.12 | [.NET 10 release notes](https://github.com/dotnet/core/tree/main/release-notes/10.0) |
| `Microting.EntityFrameworkCore.MySql` | 10.0.10 | 10.0.11 | [Releases](https://github.com/Microting/entity-framework-core-mysql/releases) |

**Shoko.Tests / Shoko.IntegrationTests / Shoko.QueueProcessor.Tests / Shoko.BuildTools / Shoko.BuildTools.Targets**
| Package | From | To | Changelog |
|---|---|---|---|
| `Microsoft.NET.Test.Sdk` | 18.8.1 | 18.10.1 | [.NET 10 release notes](https://github.com/dotnet/core/tree/main/release-notes/10.0) |
| `Microsoft.CodeAnalysis.CSharp` (BuildTools) | 5.6.0 | 5.9.0 | [Releases](https://github.com/dotnet/roslyn/releases) |
| `Microsoft.Build` / `.Framework` / `.Utilities.Core` | 18.8.2 | 18.9.6 | [Releases](https://github.com/dotnet/msbuild/releases) (see note below — 18.10.1 exists but dropped net10.0) |

## Held back (not bumped)

- **`Microsoft.Build` / `Microsoft.Build.Framework` / `Microsoft.Build.Utilities.Core`** — bumped to **18.9.6** instead of the latest 18.10.1. Confirmed via nuget.org that 18.10.x **dropped `net10.0` support entirely** and jumped straight to `net11.0` (18.9.6 targets `net10.0`+`net472`; 18.10.1 targets `net11.0`+`net472`, no overlap) — restoring 18.10.1 against this repo's `net10.0` projects falls back to the `.NETFramework` TFM group and trips `NU1701` under the warnings-as-errors policy. This isn't a "wait for a newer build" situation: there's no net10.0-targeted 18.10.x coming. 18.9.6 is the last release in the net10.0-compatible line, so that's as close to latest as this repo can get before migrating to `net11.0` (out of scope here — .NET 11 ships Nov 2026 as STS/~18mo support vs .NET 10's LTS/3yr, not worth the shorter-support tradeoff for one build-tooling dependency).
- **`xunit.v3` / `xunit.runner.visualstudio`** (3.2.2/3.1.5 → 4.0.1/4.0.0 available, already the latest 3.x releases otherwise) — xUnit v4 defaults test projects to Microsoft.Testing.Platform (MTP), and the .NET 10 SDK refuses to run MTP-mode projects through the legacy VSTest `dotnet test` path (`error: Testing with VSTest target is no longer supported`). Fixing it needs a `global.json` opt-in (`{"test": {"runner": "Microsoft.Testing.Platform"}}`) and likely CI workflow changes, which is out of scope for a plain dependency bump. Confirmed against the actual v4 packages locally, then reverted. See [xunit v4.0.0 release notes](https://xunit.net/releases/v3/4.0.0) and [dotnet/sdk#51283](https://github.com/dotnet/sdk/issues/51283).

## Test plan
- [x] `dotnet build Shoko.Server.sln -c Release` — 0 warnings, 0 errors
- [x] `dotnet test Shoko.Tests` — 3944/3944 passing
- [x] `dotnet test Shoko.QueueProcessor.Tests` — 215/215 passing

## Rebase history
- 2026-09-18 12:45 UTC - rebased onto `upstream/master` @ `6bd003e5d` (69 commits), now at `e5e482b55`
- 2026-09-18 16:05 UTC - rebased onto `upstream/master` @ `7c7f38478` (2 commits), now at `3c4182148`


revam (commented):
Don't we need to bump the version for Shoko.BuildToolds, Shoko.BuildTools.Targets, and Shoko.QueueProcessor when we dump their dependencies? For the queue processor in particular it's a shared dependency of the core and plugins which depend on it, so they have to agree for it to work. Bumping for the sake of bumping on that package should be handled with a lot more care because of this. If it's forward compatible if you have a plugin targetting a lower version than what the server have, then it's fine.

Foowy:
Yes, we would want to bump at least  Shoko.QueueProcessor. The EF Core 10.0.10 → 10.0.12 update itself should be safe. Patch releases within the same major version are not expected to introduce breaking changes, so this is fundamentally different from an EF10 → EF11 upgrade.

For plugin compatibility, plugins are loaded into their own isolated AssemblyLoadContext (PluginManager.IsolatedLoadContext). If a plugin doesn't ship its own copy of a dependency, it will resolve against the version already loaded by the server. That means a plugin built against QueueProcessor/EF Core 10.0.10 should run without issue on a server using 10.0.12.

The main risk is when a plugin bundles its own copies of EF Core or QueueProcessor rather than relying on the server's version. If that is the case, separate assemblies would be loaded into different AssemblyLoadContexts. Since AddQueueProcessor exposes EF Core types through its public DI registration surface, any of those types crossing the plugin/server boundary can result in runtime type identity issues, such as InvalidCastExceptions or failed dependency resolution. From what I can find, it would show up as a failure to load plugin when starting server.

Generally this specific dependency bump is low risk. That said, QueueProcessor remains at 6.0.0-alpha.6 even though its dependencies changed, and nothing prevents a plugin from shipping a stale copy of those assemblies. My suggestion is to document/state that these dependency updates are considered non-breaking/minimal for plugin authors.

__________________________________________________________________________________________________________________________


We can bump Shoko.BuildToolds, Shoko.BuildTools.Targets as well for cleanliness, but from what i can see in a few quick tests, a version bump here doesn't even change what actually executes at runtime. Confirming the explanation below -

- Shoko.BuildTools bumps Microsoft.Build 18.8.2→18.9.6, Microsoft.CodeAnalysis.CSharp 5.6.0→5.9.0. It's PackAsTool="true" - a dotnet tool, invoked out-of-process by plugin authors to build their plugin project. It never loads into the same process as the running Shoko server.
- Shoko.BuildTools.Targets bumps Microsoft.Build.Framework/Utilities.Core the same way, and is a DevelopmentDependency MSBuild props/targets package - again build-time only, referenced by plugin projects at compile time, not shipped or loaded at runtime.
- Both mark their Microsoft.Build* refs ExcludeAssets="runtime" - standard practice for MSBuild-adjacent tooling, meaning at runtime they rely on whatever MSBuild/Roslyn the host dotnet SDK provides, not the pinned NuGet version. So the version bump here doesn't even change what actually executes at runtime, just what API surface it compiles against.


Hopefully that answers your question or I completely misunderstood the assignment and get an F.

revam (commented):
Bump the Shoko build tools versions and queue processor version and we're good, I think.

Foowy:
Minor bumped. Force was just a rebase against master to bring branch forward to current.

revam (approved):

